PR #222, opened 2017-11-09
Fix breakdown number alignment

peterszerzo:
Fixes https://trello.com/c/AakcOszB/63-breakdown-spacing-issue, making sure sure two-digit numbers don't push the bar to the right. Anything else I should look at in this change?


peterszerzo:
It'd overflow to the left a bit, not too terrible. I wouldn't compromise on looks and consistency for the average use-case to perfect that further. Or do you have some thoughts on how to do better?

TejasQ:
I'd say we can merge/close this now and deal with it later if 3 digits ever becomes a thing. What do you think?

PR #234, opened 2017-11-15
Propagate keys to components

peterszerzo:
Makes sure `key` props are passed down to the actual dom, implemented for all components where iteration might can come up.


peterszerzo:
@Angarsk8 that's absolutely right, thanks for the catch! We can use `id` for it, as that eludes to dom id's (though not always, such as in the case of a `<label for="someid"><input id="someid"/></label>`), and should be unique on the page - I'm like 90% sure it's a good way to go. What do you think?

Angarsk8:
@peterszerzo  I agree with you. In my opinion, to use any other identifier different to `key` or `id` would be weird in that case, in fact using `id` for that purpose feels already weird, but I'd rather use `id` than something like `cid` or `compKey` or whatever else.

peterszerzo:
I think that's a good compromise - I'll set that up first thing tomorrow!

PR #236, opened 2017-11-20
Refactor/processflow

ImogenF:
- Fix node label updating.
- Truncate long node labels.
- Simplify code, tidy, improve typings.


packages/showcase/src/pages/visualizations/ProcessFlows/propDescription.ts
```diff
@@ -148,14 +148,14 @@ export default {
     {
       name: "nodes",
       description: "Provides the attribute name for accessing nodes in the input data",
-      defaultValue: "data.nodes",
+      defaultValue: "(d) => d.nodes",
```
peterszerzo: I'd remove the parentheses from the `(d)`, looks a little cleaner.

packages/showcase/src/pages/visualizations/ProcessFlows/propDescription.ts
```diff
@@ -221,56 +221,56 @@ export default {
...
-      defaultValue: 'link.stroke || "#bbb"',
+      defaultValue: '(d) => d.stroke || "#bbb"',
       type: "string",
       optional: true
     },
     {
       name: "source",
       description: "Node at which the link starts",
-      defaultValue: "link.source || undefined",
+      defaultValue: "(d) => d.source || undefined",
```
peterszerzo: Why do we have `undefined` here?

packages/visualizations/src/ProcessFlow/data_handler.ts
```diff
@@ -73,14 +69,9 @@ class DataHandler {
...
 
-  setNodeAccessors(accessors: IAccessors): void {
-    this.nodeAccessors = new NodeAccessors()
-    this.nodeAccessors.setAccessors(accessors)
-  }
-
   addNode(attrs: {}): TNode {
     extend.convert({ immutable: false })(attrs, { size: 0 })
-    return new Node(attrs, this.nodeAccessors.accessors)
+    return new Node(attrs, this.state.current.get("accessors").node)
```
peterszerzo: The `"accessors"` string is easy to mistype with little compiler help. Not a fan of the approach here.

packages/visualizations/src/utils/focus.ts
```diff
@@ -1,7 +1,7 @@
 import Events from "./event_catalog"
 import { IEvents, IState, TSeriesEl, TStateWriter } from "./typings"
 
-abstract class AbstractFocus {
+abstract class Focus {
```
peterszerzo: Abstract class naming seems a bit inconsistent - I saw an instance where an `AbstractCanvas` is imported from `./utils/canvas`, whereas the abstract focus is imported as `Focus` into `DrawingFocus`. To be honest, I liked the old naming better, despite the redundancies.

packages/showcase/src/pages/visualizations/ProcessFlows/propDescription.ts
ImogenF: Makes sense, will do that

packages/visualizations/src/ProcessFlow/data_handler.ts
peterszerzo: As in I liked the previous version, sorry, didn't mean to be not-constructive-at-all :).

packages/showcase/src/pages/visualizations/ProcessFlows/propDescription.ts
ImogenF: `source` and `target` are generally assigned during the data processing. `d.source` may not exist depending on when the accessor is called. `undefined` is not required here, but this is just to make it clear that it's not necessarily an error if `link.source()` returns `undefined`.

packages/visualizations/src/ProcessFlow/data_handler.ts
ImogenF: What would you suggest instead?

ImogenF: The previous version was a lot of extra, unnecessary code. I could replace the `state.current.get` method with just `state.current.accessors` but that would be a lot of refactoring.

peterszerzo: It makes sense a lot of sense not to do that!

packages/visualizations/src/utils/focus.ts
ImogenF: Following our discussion, abstract classes will not have the `abstract_` filename prefix but will generally be imported as `Abstract...`. However, I think the exception to this should be if an abstract class is extended by another abstract class, i.e. since `DrawingFocus` is itself an abstract class and is in the same folder as the abstract `Focus` class, it will just import `Focus` rather than `AbstractFocus`. That ok?

peterszerzo: Perfect - there's plenty of context in the file to make abstractness clear, and it seems to be a rare occurrence anyway.

packages/visualizations/src/ProcessFlow/facade.ts
```diff
@@ -1,17 +1,43 @@
...
+  series: Series
+  state: StateHandler<IConfig>
+
+  constructor(context: Element) {
+    this.context = context
+    this.events = new EventEmitter()
+    this.initializeState()
+    this.insertCanvas()
+    this.initializeComponents()
+    this.initializeSeries()
```
micha-f: Can we move the assignments into the constructor?
```
this.state = this.initialState()
this.canvas = this.createCanvas()
etc. ...
```

micha-f: We're also doing this in some other constructors in other files. Ideally, we only set instance variables in the constructor. Makes it a lot simpler to see what the initial values for these instance variables are.

PR #238, opened 2017-11-24
Enhance/side navigation header

Angarsk8:


packages/ui-components/lib/src/SideNavigation/Header/SideNavigationHeader.d.ts
```diff
@@ -4,37 +4,32 @@ import { Theme } from "contiamo-ui-theme";
...
+        isOpen: boolean;
     };
-    componentDidMount(): void;
-    getDefaultValue(): IOption;
     toggle(): void;
-    onChange(option: IOption): Promise<void>;
-    getDropdown(): JSX.Element;
+    onChange(option: IOption): void;
+    labelFor(value: string): string;
+    displayDropdown(): JSX.Element;
```
peterszerzo: How about `renderDropdown`? We already expect the render method to return a JSX.Element, could be a nice cue here.

packages/ui-components/src/Timeline/TimelineItem.tsx
```diff
@@ -105,12 +105,12 @@ const TimelineItem: React.SFC<IPropsWithTheme> = ({
   const IconLib = ReactFeather as any
   const isValidIcon = IconLib.hasOwnProperty(icon)
   const Icon = IconLib[icon]
-  color = hexOrColor(color)(theme.colors.palette[color] || theme.colors.palette.info)
+  const statusColor = hexOrColor(color)(theme.colors.palette[color] || theme.colors.palette.info)
```
peterszerzo: Awesome!

packages/ui-components/src/Upload/Upload.tsx
```diff
@@ -124,11 +124,10 @@ export default Upload
 function checkStatus(response: Response): Response {
   if (response.ok) {
     return response
-  } else {
-    const error = new Error(response.statusText)
-    ;(error as any).response = response
-    throw error
   }
+  const error = new Error(response.statusText)
```
peterszerzo: Double awesome!

PR #240, opened 2017-11-27
Various minor usability bugs

peterszerzo:


packages/ui-components/src/Select/SelectOption.tsx
```diff
@@ -25,6 +25,7 @@ const Container = glamorous.div(
 
     return {
       backgroundColor,
+      wordWrap: "break-all",
```
TejasQ: ? Are you sure this is in the right place? [Shouldn't it be here?](https://github.com/Contiamo/contiamo-ui-components/blob/034f5099cb147c19536e6649fb100c783d500b25/packages/ui-components/src/Select/SelectOption.tsx#L42)

packages/ui-components/src/Select/SelectOption.tsx
```diff
@@ -37,6 +37,7 @@ const Container = glamorous.div(
 
       "&:not(:first-child)": {
         borderTop: "1px solid",
+        wordBreak: "break-all",
```
TejasQ: `wordWrap`!!!!!

PR #241, opened 2017-11-27
Implement journey highlighting in process flows.

ImogenF:
- Manual journey highlighting
- Optional hiding of focus labels when manually focussing nodes or links


packages/visualizations/src/ProcessFlow/focus.ts
```diff
@@ -14,9 +14,10 @@ interface IBreakdowns {
 // There can only be an element focus in process flow diagrams
 class Focus extends AbstractFocus {
 
-  onElementHover(payload: { focusPoint: IFocus; d: TNode | TLink }): void {
+  onElementHover(payload: { focusPoint: IFocus; d: TNode | TLink, hideLabel: boolean  }): void {
```
peterszerzo: Does it make sense to type annotate this consistent with the similar change in `abstract_renderer`? On a different note, if the intention here is a boolean parameter that is not always present (behavior defaults to false), then I'd annotate it as `hideLabel?: boolean` (easier to understand that it's not mandatory), and either use it as if it were a boolean but with the truthiness rules, or use it as proper booleans converting it using `!!hideLabel`.

@TejasQ what do you think?

packages/visualizations/src/ProcessFlow/renderers/links.ts
```diff
@@ -102,13 +102,13 @@ class Links extends AbstractRenderer {
     return "M" + xStart + "," + yStart + "L" + xMid + "," + yMid + "L" + xEnd + "," + yEnd
   }
 
-  highlight(element: TLinkSelection, d: TLink): void {
+  highlight(element: TLinkSelection, d: TLink, keepCurrent: boolean = false): void {
```
peterszerzo: Same comment as above

packages/visualizations/src/ProcessFlow/focus.ts
ImogenF: @peterszerzo In this case, I can't add an initialiser into this line (`hideLabel: boolean = false`) - I'd have to initialise the entire `payload` object, which would then be overwritten when the method is called. 

`hideLabel` will always be present, it is initialised to `false` at an earlier stage (https://github.com/Contiamo/contiamo-ui-components/blob/6b9ae0d0b3bb085bc9d3804a45d26c556e3cf00d/packages/visualizations/src/ProcessFlow/renderers/abstract_renderer.ts#L45)

What do you think is best re: making this all a bit clearer?

TejasQ: I think the default initialization makes things quite clear.

ImogenF: @TejasQ even though it's in a different place?

TejasQ: Oh! I was not aware it was in a different place!

Man, this is a _doozie_ in that case. With the limited context that I have as of this moment, I am unable to decide on an approach I like. I will look into this over the next 30 minutes before the next recruiting call and respond shortly.

TejasQ: It all makes sense now! Okay – this current approach makes sense to me – but it's a bit of a code smell that you (@ImogenF) had to clarify that `hideLabel` is initialised to false at an earlier stage. In my opinion, this needs to be made clear somehow. 

Imagine if a newer developer joins and tries understanding the code! @Angarsk8 @peterszerzo thoughts?

TejasQ:
Looks to me like this baby needs some rebasing 🤔

ImogenF:
@TejasQ We need to come to a decision about initialising `hideLabel` as discussed above. Will rebase once that is sorted.

TejasQ:
@ImogenF I thought we did? I didn't see the discussion in the latest diff? 🤔

ImogenF:
@TejasQ The discussion is still there, just marked as outdated because I made an unrelated change to the same file.

peterszerzo:
@ImogenF what is the status on this branch? Is it still active and need another review, or a meeting for a decision to be made?

ImogenF:
@peterszerzo As far as I'm concerned, the changes are fairly trivial and add functionality but don't change existing functionality, so it can be merged if we think this is a feature we may want (highlighting whole paths rather than single nodes or links). If not though, we can just delete the branch.

peterszerzo:
Let's discuss together when I'm back I'd say!

PR #245, opened 2017-11-30
Infra/build improvements

peterszerzo:
This PR lifts build dependencies out of the individual packages into the root of the repository. The biggest challenge is getting a consistent working dev and build environment set up. Things we've tried:

## Centralized build

Steps:
- build into root, then manually distribute built lib folders into the packages.
- manually add symlinks to the packages into the root (`make link` in this PR).
- test symlinks in finder.

Issue: tsc still doesn't find inner package dependencies, so build will always fail.

## Centralized build dependencies + local build

Steps:
- copy build binaries - tsc, prettier - into individual packages.
- build from packages like previously, but with single source of truth dependencies.

Issue: build dependencies need to be copied as well. Concerned this will turn into an unmaintainable recursive mess.

## Suggested: create script to keep build deps and boilerplate uniform across packages. Local build

(a'la `create-component.sh`)


packages/showcase/src/components/Sidebar/Sidebar.tsx
```diff
@@ -20,7 +20,7 @@ export interface ISidebarLink {
 export default ({ links, location, css }: IProps) => (
   <Sidebar css={css}>
     {links.map((link, index) => {
-      const isOpen = location && link.links.map(link => link.url).includes(location.pathname)
+      const isOpen = location && link.links.map(link => link.url).indexOf(location.pathname) > -1
```
TejasQ: Why?

peterszerzo: Typescript 2.6.2 was complaining..

PR #247, opened 2017-12-07
Rebranding

### LINKED ISSUE #235: Branding
In order to emphasize that these packages can be used outside of Contiamo and set them up for successful open-sourcing, I think it's best we divorce its brand from that of the company. Here is what I'm thinking:
`@operational/components`
`@operationa/blocks`
`@operational/utils`
`@operational/theme`
`@operational/visualizations`

I opted for dropping the `-ui-` bit from the names since the fact that UI is all we're dealing with here becomes more obvious with the `contiamo` bit gone.

I'll be posting logo ideas, this is what I have so far:
operational
Still inspired somewhat by the Contiamo logo. I get a bit of a coast guard/life vest feel that I don't quite like, but I'm sure we can work it out.

Please feel free to suggest something completely different, none of this is set in stone :)

peterszerzo:
Fix #235


package.json
```diff
@@ -1,8 +1,8 @@
 {
-  "name": "contiamo-ui",
-  "version": "1.0.0",
+  "name": "operational",
```
TejasQ: operational-ui

peterszerzo: I recall calling it operational in code, and branding it `Operational UI` on the website.

peterszerzo:
@TejasQ I'll merge this as it is so we can go ahead with finishing some related changes, and to update the live site. Let's discuss the repo name on Monday!

PR #248, opened 2017-12-08
Add linter and prettier to visualizations

peterszerzo:
I fixed most tslint errors (there were only about 15) and verified that the process flow visualization still works well. The prettier script is now in place, I just haven't run it yet to keep diffs sane.

Here are the linter errors that remain, involving reassignments of function parameters:
```
ERROR: src/ProcessFlow/layout.ts[160, 5]: Attempting to reassign to parameter
ERROR: src/utils/focus_utils.ts[53, 7]: Attempting to reassign to parameter
ERROR: src/utils/process_flow_loop_handler.ts[58, 7]: Attempting to reassign to parameter
```

Returning fresh copy of the parameters being re-assigned e.g. `possiblePositions` as an object field in the return value seems like a good way to go, but I don't want to step on other design decisions. Do you want to handle this in a new commit to this PR, @ImogenF?

Also, I regret to inform that `--n` assignments are not allowed, even though I love them dearly myself. I don't see it around that much in open source, so I think it's best to do `n = n - 1` - and that'll keep our linter configuration in the mainstream as well. I hope that's ok.


peterszerzo:
Also note that Travis will keep failing on this one because prettier has not run yet. Stay tuned for lights turning green in a few commits :)

ImogenF:
Amazing, thanks for doing this @peterszerzo! I'll get on those last few linter errors now.

PR #254, opened 2017-12-12
Sidenav redesign

peterszerzo:
Redesign steps/progress:

### Renames

* `SideNavigation` -> `Sidenav` (consider `Side` as a modifier instead of a full blown noun, removing the need to uppercase the `N`, consistent with `Sidebar`).
* `SideNavigationHeader` -> `SidenavHeader` (the square-shaped, expanding logo+text combination in the corner)
* `SideNavigationItem` -> `SidenavHeader` (this is pretty much the same as the component one line above, expect that one will get a bottom border through a prop or a first-child rule.
* `SideNavigationLink` -> `SidenavItem` (these are the smaller items nested inside headers)

### Side navigation options are removed

.. in favor of header dropdowns for changing datasets. Projects can also use a combination of headers and items if their project has few datasets, but a consistent approach is recommended.

### Opinionated components

Headers will expect an `icon` prop (either an icon name or an svg element), and a `value` string prop for the text, no custom children. The end user should nest the resulting operational components in their own `<Link>` or `<a>` tags to provide the linking functionality.


packages/components/src/Sidenav/SidenavItem.tsx
```diff
@@ -0,0 +1,22 @@
+import * as React from "react"
+import glamorous, { GlamorousComponent } from "glamorous"
+import { Theme } from "@operational/theme"
+
+export interface IProps {
+  id?: string | number
+  className?: string
+  children: React.ReactNode
```
TejasQ: Maybe better to make this `never`

packages/components/src/Sidenav/SidenavItem.tsx
```diff
@@ -0,0 +1,22 @@
...
+  onClick?: () => void
+  active?: boolean
+  css?: any
```
TejasQ: Maybe better to make this `css: {}`

packages/components/src/Sidenav/SidenavItem.tsx
peterszerzo: Love it :)

packages/components/src/Sidenav/SidenavItem.tsx
peterszerzo: Willdo!

packages/components/src/Sidenav/SidenavItem.tsx
TejasQ: I have not. If `never` doesn't work, feel free to use different a implementation, but I do think it's important to communicate to other devs implementing this component that it _does not accept children_.

TejasQ: Bump because this isn't outdated though GitHub thinks it's outdated.

packages/components/src/Breakdown/Breakdown.tsx
```diff
@@ -38,9 +38,9 @@ const Container = glamorous.div(
           }
         }
       : {},
-    background: theme.colors.palette.white,
+    background: theme.colors.white,
     ":not(:first-child)": {
-      borderTop: `1px solid ${theme.colors.palette.grey20}`
+      borderTop: `1px solid ${theme.colors.gray20}`
```
TejasQ: Why grey -> gray?

packages/visualizations/src/ProcessFlow/styles.ts
```diff
@@ -1,5 +1,5 @@
 import { css } from "glamor"
-import { contiamoTheme as theme } from "@operational/theme"
+import { operational as theme } from "@operational/theme"
```
TejasQ: Could `operational` be a `const` somewhere in an app that will force users into importing it as something else often?

packages/showcase/src/components/SideNavigation/SideNavigation.tsx
```diff
@@ -3,7 +3,7 @@ import glamorous, { Div, Img, withTheme } from "glamorous"
 import { Link } from "react-router-dom"
```
TejasQ: This is unfortunate. Can we factor this dependency out? 🤔

packages/components/src/Breakdown/Breakdown.tsx
peterszerzo: It is more common in American English.

packages/showcase/src/components/SideNavigation/SideNavigation.tsx
peterszerzo: @TejasQ what is unfortunate about it?

packages/components/src/Breakdown/Breakdown.tsx
TejasQ: Good. I suggest we adhere closer to American English where possible. 👍

packages/visualizations/src/ProcessFlow/styles.ts
peterszerzo: I don't think that'll come up too often - in this case, we can just use it in the code as operational.

packages/showcase/src/components/SideNavigation/SideNavigation.tsx
TejasQ: I think it's fairly unfortunate that we ship our components with `react-router-dom`, you know? As far as I remember, the `Link`ing is up to the consumer: we just give them nice UI elements that they can link themselves.

packages/visualizations/src/ProcessFlow/styles.ts
TejasQ: 💯

packages/showcase/src/components/SideNavigation/SideNavigation.tsx
peterszerzo: This is `showcase`, not `components` :)

TejasQ: So much facepalm. I saw `src/components` in the pathspec and got confused. Let's pretend this never happened.

TejasQ: PLEASE FORGIVE ME

packages/visualizations/src/ProcessFlow/styles.ts
peterszerzo: @TejasQ on second thought, I'll leave it like this - using the `theme` alias makes styling code look the same whether it's inside a glamorous component or just raw out in the open.

We should probably think about a way to not have to import the theme statically in visualization modules, but have them passed in dynamically, so users can change it.

TejasQ: Agree. Good thinking!

packages/components/scripts/dev-server/site.tsx
```diff
@@ -3,32 +3,29 @@ import { render } from "react-dom"
...
+          </SidenavHeader>
+          <SidenavHeader label="Documentation" icon="Edit" />
+        </Sidenav>
       </ThemeProvider>
     )
   }
 }
 
-injectStylesheet(baseStylesheet(contiamoTheme))
+injectStylesheet(baseStylesheet(operationalTheme))
```
TejasQ: This is beautiful.

PR #256, opened 2017-12-13
Improve check unchanged script

peterszerzo:
Improves `./scripts/check-unchanged.sh`, which verifies on Travis that files aren't changed after lint+prettier+package are run:
* ignore `.map` files (sometimes diffs creep in for some strange reason, but we shouldn't care about this)
* improve regex readability/extensibility


TejasQ:
Dude you merged it with Travis failing...

peterszerzo:
Travis failures are exactly what this PR is addressing, bit of a Catch22 :).

Unfortunately, because tslint and prettier step on each other's toes and we have to rely on the somewhat brittle `check-unchanged.sh` script, our Travis setup will take just a tad longer to stabilize (sometimes the `.map` files are overwritten, sometimes the `.log` ones, and who knows what comes next).

PR #257, opened 2017-12-13
Use themes for process flow focus label breakdowns.

ImogenF:


packages/visualizations/src/ProcessFlow/styles.ts
```diff
@@ -14,17 +15,16 @@ const breakdownsContainerStyle = {
...
-  paddingBottom: "10px"
+  float: "left"
 }
 
 const breakdownLabelStyle = {
   display: "block",
-  marginBottom: "3px",
-  fontSize: "12px"
+  marginBottom: theme.spacing / 4,
+  fontSize: theme.typography.small.fontSize
```
peterszerzo: I'd recommend spreading out the entire `...theme.typography.small` in the first line of the object - will be more future-proof if we add other font rendering attributes into the theme. If some spacing/positioning property doesn't work, you can of course override.

packages/visualizations/src/ProcessFlow/styles.ts
```diff
@@ -34,28 +34,30 @@ const breakdownCommentLabelStyle = {
...
   pointerEvents: "none",
-  backgroundColor: "#1499CE"
+  backgroundColor: setBrightness(theme.colors.palette.info, 145)
 }
 
 const breakdownTextStyle = {
-  color: "white",
-  position: "relative"
+  color: theme.colors.palette.grey70,
+  fontSize: 12,
```
peterszerzo: Use a `...typography.something` style (see above)

packages/visualizations/package.json
```diff
@@ -22,6 +22,8 @@
     "ci": "npm run test && npm run lint && npm run prettier && npm run package"
   },
   "dependencies": {
+    "@operational/theme": "0.1.0-1",
+    "@operational/utils": "0.1.0-1",
```
peterszerzo: You need to use `contiamo-ui-theme@0.0.1-47` and `contiamo-ui-utils@0.0.1-47` for now.

ImogenF: Even though `@operational/...` is used everywhere else, like in the showcase `package.json`?

peterszerzo: I'm being silly, of course! Everything is correct just as you implemented.

Heads up for the future, it will be `import { operational } from "@operational/theme"` for the default theme.

ImogenF: @peterszerzo but not yet, right?

peterszerzo: That's right, the PR is not even up yet.

TejasQ: Man, this collaboration is beautiful. 🔥😍

PR #259, opened 2017-12-15
Simplified showcase page with next.js

peterszerzo:
This is a light remake of the showcase page in `next.js`. Components, styles, page structure, and most of the code stays the same, however:
* routing code is removed (in favor of `next.js` routing simply dictated by the folder structure).
* build code is removed (in favor of the amazing `next.js` tooling).

### Motivation

Adding documentation should be fast, easy and reliable. Currently, creating a showcase page for a component involves a 6-step process involving at least 6 different files, with a bunch of possible typos along the way. This setup removes that overhead.


peterszerzo:
Abandoning this in favor of #262, which bundles together the next migration with the homepage+header+sidebar enhancements in consecutive commits. Easier to manage changes that way.

PR #264, opened 2018-01-04
Flatten repository. Run tests against public API

peterszerzo:
This PR implements code organization/quality improvements based on a wonderful [recent refactoring of React's codebase](https://reactjs.org/blog/2017/12/15/improving-the-repository-infrastructure.html). Specifically:
- codebase has been flattened: simply `./src/Button.tsx` as opposed to `./src/Button/Button.tsx`. If additional files are required in case a component becomes too big, create an additional `Button` directory, and add such files in there. Having to take this extra step should in fact discourage writing components that are too big (it is preferable to use local state containers, or shared styles from the `utils` package, coming soon.
- tests are now run against the public exports as opposed to source files, e.g. `import { Tabs } from "../index"`. This should make sure we never publish the package with a non-existent named `Tabs` export (which has happened a few times).


TejasQ:
Yeah, this diff is a _monster_ to review. Is the gist of it just that files were moved around? Or was code changed?

peterszerzo:
The diff is because of the snapshot files. No actual source code has been changed.

PR #266, opened 2018-01-05
Feature/animate logo

TejasQ:
Updated the Icon to have a top-level CSS rotation as opposed to an attribute rotation on the `g`, and set an interval for the rotation, as required a la [webpack](https://webpack.js.org/).

GitHub truncates the [Icons.js diff](https://github.com/Contiamo/operational-ui/compare/feature/animate-logo?expand=1#diff-0cdc44ba987ed4068aa88dc997ce9386) so please do not miss it.
  
a


TejasQ:
Will do.

TejasQ:
@peterszerzo – changes made!

PR #268, opened 2018-01-08
Remove plus chip

### LINKED ISSUE #263: Should we keep PlusChip?
I don't think it makes sense to keep the `PlusChip` component around, since it is mostly used with `Chip`'s, and its functionality can be incorporated into that component like the way `Filter`'s currently work:

screen shot 2018-01-03 at 10 24 32

We could extend the `Chip`'s to only display the symbols if a label is not present, and this would guarantee that the plus sign is always the same size as the neighboring chips.

peterszerzo:
Fixes #263. I also removed some `String(snippet)` code from the showcase code snippets, in a whim of reckless impatience..


TejasQ:
yeeeeeeah we really need to do something about these _large_ diffs

peterszerzo:
@TejasQ I agree about diffs - I'll try to do better!

PR #269, opened 2018-01-08
Breadcrumbs component

peterszerzo:
This PR formalizes the breadcrumbs already live on the showcase page:

screen shot 2018-01-08 at 11 20 40 am

The goal is to support any links (same way Sidenav does it), dropdowns using `<ContextMenu/>`, and whatever else comes up. Would be nice to encourage users to only use this component on the left-hand side of the header, which would give UI's a killer amount of consistency. Thoughts?


packages/components/src/Breadcrumbs.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+const Divider = glamorous.span(({ theme }: { theme: Theme }): {} => ({
+  display: "inline-block",
+  margin: "0 6px",
+  color: theme.colors.black
+}))
+
+export default (props: IProps) => (
+  <Container className={props.className} css={props.css}>
+    {(() => {
```
TejasQ: Please add some comments here documenting this a bit for newer devs who join us.

peterszerzo:
@TejasQ @micha-f is this a good direction for our breadcrumbs needs in client projects? (in case the context isn't clear, this element appears on the left-hand side of the header, right where the side navigation ends, and replaces the side navigation options we had before)

screen shot 2018-01-08 at 12 24 58 pm

TejasQ:
@peterszerzo, I think it works quite nicely.

micha-f:
@peterszerzo I think it's a good start. We might be able to tweak the styling. The dropdown positioning seems a bit off.

peterszerzo:
@micha-f you're absolutely right - I'll work out the positioning in a later commit.

package.json
```diff
@@ -16,7 +16,7 @@
     "test": "lerna run test",
     "lint": "lerna run lint",
     "prettier": "lerna run prettier",
-    "precommit": "lerna run precommit",
+    "precommit": "lerna run precommit && git add .",
```
TejasQ: I was THIS CLOSE to opening an issue about this 😂

peterszerzo:
The travis machines are systematically giving an `eval yarn` error. Ran identical `ci` checks locally, so it seems like a false alarm. Going ahead with the merge.

PR #273, opened 2018-01-09
Add visual test cases page to process flow

peterszerzo:
Visual test cases are re-added under /visualizations/process-flow/testcases, and linked to from the docs page (right under the first code playground). I put an emphasis on speed to get this working as quick as possible, will make it nicer looking later.

### DIFF
--- a/packages/showcase/components/Header.js
+++ b/packages/showcase/components/Header.js
@@ -8,8 +8,11 @@ export default props => {
   const pathnameChunks = props.pathname.split("/").filter(s => s !== "")
   const chunk1 = pathnameChunks[0]
   const chunk2 = pathnameChunks[1]
+  // Maintain a built-up version of the current breadcrumbs path so the links can point to
+  // e.g. /components, then /components/buttons, then /components/buttons/sublink
   let currentPath = ""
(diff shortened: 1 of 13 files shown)


packages/showcase/components/Header.js
```diff
@@ -6,21 +6,16 @@ import Link from "next/link"
...
-      label: props.pathmap[`/${chunk1}/${chunk2}`].query.title
-    })
-  }
+  let currentPath = ""
+  const breadcrumbs = pathnameChunks.map((chunk, index) => {
+    currentPath += `/${chunk}`
+    return {
+      url: currentPath,
+      label: props.pathmap[currentPath].query.title
+    }
```
TejasQ: What's the point of `let currentPath` here?

peterszerzo: `currentPath` builds up the paths incrementally, so that the first link in the breadcrumb can point to `/components`, the second to `/components/buttons`, etc. I'll document it a bit better first thing tomorrow.

TejasQ: AH I SEE SICK AMAZING GREAT JOB

PR #275, opened 2018-01-10
Showcase fixes

peterszerzo:
General improvements fixing flash of unstyled content, improving next.js code splitting, fixing a few layout issues.


peterszerzo:
No, it doesn't. I'd prefer having the import explicitly, just tried out this branch to see if it fixes the FOUC, then left it in by accident.

How would you prefer it?

TejasQ:
I love explicitness soooooo very much with you there.

peterszerzo:
Cool, I'll add it back!

PR #276, opened 2018-01-10
Feature-complete Auth component

peterszerzo:
This PR implements the API proposed in the previous Auth PR, which only handled input state. Status:
* [x] processing and error states
* [x] styling improvements 
* [x] documentation
* [x] review

(documentation is a bit sparse, pending component changes as we use it in production)


packages/blocks/src/Auth.tsx
```diff
@@ -29,80 +29,117 @@ const Container = glamorous.div(({ theme }: { theme: Theme }): {} => ({
...
+              <TitleType css={{ textAlign: "center", margin: 0 }}>{this.props.title}</TitleType>
+            ) : null}
+            {this.props.error ? <ErrorNotice>{this.props.error}</ErrorNotice> : null}
+            <InputFields>
+              {this.props.username ? (
+                <Input
+                  css={inputStyle}
+                  value={this.props.username}
+                  label="User name"
+                  onChange={(v: string) => {
```
TejasQ: Consider moving this function elsewhere and referencing it here for readability.

peterszerzo: I don't think isn't worth extracting and making the file look more scattered than it already is. Not for 6-line functions that do something so repetitive and predictable.

TejasQ: Okay.

packages/blocks/src/Auth.tsx
```diff
@@ -68,8 +69,11 @@ const inputStyle: {} = {
   margin: "20px 0"
 }
 
+const exists = (s?: string): boolean => !!s || s === ""
```
TejasQ: `s`? Please keep this a little less magical to save developers of mental overhead.

packages/blocks/src/Auth.tsx
```diff
@@ -68,8 +69,11 @@ const inputStyle: {} = {
...
+
 export default class Auth extends React.Component<IProps, IState> {
   render() {
+    console.log("is username", !!this.props.username)
```
TejasQ: no

peterszerzo: Appreciate the catch!

packages/blocks/src/Auth.tsx
peterszerzo: Will do

packages/blocks/src/Auth.tsx
```diff
@@ -69,11 +69,10 @@ const inputStyle: {} = {
   margin: "20px 0"
 }
 
-const exists = (s?: string): boolean => !!s || s === ""
+const exists = (stringValue?: string): boolean => !!stringValue || stringValue === ""
```
TejasQ: Sorry to be a pain, but can you add a tiny comment explaining the intent of this function?

peterszerzo: Nooo, I think that's a very good suggestion!

packages/blocks/src/Auth.tsx
```diff
@@ -69,7 +69,14 @@ const inputStyle: {} = {
...
 
-const exists = (stringValue?: string): boolean => !!stringValue || stringValue === ""
+/*
+ * Checks whether a string value exists or not (same as !!stringValue, except it also returns
+ * true for empty string. This is used especially often in this component because this component 
+ * renders form fields if any corresponding string value is passed down in props.
+ * e.g. props.username === null -> rendern no user name form field at all
+ *      props.username === "" -> render empty form field for user name
+ */
+const isStringValue = (stringValue?: string): boolean => !!stringValue || stringValue === ""
```
TejasQ: Ah, incredible 👌

micha-f: Typo "rendern"

peterszerzo: True story.. correct in German these days though, oder? :)

PR #280, opened 2018-01-11
Piecharts

ImogenF:


peterszerzo:
Looks like there's some typescript weirdness going on:

screen shot 2018-01-11 at 2 19 13 pm

ImogenF:
@peterszerzo I'll see if I can figure that out

packages/visualizations/src/PieChart/typings.ts
```diff
@@ -0,0 +1,67 @@
...
+
+export interface IDataAccessors {
+  data: (d: any) => any
+}
+export interface ISeriesAccessors {
+  name: (d: TDatum) => string
+  renderAs: (d: TDatum) => any
+}
+
+export interface IAccessorsObject {
```
peterszerzo: I wonder if we can make it clearer in the code that this is just a more specific version of `IAccessors` (the word object doesn't add any meaning, maybe a clunky-but-informative `IPieChartAccessors`). Is the generic `IAccessors` type necessary? If not, I'd take it out entirely and just call this one here `IAccessors`.

packages/visualizations/src/utils/event_catalog.ts
```diff
@@ -7,6 +7,14 @@ const Events = {
   },
   FOCUS: {
     CLEAR: "focus:clear",
+    COMPONENT: {
+      CLICK: "focus:component:click",
+      HOVER: "focus:component:hover",
+      OUT: "focus:component:out",
```
peterszerzo: What does `OUT` mean here, as opposed to `CLEAR`?

packages/visualizations/src/PieChart/renderers/abstract_renderer.ts
```diff
@@ -0,0 +1,395 @@
...
+  const p1: { x: number; y: number } = { x: initialX + 2, y: y(initialX + 2) }
+
+  // Solve for 0
+  const m: number = (p0.y - p1.y) / (p0.x - p1.x)
+  const xZero: number = -p0.y / m + p0.x
+
+  // Find nearest integer value for x that has y > 0
+  let xInt: number = Math.round(xZero)
+  let i: number
+  for (i = 0; i <= 10; i = i + 1) {
```
peterszerzo: `for (let i = 0; i <= 20; i = i + 1)`

packages/visualizations/src/PieChart/renderers/abstract_renderer.ts
```diff
@@ -0,0 +1,395 @@
...
+import { defaults, find, filter, forEach, isFunction, keys, map, reduce, values } from "lodash/fp"
+import * as d3 from "d3-selection"
+import "d3-transition"
+import { pie as d3Pie, arc as d3Arc } from "d3-shape"
+import { interpolateObject } from "d3-interpolate"
+import * as styles from "./styles"
+
+// y is a step-function (with two x values resulting in the same y value)
+// on the positive integer domain which is monotonic decreasing
+function approxZero(y: (x: number) => number, initialX: number): number {
```
peterszerzo: `const approxZero = (y: ...): number => {` - this makes it impossible to accidentally reassign the `approxZero` variable

packages/visualizations/src/PieChart/renderers/abstract_renderer.ts
```diff
@@ -0,0 +1,395 @@
...
+    if (y(xInt) <= 0) {
+      xInt = xInt - 1
+    }
+  }
+
+  return xInt
+}
+
+// Accessors of series in prepared data
+function dataKey(d: TDatum): string {
```
peterszerzo: See earlier comment in this file

packages/visualizations/src/PieChart/renderers/abstract_renderer.ts
```diff
@@ -0,0 +1,395 @@
...
+  return d.data.key
+}
+function dataLabelValue(d: TDatum): string {
```
peterszerzo: See earlier comment in this file

packages/visualizations/src/utils/component_focus.ts
```diff
@@ -0,0 +1,90 @@
...
+    el: TD3Selection,
+    events: IEvents,
+    payload: { component: TD3Selection; options: IObject }
+  ) {
+    this.state = state
+    this.el = el.append("xhtml:div").attr("class", `${styles.focusLegend} ${styles.componentFocus}`)
+    this.events = events
+    this.uid = uniqueId("componentFocusLabel")
+
+    const componentPosition: ClientRect = payload.component.node().getBoundingClientRect(),
```
peterszerzo: Can we have a separate `const` or `let` statement for each variable? I'm having a hard time scanning this code.

packages/visualizations/src/utils/drawing_canvas.ts
```diff
@@ -0,0 +1,179 @@
...
+  appendSeriesDrawingGroups(seriesElements: string[] | string[][]): void {
+    const series: TD3Selection = this.elements.drawing.append("svg:g").attr("class", "series-drawings-group")
+    this.elements.series = reduce((memo: IObject, se: string | string[]): IObject => {
+      if (isArray(se)) {
+        const renderer: string = se[0]
+        const clip: string = se[1]
+        memo[renderer] = series
+          .append("svg:g")
+          .attr("class", "series-" + renderer)
+          .attr("clip-path", "url(#" + this.prefixedId(`_${clip}`) + ")")
```
peterszerzo: You can nest templates, as in have another backtick template inside the `${}` of another backtick template. Mixing templates and simple string concatenation with `+` makes this read confusing.

packages/visualizations/src/utils/event_catalog.ts
peterszerzo: Or more broadly, I wonder if the nesting here confuses more than helps (`out` inside `component` inside `focus` seems to sandwich the name of an entity between names of modifiers). What if we kept them flat and also gave them a super rigid type definition so we get some extra compiler help for typos (unless that's already the case)?

ImogenF: `FOCUS.CLEAR` can be called from the outside to remove any focus that might be there. The component focus is a particular type of focus, so `OUT` is triggered specifically when the mouse leaves that component. Essentially though they do do the same thing. 

The auto-generated `event_catalog.d.ts` should do the job re: typos, right?

peterszerzo: Yes, you're right about the typings, happy to see it works with so little code :)

Wouldn't `HOVER -> MOUSEOVER` and `OUT -> MOUSEOUT` make it clearer that these events are related/parallel?

ImogenF: True, happy to rename to `mouseover` and `mouseout`. 

And yes, depending on the position of the label it may be that the mouse leaves the component/element but is still over the label, so we need an event to be triggered once the mouse leaves the label.

peterszerzo: Riiight, makes sense!

PR #286, opened 2018-01-16
Enforce and document single child for OperationalUI component

### LINKED ISSUE #284: <OperationalUI /> only accepts one child
`<OperationalUI />` only accepts one child, and cannot contain multiple children. **Please make this explicit in the README, and log a proper error if the following is attempted:**

```javascript
<OperationalUI>
    <div>foo</div>
    <MyComponent bar={baz} />
</OperationalUI>
```

peterszerzo:
Fixes #284


README.md
```diff
@@ -39,6 +39,8 @@ render(<MyInterface />, document.querySelector("#app"))
 
 Wrapping the interface inside an `OperationalUI` is required for the components to render properly - it does important setup work and allows for customization later (see [package docs](https://www.npmjs.com/package/%40operational%2Fcomponents) for details).
 
+Note that this component may only have a single child.
```
TejasQ: Perhaps bolder? **Note:** this component...

peterszerzo: I find it a little aggressive to yell in a readme.

TejasQ: Okay, but I'd rather be clear about gotchas so people don't lose heart wrestling with some error like us with hard-to-understand TS errors. That said, let's do it your way.

TejasQ: Also, I don't see an error message thrown in the diff?

peterszerzo: We're delegating that to React: https://reactjs.org/docs/react-api.html#reactchildrenonly, as in the error message is thrown automatically just by using `React.Child.only`. It is super clear, shows up in both JS and TS.

TejasQ: Okay, but I tried to implement it in `cookie-cutter-ui` and got an error I couldn't understand because it was swimming in an ocean that is TS' painful error log trace. _Just to humor me_, could you throw together a quick implementation of this and test the error for yourself?

TejasQ: It could very well be that I simply do not have the capacity to comprehend the clear error too. ¯\_(ツ)_/¯

packages/components/scripts/dev-server/site.tsx
```diff
@@ -23,6 +23,7 @@ class Site extends React.Component<{}, {}> {
   render() {
     return (
       <OperationalUI>
+      <p>adsf</p>
```
TejasQ: what is this

TejasQ: OH ITS DEV SERVER

TejasQ: nevermind 😓

PR #287, opened 2018-01-16
Move showcase header styles to components

### LINKED ISSUE #285: Standard header
As people pick up and learn Operational, a fun starter project could be to build a page that looks similar to the showcase in terms of layout: with a `Sidenav` and a `Header`.

What's really weird, is that the exported `Header` from `@operational/components` does _not_ look as expected out of the box: I would expect a proper header, as on the showcase, upon direct inclusion of `<Header />`.

Turns out, [it's styled specifically that way _just for the showcase_](https://github.com/Contiamo/operational-ui/blob/master/packages/showcase/components/Header.js#L24-L29). To me, this does not make sense. Can we adjust the exported `Header` component to be this way?

peterszerzo:
Fixes #285


packages/components/lib/Header.d.ts
```diff
@@ -3,9 +3,8 @@ import * as React from "react";
 export interface IProps {
     id?: string | number;
     className?: string;
-    css?: any;
+    css?: {};
```
TejasQ: [`glamorous` actually has a `CSSProperties` type that you can import and use here 😇 ](https://github.com/paypal/glamorous/blob/8fb386d4184e7f69455c9d8e54bfe0f8c0920aa1/typings/css-properties.d.ts#L1932)

packages/components/src/Header.tsx
```diff
@@ -6,30 +6,24 @@ import { hexOrColor, readableTextColor } from "@operational/utils"
 export interface IProps {
   id?: string | number
   className?: string
-  css?: any
+  css?: {}
```
TejasQ: and here

peterszerzo: @TejasQ that should be a different commit, as this is the current standard for components across the codebase.

TejasQ: ok

TejasQ: in that case (see next comment)

PR #291, opened 2018-01-17
Spinners/progress elements

### LINKED ISSUE #277: Spinner center has background that overlays poorly
In the image below (`Auth` block), the white background used to hide the inner part of the spinner's circle hides part of the underlying input field's border. This makes overlay spinners tricky to use.

screen shot 2018-01-10 at 4 46 28 pm

@TejasQ do you know a fix for it without changing how the spinner is made? Or shall I just go ahead and replace it with an SVG?

peterszerzo:
Fixes #277. Test like so:
`cd packages/showcase`
`yarn start`
`http://localhost:3000/components/loaders/


TejasQ:
Fun fact, `cd packages/showcase` isn't necessary because the top-level `yarn start` runs `cd` for you!

TejasQ:
I LOVE THAT THERE'S A PROGRESS BAR NOW!

packages/showcase/pages/components/loaders.js
```diff
@@ -0,0 +1,49 @@
+import * as React from "react"
+import { Card, CardHeader, Spinner, Progress } from "@operational/components"
+
+import Layout from "../../components/Layout"
+import Table from "../../components/PropsTable"
+import Playground from "../../components/Playground"
+
+const spinnerSnippet = `
+<Spinner/>
```
TejasQ: What props does this support?

peterszerzo: Just the standard `css` and `className` - I should make that clear for all components actually..

TejasQ: COOL I LOVE IT

packages/components/src/Progress.tsx
```diff
@@ -18,65 +19,39 @@ const Container = glamorous.div(
...
-const BarContainer = glamorous.div(
-  {
-    width: "100%",
-    height: "100%",
-    overflow: "hidden"
-  },
-  ({ theme }: { theme: Theme }) => ({
-    backgroundColor: theme.colors.gray30,
-    border: `1px solid ${theme.colors.gray20}`
+  ({ theme, fadeParent }: { theme: Theme; fadeParent: boolean }) => ({
```
TejasQ: what is fadeParent

packages/components/src/Spinner.tsx
```diff
@@ -8,81 +8,64 @@ export interface IProps {
...
-  <Container
-    key={props.id}
-    css={props.css}
-    className={props.className}
-    color={props.color}
-    spinnerSize={props.size}
-    spinDuration={props.spinDuration}
-  />
+  <Container key={props.id} css={props.css} className={props.className}>
+    <Element />
```
TejasQ: Can we rename Element to something more descriptive? 😄

packages/showcase/components/Layout.js
```diff
@@ -2,12 +2,15 @@ import glamorous, { ThemeProvider } from "glamorous"
...
 import { baseStylesheet, darken } from "@operational/utils"
-import { OperationalUI } from "@operational/components"
+import { OperationalUI, Progress } from "@operational/components"
+import Router from "next/router"
 
 import Header from "../components/Header"
 import Sidenavigation from "../components/Sidenavigation"
 import nextConfig from "../next.config"
 
+Router.onRouteChangeComplete = () => {}
```
TejasQ: ?

packages/components/src/Spinner.tsx
peterszerzo: Good call!

packages/showcase/components/Layout.js
peterszerzo: I'm removing it now, it was testing code. Thanks for the catch! :)

packages/components/src/Progress.tsx
peterszerzo: If you include `fadeParent`, it adds a near-opaque white cover over the direct parent to make it clear that the content is not accessible/dated/something new coming.

TejasQ: Nice! Is that a `prop`?

peterszerzo: Yes, it's documented in the Progress' propDescription - but I copied in my response right above to make it clearer.

packages/components/src/Spinner.tsx
```diff
@@ -23,22 +23,20 @@ const size = 30
 
 const Container = glamorous.div(({ theme }: { theme: Theme }): {} => ({
   label: "spinner",
-  margin: "auto",
   width: size,
-  height: size,
-  position: "relative",
-  transform: "translateZ(0)"
+  height: size
```
TejasQ: What is `size`?

peterszerzo: There a `const size = 30` a few lines up.

PR #300, opened 2018-01-26
Enhance/general

### LINKED ISSUE #297: Theme package README is outdated
It currently displays the use of `ThemeProvider` which doesn't exist anymore...

peterszerzo:
General enhancements (showcase website, documentation, minor component glitches), fixes #297.


packages/components/src/DatePicker/DatePicker.styles.ts
```diff
@@ -25,23 +25,32 @@ const Container: GlamorousComponent<{ isExpanded: boolean }, {}> = glamorous.div
   })
 )
 
-const Toggle: GlamorousComponent<{ onClick?: {} }, {}> = glamorous.div(({ theme }: { theme: Theme }): {} => ({
+export const Toggle: GlamorousComponent<{ onClick?: {} }, {}> = glamorous.div(({ theme }: { theme: Theme }): {} => ({
   position: "absolute",
-  bottom: 0,
+  top: 1,
```
TejasQ: waht

packages/components/src/DatePicker/DatePicker.styles.ts
```diff
@@ -25,23 +25,32 @@ const Container: GlamorousComponent<{ isExpanded: boolean }, {}> = glamorous.div
...
   cursor: "pointer",
-  right: 0,
-  width: inputHeight,
-  height: inputHeight,
+  right: 1,
```
TejasQ: Seriously, why?

peterszerzo:
@TejasQ the toggle element sits inside the borders of the input field, and should get a darker background on hover. Unfortunately, if I put `top: 0` and `bottom: 0`, the background will overlap the parent's borders, which looks terrible. Could use a `background-clip` property here, but since all dimensions are so strictly determined, I'd rather go with fewer styling primitives and slightly better browser support.

The layout/markup can probably be restructured to avoid this, but I'd leave that to a later point.

PR #305, opened 2018-01-30
Improve record component: naming, styling defaults, api

### LINKED ISSUE #299: Records improperly styled
wat

- There is no background color (a la `Card`s)
  - are they supposed to go inside a `<Card />`? This is not immediately clear to a consumer.

- Weird behavior with `RecordSummary` when its expanded (See GIF)
- Can we please rename `RecordSummary` -> `RecordHeader` to keep consistency with `Card`'s `CardHeader`?

peterszerzo:
Fixes #299


peterszerzo:
@TejasQ can we merge this and do further updates in a later PR, so we don't leave this hanging too long?

PR #310, opened 2018-02-09
Fix document click event unbind for context menus

peterszerzo:
- [x] remove `openOnHover` (menu can hide again while cursor is moved from the main body over to it, especially when the menu is repositioned - look for a more robust solution in a different commit)
- [x] add `open`, `onClick` and `onOutsideClick` props to allow for a controlled override of the internal open state. This makes the component usable as an autocomplete suggestions component.


packages/components/src/ContextMenu.tsx
```diff
@@ -41,6 +41,11 @@ export default class ContextMenu extends React.Component<IProps, IState> {
     isOpen: false
   }
 
+  constructor(props: IProps) {
```
TejasQ: 🤔 

Using class properties syntax might work better here, allowing skipping the constructor declaration altogether.

PR #312, opened 2018-02-13
Visualization color palettes

peterszerzo:


packages/visualizations/src/utils/VisualizationWrapper.tsx
```diff
@@ -25,6 +31,8 @@ class VisualizationWrapper extends React.Component<IProps, {}> {
   }
 
   componentDidMount() {
+    // Work with the theme here
+    console.log(this.props.theme)
```
micha-f: I assume this isn't just a leftover but was added on purpose?

peterszerzo: @micha-f thanks for the catch - yes, that's what is was for, but then we decided to do this in a different commit. I'll remove the console log and replace it with a `@todo` mark

PR #313, opened 2018-02-15
Sunburst improvements.

ImogenF:
- Added `suppressAnimation` config option to simplify rendering in case of very large datasets.
- Added arrows to indicate if a path is not fully displayed (unfortunately not clickable)
- Optional value assignment: any node not given a value is assigned a value equal to the sum of its children.
- Refactoring / tidying of code.
- Some style improvements.


micha-f:
Looks great!

In the showcase the breadcrumb looks different (no backgrounds) and the arrows are not showing up:

image

ImogenF:
@micha-f can you double-check the breadcrumbs please? Fixing the breadcrumbs and adding the arrows were 2 of the main things I focussed on, so they definitely should be working!

peterszerzo:
@ImogenF looks great: here is my visual feedback:
- the arrows are a little too close to the edge of the visualization. I think it'd look better with a more geometric arrow shape of a uniform 1px width, closing in at 90 degrees. To elude to ellipses a bit more, I tried a little sketch of three arrows below - a little cheesy and the proportions could use a bit more work, but maybe it helps comprehension?  
- without being aware of the technical hell involved, clickability and a hover effect on the arrow are important. Naive thought: can the arrow be rendered in the same SVG group as the last arc, making this a bit easier?
- breadcrumb elements that are truncated with an ellipsis need a way to show the full thing. I don't know if it works for SVG elements, but a `<div title="LongThing">LongTh...</div>` is a good way to add an expand hover for cheap.
- `suppressAnimation -> disableAnimations/noAnimations/animations=false/animate=false` to make it a bit easier to remember. 

screen shot 2018-02-19 at 10 48 39 am

micha-f:
@ImogenF Arrows and breadcrumbs look good 👍 . 
I'm looking forward to the color assignment so that our charts in the showcase look nicer ;)

ImogenF:
Thanks for the feedback.

@peterszerzo: 
- The arrows have been redesigned and are now clickable - they have the same hover behaviour as the element they relate to *except* showing focus labels. 
- The distance between the arrows and the viz is now configurable via the `arrowOffset` config option. 
- Breadcrumb items now show a tooltip on hover. Since the ellipsis truncation is done automatically, every item has a tooltip regardless of whether the whole label is visible already or not.

@micha-f Working on the colour assignment now, but in a different branch.

peterszerzo:
This is looking wonderful - my only last suggestion is a default `arrowOffset` of 10 - that would make it look the arrow a bit better spaced out.

PR #320, opened 2018-02-26
Implement automatic color assignment in sunburst charts.

ImogenF:


packages/utils/src/colorAssigner.ts
```diff
@@ -1 +1,53 @@
-// @todo - implement color assigner
+type usageCountType = {
```
peterszerzo: `UsageCountType` (pascal case)

packages/utils/src/colorAssigner.ts
```diff
@@ -1 +1,53 @@
...
+    return color
+  }
+
+  const assignColor = (key: string, color?: string): string => {
+    // @TODO color must be a hex
+    return addAssignment(key, color || nextColor())
+  }
+
+  return (key: string, color?: string): string => {
+    return getColor(key) || assignColor(key, color)
```
peterszerzo: What is the use-case for the optional color argument? It makes the API more complicated to understand, especially if we start adding parameters that customize the assignment logic later on.

peterszerzo: Technically, I can pass in colors that weren't part of the original palette right?

packages/visualizations/lib/Sunburst/typings.d.ts
```diff
@@ -23,7 +24,7 @@ export interface IAccessors {
         data: (d: IObject) => IObject;
     };
     series: {
-        color: (d: TDatum) => string;
+        color: (d: TDatum, colorAssigner: (key: string, color?: string) => string) => string;
```
peterszerzo: Does the config need to include a readily created color assigner? Is there a disadvantage to just accepting a color palette and instantiating an assigner within the visualization?

packages/utils/src/colorAssigner.ts
TejasQ: I think this ought to be an interface. As a general rule, anything with key/value pairs is recommended to be an interface for the main reason being that interfaces can be implemented, extended and merged whereas type literals serve simply as aliases to other type literals.

[More Info](https://www.briangonzalez.org/post/interface-types-vs-type-aliases-typescript)

packages/utils/src/colorAssigner.ts
ImogenF: You can pass in colours that aren't in the original palette, yes, but the main use case is that if some colours have been predefined in the input data, they are saved in the colorAssigner which prevents them from then being reused.

packages/visualizations/lib/Sunburst/typings.d.ts
ImogenF: The idea was that the colorAssigner can also be used in custom accessors. Otherwise the colorAssigner would need to be instantiated wherever the chart is defined (and wouldn't have access to the visualisation's default palette if the theme isn't passed in) if the use wanted to use it in the custom accessor. Happy to discuss this further though, I spent a while going around in circles with this...

packages/utils/src/colorAssigner.ts
peterszerzo: I agree - and that should just be swapping the word `type` with `interface`, and everything works the same way.

packages/visualizations/lib/Sunburst/typings.d.ts
micha-f: I'm not sure this is the nicest way to do things. I'd say either you use the default accessor (which uses a color assigner internally) or you provide your own accessor (using or not using a color assigner).

micha-f: `Accessors` should only take a datum and return a value

packages/utils/src/colorAssigner.ts
peterszerzo: I don't follow, what do you mean by predefined in the input data? Is that a different mechanism than accessors?

packages/visualizations/lib/Sunburst/typings.d.ts
peterszerzo: That sounds right to me as well.

ImogenF: The problem that I was struggling with is that if the palette is changed via the config, then you need to update the color accessor *only* if the default accessor is being used.

peterszerzo: In that case, I think it's safe to just check for that as `whateverAccessorWindsUpDominating === defaultAccessor`.

peterszerzo: As in, yes, I agree, that sounds a bit weird, but I think it's better to do weird things in the implementation than in the public API.

ImogenF: Ok, I'll see if I can find a nicer way to do this

packages/utils/src/colorAssigner.ts
peterszerzo: I would be inclined to say that if that's what the user wants, they should be on their own - have them manage color assignment outside somewhere, and pass results down to both charts. Otherwise, we'd be encouraging some kind of semiautomatic behavior that's hard to wrap one's head around `=>` bugs that the library gets blamed for.

packages/visualizations/lib/Sunburst/typings.d.ts
TejasQ: Casual reminder that we ought to suggest nicer ways to do things when we say there's a nicer way to do things. _I know we do this_ but just making sure it stays a habit.

ImogenF:
@peterszerzo @micha-f I've refactored the color accessor and removed the optional `color` parameter from the colorAssigner. Would appreciate any further feedback you might have!

micha-f:
@ImogenF Looks good!

PR #323, opened 2018-02-27
Fix sunburst breadcrumb.

ImogenF:


micha-f:
This works, LGTM @ImogenF

PR #324, opened 2018-02-27
Add dependencies to package.json

bmwilly:
Additional dependencies after building with
```
npm install @operational/components react react-dom
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,5 +29,10 @@
   },
   "workspaces": [
     "packages/*"
-  ]
+  ],
+  "dependencies": {
+    "@operational/components": "0.1.0-13",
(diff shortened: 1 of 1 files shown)


bmwilly:
not sure why I can't add reviewers, so paging @peterszerzo @TejasQ

peterszerzo:
Ideally, the root `package.json` should not have these dependencies, they should be in `packages/*/package.json` instead. I can see how it made starting out with the library easier, but they will start going out of sync and confuse contributors. I think we can improve your onboarding experience without duplicating dependencies.

TejasQ:
Peter is absolutely right.

PR #329, opened 2018-03-01
Refactor pie chart accessors, add color assigner.

ImogenF:


packages/visualizations/src/PieChart/focus.ts
```diff
@@ -3,43 +3,38 @@ import AbstractDrawingFocus from "../utils/focus"
...
+      .text(payload.d.key)
 
     content
       .append("xhtml:li")
       .attr("class", "series")
       .html(
         '<span class="value">' +
-          dataValue(datum) +
+          payload.d.value +
           '</span> \
```
peterszerzo: This string generation looks a bit brittle, I don't even understand what the `\` does or in which browser it would get quirky. Can we try to see if it works with a backtick syntax, something like this:
```
.html(`
  <span class="value">${payload.d.value}</span>
  <span class="percentage">(${percentageString(payload.d.percentage)})</span>
`)
```

packages/visualizations/src/PieChart/renderers/abstract_renderer.ts
```diff
@@ -30,12 +31,8 @@ const approxZero = (y: (x: number) => number, initialX: number): number => {
   return xInt
 }
 
-// Accessors of series in prepared data
-const dataKey = (d: TDatum): string => {
-  return d.data.key
-}
-const dataLabelValue = (d: TDatum): string => {
-  return d.data.percentage ? d.data.percentage.toFixed(1) + "%" : undefined
+const percentageString = (d: TDatum): string => {
```
peterszerzo: Can this return an explicit null if the percentage isn't available? Otherwise, I could see it silently producing non-sensical labels, since I'd pick up on `Percentage: null` more easily than `Percentage:      `.

packages/visualizations/src/PieChart/renderers/gauge.ts
```diff
@@ -177,21 +177,21 @@ class Gauge extends AbstractRenderer {
...
-      )
+    const range: number[] = map((value: number): number => value + Math.PI / 2)(this.angleRange())
+    const angle = (d: TDatum): number =>
+      d3ScaleLinear()
+        .range(range)
+        .domain([0, this.target])(d.value)
+    const xOuter = (d: TDatum): number => -d.r * Math.cos(angle(d))
+    const yOuter = (d: TDatum): number => -d.r * Math.sin(angle(d))
+    const xInner = (d: TDatum): number => -d.inner * Math.cos(angle(d))
+    const yInner = (d: TDatum): number => -d.inner * Math.sin(angle(d))
```
peterszerzo: Super cool trigonometry!!! :)

packages/visualizations/src/PieChart/renderers/abstract_renderer.ts
peterszerzo: The type signature needs an update to `string | null`

PR #338, opened 2018-03-02
Simplify build scripts. Tests on CI

### LINKED ISSUE #331: Improve CI scripts
- We should run all tests on CI
- We should run lint and prettier without --fix and fail
=> that's what CI is for

peterszerzo:
Fixes #331, like so:
* packages define their own `lint`, `prettier`, `test` and `package`. In order to restore some readability, they don't define their own `precommit`. Due to the lack of usage, removing `test:watch`.
* at the root level, there is a `ci:local` script that runs the above scripts for each package. This script is run on precommit, with all its changes added to the commit.
* on Travis (tests now working), the script is run again, and `npm run ensure:clean:repo` makes sure nothing changes. If a snapshot changes or a semicolon is added back by a script compared to the version of the code that was pushed, everything fails.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,11 +16,11 @@
     "test": "lerna run test",
     "lint": "lerna run lint",
     "prettier": "lerna run prettier",
-    "precommit": "lerna run precommit && git add .",
+    "precommit": "npm run ci:local && git add .",
     "package": "lerna run package",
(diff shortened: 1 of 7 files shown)


micha-f:
Awesome! 👍

PR #339, opened 2018-03-05
Fix pie chart percentage string.

ImogenF:


ImogenF:
@peterszerzo I see changes in the compiled .js too... the new commit has 3 changed files, right?

peterszerzo:
Oh never mind, it was hidden because it was too big - I checked it and all is well!

PR #340, opened 2018-03-05
New showcase layout

peterszerzo:
Showcase is getting simpler, more maintainable and mobile-friendly. The steps:
* [x] docs pages to Markdown
* [x] single-page custom website
* [x] single-page visual test case website


packages/showcase/components/Layout.js
```diff
@@ -30,15 +30,13 @@ const Content = glamorous.div({
   }
 })
 
-const PageContent = glamorous.div(({ theme }) => ({
+const PageContent = glamorous.div(({ theme }: { theme: Theme }): {} => ({
   display: "flex",
   alignItems: "flex-start",
+  position: "relative",
   flex: "1",
   padding: 16,
```
TejasQ: ? `16`? Why not `theme.spacing`?

packages/showcase/components/Layout.js
```diff
@@ -78,6 +76,20 @@ const PageContent = glamorous.div(({ theme }) => ({
   }
 }))
 
+const SidePanels = glamorous.div(({ theme }: { theme: Theme }): {} => ({
+  position: "absolute",
+  right: 16,
```
TejasQ: ? `16`? Why not `theme.spacing`?

packages/showcase/components/Layout.js
```diff
@@ -78,6 +76,20 @@ const PageContent = glamorous.div(({ theme }) => ({
...
+  bottom: 16,
+  top: 52 + 16,
+  width: 400 - 16,
+  "& > *": {
+    height: "calc(50% - 8px)"
+  },
+  "& > *:not(:first-child)": {
+    marginTop: 16
```
TejasQ: ? `16`? Why not `theme.spacing`?

TejasQ:
As discussed on Slack, I'm not sure the panels on the right would be big enough to contain the code playgrounds.

packages/showcase/components/Layout.js
peterszerzo: @TejasQ Because we haven't agreed on spacing yet, whether it'll be 16 or 12. Intelligent agent view has 12, I'd argue it looks a bit better. I'll bring it up for discussion as soon as I work this out in a bit more detail.

peterszerzo:
@TejasQ they won't have code snippets, they'll just have a showcase of finished uses of the component. Code samples will stay where they are.

packages/showcase/components/Layout.js
TejasQ: Can we still not have it as a variable right now whatever the final decision will be so it'll be easier to change later? 🤔

peterszerzo: I agree, it's a better way to go - I'll do that!

packages/showcase/components/Layout.js
```diff
@@ -103,6 +103,7 @@ export default class Layout extends React.Component {
 
   render() {
     const { pathname } = this.props
+    const children = React.Children.toArray(this.props.children)
```
TejasQ: Why? What's happening here? Please leave comments next time when writing something like this.

peterszerzo: The idea is to use the first child of the layout component as main content, the second as top-right panel, the third as bottom-right panel. I'll add this as comments.

TejasQ: Ummmm CSS grid does this automatically... would it be worth looking into that?

peterszerzo: You're totally right, I'll do that!

packages/showcase/components/Layout.js
```diff
@@ -69,7 +69,7 @@ const Grid = glamorous.div(({ theme }: { theme: Theme }): {} => ({
     gridRowStart: 1,
     gridRowEnd: "span 1"
   },
-  "& > :nth-child(2)": {
+  "& > :nth-child(3)": {
```
TejasQ: Can we please have a comment here explaining _what_ the 3rd child is?

peterszerzo:
@TejasQ no, it doesn't compile. @ImogenF, if you point me in the right direction I can batch update the files. The visual test website is almost finished :)

ImogenF:
@TejasQ `viz.data` expects data in whichever format the given visualization needs it, so it is not necessarily an array of `Datum`s. So for the visualizations we currently have:

Process flows expect the form ```{
  journeys: Journey[]
  nodes: Node[]
}```

Sunbursts expect ```{
  name: string
  value: number
  children?: RawData[]
}```

Pie charts expect `Datum[]`

Additionally, any of these data structures can theoretically be nested inside some other array or object, and the visualization is told how to get to this data structure using the data accessor.

So there is no standard input for the data method at all, unfortunately...

TejasQ:
Hmmm. I think the code for the test cases ought not be lax here. I think strongly typing the data that we pass into `viz.data(THIS_STUFF)` is quite essential.

peterszerzo:
@TejasQ I technically agree, but that will push us down the rabbithole of exporting a bunch of data interfaces from the visualizations package (which currently aren't exported). Are we sure we want that extra formality? Can't the users just supply data and get help from the compiler automatically? (of course, it never works that well and that automatically, but still, there is something in that simplicity)

TejasQ:
@peterszerzo I think you misunderstand me: I don't mean that the visualizations ought to strongly type their inputs for the `.data()` call _right now_, though in the future, this would be amazing.

What I _am_ suggesting is that the files in `./packages/visual-tests` play nice with the compiler for now without using `any` is all.

ImogenF:
@TejasQ The visualizations do have strongly-typed data inputs...

TejasQ:
@ImogenF ah, I see. Wonderful! Then I guess what I'm asking for is that the files in `./packages/visual-tests` call `viz.data()` with the right arguments for their visualization.

ImogenF:
@TejasQ isn't that then what @peterszerzo was saying about "the rabbithole of exporting a bunch of data interfaces from the visualizations package"

TejasQ:
@ImogenF Not to my understanding, because the visualizations package can remain untouched and ought not have anything exported from it. Rather, the change I think that would make the most sense, that I have failed to correctly articulate so far (sorry!) would be:

```js
const data = {
  data: {
    journeys: [],
    nodes: [],
  }
  width: 200
}

// later...
viz.data(data); //<- incorrect AFAIK
```

the way it is right now in this PR, would become 

```js
const data = {
  data: {
    journeys: [],
    nodes: [],
  }
  width: 200
}

//later
viz.data(data.data) // <- not the most readable, but should work
```

ImogenF:
@TejasQ Neither of the above options *should* make TS complain from within the visualization, as long as the data accessor is set correctly - it should only complain if the data returned by the accessor is not in the correct format.

TejasQ:
Hmmm. Maybe it was the data accessor then. What do you think, @peterszerzo?

peterszerzo:
@TejasQ there were a few minor issues in the typings of the data assignment method, which the test cases helped fix - so everything compiles now.

Data assignment remains dynamic for now, but we can always revisit it at some point later in time.

packages/visual-tests/src/components/Marathon.tsx
```diff
@@ -1,39 +1,77 @@
...
 
-class Marathon extends React.Component {
+class Marathon extends React.Component<Props, State> {
   static defaultProps = {
     timeout: 0
   }
 
   state = {
-    tests: [],
+    tests: [] as ITest[],
```
TejasQ: Why the assertion here? Would `state: State` on the preceding line not work?

packages/visual-tests/src/components/Marathon.tsx
```diff
@@ -46,22 +84,28 @@ class Marathon extends React.Component {
     })
   }
 
-  test = (description, fn) => {
+  test = (description: string, fn: (done?: ((a: any) => void)) => void): void => {
     this._tests.push({ description, fn })
   }
 
-  expect = actual => {
+  expect = (actual: any): { toBe: any } => {
```
TejasQ: Is there a plan to type this stronger later?

packages/visual-tests/src/components/Marathon.tsx
```diff
@@ -1,39 +1,77 @@
...
     id: 0
   }
 
-  _tests = []
+  container: HTMLElement
 
-  setStateById = (updater, ignoreId) => {
+  private _tests: ITestWithRunner[] = []
+
+  setStateById = (updater: (prevState: State, props: Props) => { id: number }, ignoreId?: boolean): Promise<void> => {
```
TejasQ: Some documentation of what this does in JSDoc style would be nice

packages/visual-tests/src/components/Marathon.tsx
```diff
@@ -74,7 +118,7 @@ class Marathon extends React.Component {
     const currentTestId = this.state.id
 
     if (test.fn.length === 0) {
-      await sleep(timeout)
+      await sleep(timeout as any)
```
TejasQ: ?

packages/visual-tests/src/components/Marathon.tsx
```diff
@@ -74,7 +118,7 @@ class Marathon extends React.Component {
     const currentTestId = this.state.id
```
TejasQ: Since you destructure on line 110, why not use destructured `id` here?

packages/visual-tests/src/components/Marathon.tsx
```diff
@@ -143,13 +185,13 @@ class Marathon extends React.Component {
     this.startTests()
   }
 
-  componentDidUpdate(prevProps, prevState) {
+  componentDidUpdate(prevProps: Props, prevState: State) {
     if (prevProps.test !== this.props.test) {
```
TejasQ: Definitely need more doc here.

packages/visual-tests/src/components/Marathon.tsx
```diff
@@ -1,39 +1,77 @@
 import * as React from "react"
```
TejasQ: This whole file seems a little too magical. Can we please add hefty documentation where relevant?

packages/visual-tests/src/components/MarathonTestResults.tsx
```diff
@@ -20,7 +19,7 @@ const Item = glamorous.li({
   }
 })
 
-const Title = glamorous.p(({ theme }) => ({
+const Title = glamorous.p(({ theme }: { theme: Theme }): any => ({
```
TejasQ: It outputs `CSSProperties`.

packages/visual-tests/src/index.tsx
```diff
@@ -0,0 +1,3 @@
+import Marathon from "./components/Marathon"
+
+console.log(Marathon)
```
TejasQ: ?

packages/visual-tests/src/App.tsx
```diff
@@ -0,0 +1,70 @@
...
+import glamorous from "glamorous"
+import { Sunburst, VisualizationWrapper } from "@operational/visualizations"
+import { Theme } from "@operational/theme"
+import { OperationalUI, Button, Card, CardHeader, Grid } from "@operational/components"
+
+import Marathon, { IMarathon } from "./components/Marathon"
+
+import allTestCases from "./TestCases"
+
+const testcases = allTestCases[2]
```
TejasQ: What?

packages/visual-tests/src/App.tsx
```diff
@@ -0,0 +1,70 @@
...
+                      case: index
+                    }))
+                  }, 50)
+                }}
+              >
+                {testcase.title}
+              </TestToggle>
+            ))}
+            {this.state.case || this.state.case === 0 ? (
+              <Marathon test={testcases[this.state.case].marathon as ((a: IMarathon) => void)} timeout={2000} />
```
TejasQ: Why the assertion?

packages/visual-tests/src/TestCases/PieChart/gauge-3.ts
```diff
@@ -1,7 +1,5 @@
-import Layout from "../../../../components/Layout"
 import { PieChart } from "@operational/visualizations"
-import Marathon from "../../../../components/Marathon"
-import { Card, CardHeader } from "@operational/components"
+import { IMarathon } from "../../components/Marathon"
```
TejasQ: We prefix with `I` again?

packages/website/src/components/StaticContent.tsx
```diff
@@ -13,6 +13,9 @@ interface Props {
 
 const Container = glamorous.div(({ theme }: { theme: Theme }): {} => ({
   label: "showcasestaticcontent",
```
TejasQ: Maybe `showcase-static-content`?

packages/website/src/components/StaticContent.tsx
```diff
@@ -13,6 +13,9 @@ interface Props {
...
+  fontSize: "24px",
```
TejasQ: This can be a number `24`.

packages/website/src/index.tsx
```diff
@@ -31,13 +31,13 @@ const TitleBar = glamorous.div(({ theme }: { theme: Theme }): {} => ({
   "& h1": {
     ...theme.typography.title,
     fontWeight: 400,
-    fontSize: "2.5rem",
+    fontSize: "36px",
```
TejasQ: Like, all these can be numbers...

packages/website/src/index.tsx
```diff
@@ -113,11 +140,39 @@ export default class App extends React.Component<{}, {}> {
...
+          </Card>
+          <Card>
+            <CardHeader css={customCardHeaderStyles}>Examples</CardHeader>
+            <StaticContent
+              markdownContent={`
+We have some CodeSandbox demos for [simple interface elements](https://codesandbox.io/embed/zq5nm42x84), a [full-featured UI]() and [ visualizations]().
+              `}
+            />
+          </Card>
+          <Card>
```
TejasQ: I think we don't need this section.

packages/visual-tests/src/components/Marathon.tsx
peterszerzo: I'd estimate typing and documenting this file to be a daylong project - and that for something that is internal only, and that has been working reliably over the past few months. Worth it you think?

peterszerzo: (I dug out the TS version of this file from January, no modifications)

TejasQ: @peterszerzo I think at the very least we need to have an open issue for it. As far as I know, there isn't a nice visual test runner like this in the OSS world, so we _could_ factor it out into a nice contribution to the community later.

peterszerzo: Makes sense - just opened #367.

PR #341, opened 2018-03-05
Add automatic node label positioning to process flow.

ImogenF:


packages/showcase/pages/visualizations/process-flow/index.js
```diff
@@ -290,8 +290,8 @@ const propDescription = {
 
     {
       name: "labelPosition",
-      description: "Node label position",
-      defaultValue: 'd => d.labelPosition || "right"',
+      description: 'Node label position - "auto", "top", "bottom", "left", "right" or "middle"',
+      defaultValue: 'd => d.labelPosition || "auto"',
```
micha-f: In the facade, the default is still `labelPosition: (d: INodeAttrs): string => d.labelPosition || "right"`

ImogenF: @micha-f good spot - have fixed.

PR #342, opened 2018-03-05
Simple idiomatic input

### LINKED ISSUE #321: <Input /> is wrapped in a div
Is this intended behavior? It could be a little bit surprising to consumers: "I want an input, but why is there automatically a _forced div around it?!"_

Is this something we can/should do differently?

peterszerzo:
Fixes #321, a first of three commits aimed at making input fields simpler to style.
* [x] remove `withLabel` HOC - instead, move label styling into `src/utils/mixins`, to be used across Select, DatePicker and Checkbox (separate commits).
* [x] conditionally render `<label>` tag.
* [x] `css` and `className` attributes should work differently whether there's a label or not. Handle this in a maintainable way.


packages/components/src/Input.tsx
```diff
@@ -24,16 +21,11 @@ export interface Props {
...
-    marginBottom: theme.spacing / 4
-  }
-}))
+const Label = glamorous.label(mixins.label)
 
-const InputField = glamorous.input(({ theme, disabled }: { theme: Theme; disabled: boolean }) => ({
-  label: "inputfield",
+const InputField = glamorous.input(({ theme, disabled }: { theme: Theme; disabled: boolean }): {} => ({
+  ...theme.typography.body,
+  label: "input",
```
TejasQ: There's a CSS property `label`?

peterszerzo: @TejasQ that's a glamorous prop that adds the string as part of the class name in development mode. Makes components easier to find if you don't use react devtools.

packages/components/src/Input.tsx
```diff
@@ -66,6 +60,15 @@ const Input = (props: Props) => {
       }}
     />
   )
+  if (props.label) {
+    return (
+      <Label htmlFor={domId}>
```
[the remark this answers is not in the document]
peterszerzo: @TejasQ we should still have it according to WCAG accessibility specs. With that in mind, I am tempted to add an `inputId` prop that developers can use to set up the `for - id` wiring, and make it their responsibility. I don't think we can use `id` for it, since it means container id for all other components, and people will probably start getting used to that.

What do you think?

packages/components/src/Input.tsx
TejasQ: That is _so cool_. TIL. Thank you!

packages/components/src/Input.tsx
```diff
@@ -67,9 +62,9 @@ const Input = (props: Props) => {
   }
   if (props.label) {
     return (
-      <Label htmlFor={domId} css={props.css} className={props.className} key={props.id}>
-        {props.label}
-        <InputField {...commonInputProps} key={props.id} />
+      <Label htmlFor={forAttributeId} css={props.css} className={props.className} key={props.id}>
+        <span>{props.label}</span>
```
TejasQ: Why the span?

packages/components/src/utils/mixins.ts
```diff
@@ -8,15 +10,15 @@ export const inputFocus = ({ theme }: { theme: Theme }): {} => ({
...
-    marginBottom: theme.spacing / 4,
-    ...theme.typography.small,
-    fontWeight: 600,
-    opacity: 0.7
-  }
-})
+  minWidth: 240
+}))
+
+export const LabelText = glamorous.span(({ theme }: { theme: Theme }): {} => ({
```
TejasQ: Can a `Label` or a `LabelText` be used without an `Input`?

peterszerzo: @TejasQ only internally - they will be re-used for `Select`, `DatePicker` and `Checkbox`.

PR #346, opened 2018-03-08
Explicit button typography

peterszerzo:
Sets typography styles explicitly on buttons. This ensures that nesting a button inside e.g. a card header doesn't make it inherit crazy heading styles.


TejasQ:
Needs rebase!

PR #347, opened 2018-03-08
Improve card headers

peterszerzo:
This PR improves card headers like the image below, with the following goals:
* replacing bottom border with a background color, making them easier to recognize and look a bit sharper (and it's all the design craze right now). To make this work, the shadow bumps up 1px in blur.
* we have been moving away from bold text, formalized by going `font-weight: 600 -> 400` for all first-level headings, and adding an opacity to second-level headings while keeping font size and weight the same.

screen shot 2018-03-08 at 2 43 19 pm


packages/components/src/CardHeader.tsx
```diff
@@ -14,13 +14,13 @@ const Container = glamorous.div(({ theme }: { theme: Theme }): any => ({
   label: "cardheader",
   display: "flex",
   alignItems: "center",
+  backgroundColor: theme.colors.cardHeaderBackground,
```
TejasQ: Why the specific one color just for this? 🤔

peterszerzo: If there is general body `background` and `cardBackground` in the theme, I think this needs to be in there too - the contrast between them is super important. Unless we mandate card backgrounds to be white and automatically lighten `background` by a specific amount to get this, which would keep the theme nice and simple. What do you think?

TejasQ: Sounds like you've thought it through. Let's do it.

PR #348, opened 2018-03-08
Improve OperationalUI wrapper

peterszerzo:
* `<OperationalUI withBaseStyles />` now injects a base stylesheet that takes care of any styling setup required for a new operational UI: body height+margin+background, global font family+size definitions, global link colors, TBD. Together with the future `Layout` component, getting started on a new UI should be a ten-liner, with everything we need self-contained in the app entry point
* no more single-child requirement with `React.Fragment`


peterszerzo:
@TejasQ @micha-f having doubts on finding a good name for `withBaseStyles`. Maybe it should be `withBaseStylesheet` because it is a stylesheet and not some inline styles? Or `withNormalize`, since it's kind of like normalize?

TejasQ:
Let's cross the name bridge when we come to it.

peterszerzo:
Cool - the name is good enough for now :)

PR #349, opened 2018-03-08
Opinionated grids

peterszerzo:
The props in the previous grid implementation set out to mimic the CSS grid layout properties and pass them along to the container. This would look exactly the same if one put the same props on a glamorous component, `=>` a useless abstraction originally intended to also include an older browser polyfill, which is too big a task anyway.

Instead, grids should concisely express common layouts like `<Grid type="3x2" />` for a standard 3x2 grid, and `<Grid type="IDE" />` for the layout of an in-browser code editor (this one can have more features such as accommodating an optional pop-up console). If one wants something else, the grid layout is simple enough to use. We could include older-browser-compatible versions of these later on, which are much simpler to implement for specific layouts as opposed to any grid definition.


packages/components/src/Grid.tsx
```diff
@@ -1,54 +1,50 @@
...
+  return {
+    label: "Grid",
+    width: "100%",
+    height: "100%",
+    display: "grid",
+    padding: theme.spacing,
+    gridColumnGap: theme.spacing,
+    gridRowGap: theme.spacing,
+    ...(() => {
+      switch (gridType) {
```
TejasQ: Are we sure we want to do it this way style-wise as opposed to making a function somewhere above called `getGridType()` or something?

peterszerzo: Extracting is a great idea here!

PR #350, opened 2018-03-09
Layout component

peterszerzo:
Layout component + some changes:
* remove `width` prop for side navigation (header doesn't have it either). Moving these dimensions into `./src/constants.ts`.
* add `loading` prop, including a logglable `<Progress />` component inside the implementation of this layout component (to seep into other components as well).


TejasQ:
Needs rebase.

PR #351, opened 2018-03-12
Improve table styling

peterszerzo:
Improve table styling per image below:
* remove bold from headings, lighten text for more emphasis on the content.
* min-width to make it look a bit less sad.
* remove `gray10 | gray20` color references for named ones (we'll start doing this across the codebase).

screen shot 2018-03-12 at 9 59 35 am


packages/components/src/Table.tsx
```diff
@@ -12,24 +12,32 @@ export interface Props {
...
+  },
+  "& tr:first-child": {
+    borderTop: 0
+  },
+  "& tbody tr:last-child": {
+    borderBottom: 0
   },
   "& td": {
-    border: "1px solid",
+    padding: "8px 16px",
```
TejasQ: Can/should we use theme spacing variables here?

peterszerzo: We probably should!

TejasQ:
Will merge when conflicts resolved.

PR #352, opened 2018-03-12
Progress component errors + reload

peterszerzo:
`<Progress>` now supports an error state per image below. Syntax looks like so:

```
<Card css={{ position: "relative" }}>
  <CardHeader>Hello</CardHeader>
  <Progress
    error={this.state.isError ? "Loading error" : null}
    onRetry={() => {
      this.setState(prevState => ({
        isError: null
      }))
      /* call some method to retry fetching data */
    }}
  />
</Card>
```

screen shot 2018-03-12 at 11 05 24 am


peterszerzo:
I'm starting to change my mind about whether we want to add a `loading` prop to `<Card>`, and pass that information through to a `<Progress>` component rendered internally. Adding functionality through specialized components is natural to the API (`<Card><CardHeader>Header</CardHeader></Card>` as opposed to `<Card header="Header" />`), there's nothing new to learn, and making sure `position: relative` is added to the card's styling is just not that big of a deal.

What do you think @micha-f @TejasQ?

TejasQ:
@peterszerzo I think a couple of things: I think composing components is very React and a good way to educate users through use of the library what compositions are possible and recommended: Progress inside Card being one of them.

I think hiding away implementation details to use Progress internally with a loading prop on Card also has its benefits (such as less boilerplate), but will not be as educational to our users, allowing the community to compose components in more interesting ways, you know?

Can we offer both? 🤔

packages/components/src/Progress.tsx
```diff
@@ -39,20 +43,61 @@ const fillProgress = css.keyframes({
...
+const Bar = glamorous.div(({ theme, isError }: { theme?: Theme; isError: boolean }) => ({
+  width: "100%",
+  height: 2,
+  backgroundColor: theme.colors.info,
+  ...isError
+    ? {
+        backgroundColor: theme.colors.error
+      }
+    : {
+        animation: `${fillProgress} cubic-bezier(0, 0.9, 0.26, 1) forwards 20s`
```
TejasQ: Can we have this entire animation, with timing function be a mixin and reused everywhere?

peterszerzo: I don't think this particular animation will be re-used anywhere, or do you?

TejasQ: Fair point. Okay! Let's leave it here.

peterszerzo:
@TejasQ for select components where adding a progress makes sense, I'd love to add both. Some questions:
* what is a syntax that is intuitive to use both ways? Does it make sense to do `<Card progress={/* same props a Progress component would take */} />`, so it is clearer what component is being rendered, and there's no new prop structure to keep in mind?
* how do we make this code simple enough to maintain, and when is the right time to implement everything comprehensively?

TejasQ:
@peterszerzo I'd like to have

## Composition style
```jsx
<Card>
  <Progress onRetry={() => blah()} />
  My card things
</Card>
```

and

## Terse style
```jsx
<Card progress={{onRetry: () => blah()}}>
  My card things
</Card>
```

Of course, I'd suggest the `progress` prop could also be a boolean that just shows progress for no apparent reason.

peterszerzo:
So to summarize, we support these three:

# Composition

```
<Card>
  <Progress onRetry={() => blah()} />
  My card things
</Card>
```

# Terse

Take the same props structure the progress component would expect.

```
<Card progress={{error: "1234", onRetry: () => blah()}}>
  My card things
</Card>
```

# Extra terse

Adds the most super extra basic progress component there is.

```
<Card progress />
```

Does that look right?

TejasQ:
@peterszerzo I'd like that. What do you think?

micha-f:
Do these three options result in the same look?

peterszerzo:
@micha-f yes, they do, as long as the card is positioned relative in the first example (composition). For the other ones, that positioning will be taken care of by the implementation.

micha-f:
Why do we need multiple ways of doing things then?
```
<Card>
  <Progress onRetry={() => blah()} />
  My card things
</Card>
```
Seems sufficient and simple. You can implement all the logic of when to display it yourself.

peterszerzo:
@micha-f it's a balance between how easily we want it to work out of the box, and how much extra code we want to add to support the props approach. If I were to write this just for myself, I wouldn't even think about adding props, but I'm not the only one using the library. @TejasQ why don't we just put off this decision, and focus on getting the library right with just the composition? I think it'll be clearer what's best, especially once I vamp up the documentation.

micha-f:
If we add composition AND props everywhere, we'll clutter everything. Isn't it easier to simply state "Operational UI generally prefers composition over props" and people can rely on that?

TejasQ:
@peterszerzo okay.

PR #354, opened 2018-03-12
Grid fixes

peterszerzo:
Grids defined as `<Grid type="2x3" />` are now guaranteed to have a equal column/row sizes. Also refactored the implementation to allow any `<Grid type="MxN" />` later on (for now, it just makes the implementation snappier and nicer to read).


packages/components/scripts/dev-server/site.tsx
```diff
@@ -41,7 +41,25 @@ class Site extends React.Component<{}, State> {
           <Sidenav />
           <Header />
           <Grid type="3x2">
-            <Card><CardHeader>Hello</CardHeader></Card>
+            <Card>
+              <CardHeader>Hello</CardHeader>
+              jalsfgfdlsjgfdslhgdshgsf
```
TejasQ: Perfect.

packages/components/src/Grid.tsx
```diff
@@ -10,28 +10,26 @@ export interface Props {
...
+  if (!isNaN(cols) && !isNaN(rows)) {
+    return {
+      gridTemplateColumns: `repeat(${cols}, 1fr)`,
+      gridTemplateRows: `repeat(${rows}, 1fr)`
+    }
+  }
+  // Because GridType is defines types that all fall into the
+  // cases above, this piece of code is never reached.
+  // (may change if we start accepting arbitrary NxM values)
+  return {}
```
TejasQ: This is really powerful.

TejasQ:
Will approve and merge when the conflicts are sorted.

TejasQ:
For the future, it might be worth not assuming a `1fr` sizing for each grid cell but instead allowing the user to customize that as well...

peterszerzo:
@TejasQ the intention here is to give a super simple one-liner grid, and telling the user that if they want anything else, they should implement it themselves using CSS grid layout props directly.

PR #356, opened 2018-03-13
Improve efficiency of process flow loop handler.

ImogenF:
Locally, this seems to have improved the speed of the unlooper by a factor of ~8.


TejasQ:
Thanks!

PR #357, opened 2018-03-13
Refactor visualisations.

ImogenF:
- Removed abstract classes.
- Added class interfaces.
- Improved general types and interfaces.


packages/visualizations/lib/PieChart/canvas.d.ts
```diff
@@ -1,9 +1,34 @@
-import AbstractDrawingCanvas from "../utils/drawing_canvas";
-import { TD3Selection, IState, TStateWriter, IEvents } from "./typings";
-declare class Canvas extends AbstractDrawingCanvas {
-    constructor(state: IState, stateWriter: TStateWriter, events: IEvents, context: Element);
-    createEl(): TD3Selection;
+import { Canvas, D3Selection, EventBus, Object, SeriesEl, State, StateWriter } from "./typings";
+declare class PieChartCanvas implements Canvas {
```
peterszerzo: I see three kinds of verbs being used here: `append`, `attach` and `insert`. Can we get away with two? Or even just `render` for all? This being a high-level overview, I don't think I want to know whether stuff is attached to a container next to some siblings or just rendered in their own container.

packages/visualizations/lib/PieChart/facade.d.ts
```diff
@@ -1,30 +1,30 @@
...
+    insertCanvas(): PieChartCanvas;
+    insertComponents(): Components;
     insertSeries(): Series;
-    data<T>(data?: T): T;
-    config(config?: Partial<IConfig>): IConfig;
-    accessors(type: string, accessors: IObject): IObject;
+    data(data?: Data): Data;
+    config(config?: Partial<PieChartConfig>): PieChartConfig;
+    accessors(type: string, accessors: Accessors<any>): Accessors<any>;
     on(event: string, handler: any): void;
```
peterszerzo: Is this for native dom events or for pre-defined event types? If the latter, can we have a `"EVENT1" | "EVENT2"` type here?

packages/visualizations/lib/PieChart/focus.d.ts
```diff
@@ -1,9 +1,15 @@
...
-declare class Focus extends AbstractDrawingFocus {
-    onElementHover(payload: {
-        focusPoint: IObject;
-        d: IObject;
-    }): void;
+import ComponentFocus from "../utils/component_focus";
+import { D3Selection, EventBus, Focus, HoverPayload, Object, SeriesEl, State, StateWriter } from "./typings";
+declare class PieChartFocus implements Focus {
+    el: SeriesEl;
+    componentFocus: ComponentFocus;
```
peterszerzo: What is element and what is component in this case?

packages/visualizations/lib/PieChart/legend.d.ts
```diff
@@ -1,11 +1,16 @@
...
+    events: EventBus;
+    legend: D3Selection;
+    state: State;
+    stateWriter: StateWriter;
+    constructor(state: State, stateWriter: StateWriter, events: EventBus, el: D3Selection);
+    draw(): void;
     updateComparisonLegend(): void;
-    currentOptions(datum: IObject): IObject;
+    data(): LegendDatum[];
+    onComponentHover(d: LegendDatum, el: HTMLElement): void;
```
peterszerzo: There was `onElementHover` in `focus`, does it make sense to have them consistent?

packages/visualizations/lib/PieChart/renderers/donut.d.ts
```diff
@@ -1,11 +1,41 @@
...
+    previous: Partial<ComputedData>;
+    state: State;
+    type: "donut" | "polar" | "gauge";
+    value: RendererAccessor<number>;
+    constructor(state: State, events: EventBus, el: D3Selection, options: Object<any>);
+    updateOptions(options: Object<any>): void;
+    setData(data: Datum[]): void;
+    draw(): void;
+    initialDraw(): void;
+    updateDraw(): void;
```
peterszerzo: Is there a need for all three draw methods in the public API of this class? Can't it keep track of whether it has been drawn already, and pick between private `initialDraw` and `updateDraw` based on that?

packages/visualizations/lib/PieChart/renderers/polar.d.ts
```diff
@@ -1,20 +1,44 @@
...
+    centerDisplayString(): string;
+    compute(): void;
+    angleValue(): number;
+    computeArcs(computed: Partial<ComputedData>): ComputedArcs;
+    computeOuter(drawingDims: {
+        width: number;
+        height: number;
+    }, scaleFactor?: number): (d: Datum) => number;
+    computeInner(data: ComputedDatum[], outerRadius: (d: Datum) => number): number;
     hoverOuter(radius: any): any;
```
peterszerzo: I think this needs a noun in the name, `hoverOuterSomething` to make it sound a bit more clear

packages/visualizations/lib/PieChart/renderers/gauge.d.ts
```diff
@@ -1,24 +1,51 @@
...
+declare class Gauge implements Renderer {
+    color: RendererAccessor<string>;
+    comparison: Datum;
+    computed: ComputedData;
+    currentTranslation: [number, number];
+    data: Datum[];
+    drawn: boolean;
+    el: D3Selection;
+    events: EventBus;
     extent: string;
```
peterszerzo: What does this mean?

packages/visualizations/lib/PieChart/renderers/gauge.d.ts
```diff
@@ -1,24 +1,51 @@
...
+    key: RendererAccessor<string>;
+    previous: Partial<ComputedData>;
```
peterszerzo: I think `previousComputed` can help clarify this a bit. Moving it next to `computed` also.

packages/visualizations/lib/PieChart/canvas.d.ts
ImogenF: Good point - I'll change them all to `insert`. They're not being rendered yet, and `append` suggests they're added to the end (which is usually, but not *always* the case), and I don't see much of a difference between `insert` and `attach`.

packages/visualizations/lib/PieChart/facade.d.ts
ImogenF: Pre-defined, but there are quite a few options. Could be any of the ones listed here:
https://github.com/contiamo/operational-ui/blob/refactor/viz/abstract-classes/packages/visualizations/src/utils/event_catalog.ts

Is there a nice way to use these without typing them all out one by one?

packages/visualizations/lib/PieChart/focus.d.ts
ImogenF: Element is a rendered representation of a datum - i.e. a pie chart segment, a bar in a bar chart, a node or a link in a process flow, etc. 

Component is a legend item, an axis, etc. Component foci are not used in the current frontend although they are in the frontiamo visualisations. The idea behind them is that if you hover over one of these components, it's highlighted slightly and has an `onClick` event attached to it, so that the user can click on a series name in the legend, or on an axis, and be taken to the relevant configuration menu.

packages/visualizations/lib/PieChart/legend.d.ts
ImogenF: See previous comment - `onElementHover` and `onComponentHover` do different things.

packages/visualizations/lib/PieChart/renderers/donut.d.ts
ImogenF: Sure. Haven't really got to grips with making certain methods private, but makes sense.

packages/visualizations/lib/PieChart/renderers/polar.d.ts
ImogenF: 👍

packages/visualizations/lib/PieChart/renderers/gauge.d.ts
ImogenF: `extent`? It refers to whether the gauge is shown as a semicircle or as a full circle. Will update the typings to reflect the available options.

packages/visualizations/lib/PieChart/renderers/gauge.d.ts
ImogenF: Agree with the name change. What the general opinion on alphabetising vs. grouping similar attributes?

packages/visualizations/lib/PieChart/canvas.d.ts
peterszerzo: I don't think rendering necessarily means that it's on the screen - in React, it only means creating some super fictional stuff. At the same time, it makes it clear that it creates some sort of visual representation. Inserting makes me thing of arrays.

What do you think, @TejasQ?

packages/visualizations/lib/PieChart/facade.d.ts
peterszerzo: Oh, right, I forgot you used them through a statically created object. No change required here then!

packages/visualizations/lib/PieChart/canvas.d.ts
ImogenF: In the visualisations though, `render` is used for actually displaying the different elements. So a legend, for instance, would be `insert`ed (or `attach`ed or whatever) into the DOM by the `Canvas`, and then `render`ed by the `Legend` class

packages/visualizations/lib/PieChart/renderers/donut.d.ts
ImogenF: @peterszerzo is it possible to avoid certain methods showing up in the auto-generated `.d.ts` files? Or are we happy to just have `private initialDraw(): void;`

packages/visualizations/lib/PieChart/renderers/gauge.d.ts
peterszerzo: Can we do `extent: "semi" | "full'` for now?

TejasQ:
just here to be involved and stuff

ImogenF:
@peterszerzo have address all of your comments

PR #364, opened 2018-03-19
Improve general styling

peterszerzo:
This is a more substantial PR on making operational ui more attractive. So far, the following improvements are identified:
* [x] `2px -> 4px` border radii for a more welcoming appearance, including on cards, which removes the sharp corners where the shadow looks broken.
* [x] friendlier colors for side navigation, body background and card background - all variations of the `info` color with varying degrees of desaturation.
* [x] fix and revive neglected components (e.g. `<Switch />`).
* [x] friendlier side nav + header dimensions (`52px -> 60px`).

More suggestions welcome :)

screen shot 2018-03-19 at 6 34 35 pm


packages/components/src/Switch.tsx
```diff
@@ -49,28 +51,29 @@ const Button = glamorous.div(
...
     backgroundColor: theme.colors.gray80,
+    zIndex: theme.baseZIndex,
     "&:after": {
       content: " ",
       position: "absolute",
       width: "100%",
       height: "100%",
       top: 0,
-      left: 0,
+      left: -height / 2,
```
TejasQ: What is this magic?

peterszerzo: It is a strip that slides in from the left and fills up the gray backdrop with the same color as the circular button's border.

micha-f:
Looking good so far 👍

peterszerzo:
I think this is enough changes in this PR. Opening issues for the remaining tasks.

PR #365, opened 2018-03-20
WIP: Refactoring certain inputs

TejasQ:
# In this PR
- [x] Refactor `DatePicker` to no longer use the `withLabel` internal HOC, but rather, use `Input`'s style of label handling
- [x] Refactor `Select` to no longer use the `withLabel` internal HOC, but rather, use `Input`'s style of label handling.


packages/components/src/DatePicker.tsx
```diff
@@ -91,87 +90,85 @@ class DatePicker extends React.Component<Props, State> {
...
-        }}
-        key={this.props.id}
-        isExpanded={this.state.isExpanded}
-      >
-        <Toggle
-          onClick={(ev: any) => {
-            this.setState(prevState => ({
-              isExpanded: !prevState.isExpanded
-            }))
+      <Container>
```
peterszerzo: Since `Input` does it differently (the input field is nested), I think we should consider the two options and implement this feature consistently.

Relying solely on `domId`'s to link inputs and labels a'la `<label for="">` sounds dangerous. Having two date pickers with no id and the same label or no label would screw up the focus behavior, which is guaranteed not to happen when you nest `input` inside `label`. On the other hand, always having a `domId` calculated no matter what is potentially a good outcome for accessibility. 

In the nested option, focus would work reliably out of the box, no strings attached, and the developer's responsibility would be supplying `id` (or unique label) for accessibility only.

Unless I'm missing something, the latter seems more attractive.

packages/components/src/DatePicker.tsx
```diff
@@ -91,87 +90,85 @@ class DatePicker extends React.Component<Props, State> {
...
+        {label && (
+          <Label htmlFor={id} css={css} className={className} key={id}>
+            <LabelText>{label}</LabelText>
+          </Label>
+        )}
+        <Container
```
peterszerzo: Re-using the same container glamorous component as the top-level container is an unexpected pattern, and may lead to a frustrated coworker. If the repetition/indirection/imports are daunting, consider just doing `<glamorous.Div css={{ }}>` straight up instead.

packages/components/src/DatePicker.tsx
TejasQ: I agree, although the `Input` seems to be _very_ nested inside this component... I'm wondering if it makes sense to put a label so deeply in there vs. presenting it up front.

packages/components/src/DatePicker.tsx
TejasQ: Aren't components _supposed_ and _intended_ to be reusable by design?

packages/components/src/DatePicker.tsx
peterszerzo: The label can still be all the way at the top or almost all the way at the top, like so:

```
<possiblyAContainer>
  <label>
    <bunchOfStuff>
      <input />
    </bunchOfStuff>
  </label>
</possiblyAContainer>
```

I'm sure there is room for simplification here, but keep in mind some complexity is there for all those event listeners that need to play nicely with focusing/blurring etc.

packages/components/src/DatePicker.tsx
peterszerzo: It is a bad idea to abstract components that have different responsibilities and are likely to evolve in different directions as the component is maintained. That would be extracting code just because it can be extracted. It is also an unexpected pattern: a `Container` is usually used as a top-level wrapper, and it would be unexpected to find it nested deeper down the tree. A new developer will come along and will start wondering why changing the top-level wrapper has unexpected consequences down the tree, and there will be no warning.

packages/components/src/DatePicker.tsx
TejasQ: Will change! 👍

packages/components/src/DatePicker.tsx
TejasQ: Will change! 👍

packages/components/src/DatePicker.tsx
```diff
@@ -94,80 +94,86 @@ class DatePicker extends React.Component<Props, State> {
...
-              onChange={this.props.onChange}
-            />
-          </Card>
-        </Container>
+        </Card>
       </Container>
     )
+
+    return (
+      <glamorous.Div>
```
peterszerzo: Can we get away without this container?

TejasQ:
@peterszerzo one issue: when there _is_ a label, clicking on the start date without having an end date closes the DatePicker. However, when there is _no_ label, clicking on a start date leaves the `DatePickerCard` open until the user picks an end date. I can't seem to figure out why. Would appreciate your eyes here.

peterszerzo:
@TejasQ why don't we always render the label for now, and only omit the `LabelText` if a label wasn't specified? Then we can give the component some proper attention later.

TejasQ:
@peterszerzo why don't we do it properly (read: give it proper attention) right now? We're not pressed for time or anything...

peterszerzo:
Are we not? :)

TejasQ:
@peterszerzo nope. We'll get there when we get there, but I would rather not cut corners along the way.

packages/components/src/DatePicker.tsx
```diff
@@ -91,38 +91,34 @@ class DatePicker extends React.Component<Props, State> {
   }
 
   render() {
-    const { start, end, label, id } = this.props
+    const { onChange, placeholder, start, end, label, id, css, className } = this.props
+    const { isExpanded, month, year } = this.state
     const domId = id || (label && label.toLowerCase ? label.toLowerCase().replace(/\s/g, "-") : null)
     const placeholderDays = monthStartDay(this.state.year, this.state.month)
```
peterszerzo: Use the destructured variables here, and in the line below.

packages/components/src/DatePicker/DatePicker.styles.ts
```diff
@@ -1,31 +1,45 @@
...
+
+import Card from "../Card"
 import * as mixins from "../utils/mixins"
 
 const inputHeight: number = 31
 
-export const Container: GlamorousComponent<{ isExpanded: boolean }, {}> = glamorous.div(
-  ({ isExpanded, theme }: { isExpanded: boolean; theme: Theme }): {} => ({
+export const Container: GlamorousComponent<{ isExpanded?: boolean }, {}> = glamorous.div(
+  ({ isExpanded, theme }: { isExpanded?: boolean; theme: Theme }): {} => ({
```
peterszerzo: Can we avoid the redundant typings here? Adds a lot of boilerplate.

packages/components/src/DatePicker.tsx
```diff
@@ -104,25 +104,15 @@ class DatePicker extends React.Component<Props, State> {
...
-            }))
-          }
-        >
-          <Icon name={isExpanded ? "ChevronUp" : "ChevronDown"} size={14} />
-        </Toggle>
         {!!(start && end) && (
-          <ClearButton
-            onClick={(ev: MouseEvent): void => {
+          <Toggle
+            onClick={(ev: any): void => {
```
TejasQ: Why?

packages/components/src/DatePicker.tsx
```diff
@@ -133,8 +123,9 @@ class DatePicker extends React.Component<Props, State> {
           value={[start, end].filter(s => !!s).join(" - ")}
           placeholder={placeholder || "Enter date"}
           onClick={(ev: any) => {
+            console.log(ev.target)
             this.setState(prevState => ({ isExpanded: !prevState.isExpanded }))
-            this.inputNode && this.inputNode.blur()
+            this.inputNode && this.state.isExpanded && this.inputNode.blur()
```
TejasQ: Nice!

packages/components/src/DatePicker/DatePicker.Month.tsx
```diff
@@ -53,7 +53,8 @@ const Month = ({ year, month, start, end, onChange }: Props) => {
             selected={isSelected(date, { start, end })}
             key={index}
             isPlaceholder
-            onClick={() => {
+            onClick={(ev: any) => {
```
TejasQ: Can we tighten up the `any`s in this file?

packages/components/src/DatePicker/DatePicker.styles.ts
```diff
@@ -35,7 +36,7 @@ export const DatePickerCard = glamorous(Card)(
     boxShadow: theme.shadows.popup,
     top: inputHeight + 4,
     padding: `${theme.spacing * 3 / 4}px ${theme.spacing}px ${theme.spacing * 4 / 3}px`,
-    width: 210 + 2 * theme.spacing,
+    width: 215 + 2 * theme.spacing,
```
TejasQ: We ought to variable-ize this `215`.

peterszerzo: Will do!

packages/components/src/DatePicker/DatePicker.Month.tsx
peterszerzo: I don't see value in getting these so accurate to be honest. I keep trying all kinds of `React.SyntheticEvent<>` businesses, then watch them break at the next patch upgrade of either the typings or the typescript compiler..

packages/components/src/DatePicker.tsx
peterszerzo: Sometimes `MouseEvent` works, sometimes it doesn't..

packages/components/src/DatePicker/DatePicker.Month.tsx
TejasQ: I see your point. Do you think it makes sense to invest some time looking into the `React.SyntheticEvent` type and understanding more clearly how it works?

packages/components/src/DatePicker.tsx
TejasQ: Surely that cannot be!

packages/components/src/DatePicker/DatePicker.Month.tsx
peterszerzo: I don't personally. I understand I'm letting code quality down by potentially mistyping `preventDefault`, but I also feel bad about slowing everything down spending hours studying typings.

TejasQ: I see. Personally, I'd see it as an _investment_ of time that'll help our code in the future, but okay.

PR #368, opened 2018-03-27
Improve layout+sidenav interop

peterszerzo:
`Layout` component is improved to support the case where the side navigation is expanded. Additional improvements:
* `Sidenav` is now sized as `100%` height instead of `100vh`, making it usable everywhere in any container.
* no more customizable `Sidenav` width, whether expanded or not. Having different widths in different apps looks unprofessional, and I'd like to arrive at an optimal width (currently 240px) that'll work across the board.


packages/components/src/Layout.tsx
```diff
@@ -43,11 +43,20 @@ const Container = glamorous.div(({ theme }: { theme: Theme }): {} => ({
...
 }))
 
-const Layout = (props: Props) => (
-  <Container css={props.css} className={props.className}>
-    {props.loading ? <Progress /> : <div />}
-    {props.children}
-  </Container>
-)
+const Layout = (props: Props) => {
+  const sidenavProps = (React.Children.toArray(props.children)[0] as any).props as { expanded?: boolean }
```
TejasQ: 1. Can we add a comment here that says the first child is expected to be a `SideNav` and we want its expanded prop?

1. Can we validate that we are, in fact, getting `sidenavProps` and not `myOtherComponentBecauseIDontKnowOperationalUiProps`?

1. Lastly, since we're only using the expanded prop, why not destructure to

```js
const { expanded } = (React.Children.toArray(props.children)[0] as any).props
```

?

packages/components/src/Layout.tsx
```diff
@@ -43,11 +43,20 @@ const Container = glamorous.div(({ theme }: { theme: Theme }): {} => ({
...
+  return (
+    <Container css={props.css} className={props.className} isSidenavExpanded={Boolean(sidenavProps.expanded)}>
+      {/* Absolute positioning is required in the placeholder in order to remove
```
TejasQ: `is required in the placeholder` _what?_ 

What placeholder? The word placeholder isn't used here as a variable or anything... I'm not sure what this comment means. Could we possibly make this comment a little clearer?

TejasQ:
As an afterthought: can you attach GIFs/screenshots to future visual-related PRs?

packages/components/src/Layout.tsx
peterszerzo: I'm improving the code comments, in response to both of your PR comments. In response to 3, I think naming an intermediate `sidenavProps` object helps another developer understand the implementation.

peterszerzo:
@TejasQ that's a great idea with the screenshots - I'll do that!

PR #374, opened 2018-04-03
Improve tables

peterszerzo:
* friendlier looks (merging tweaks from `pantheon-ui`)
* `EmptyView` displayed when there are no rows


packages/components/src/Table.tsx
```diff
@@ -12,10 +12,11 @@ export interface Props {
 
 const Container = glamorous.table(({ theme }: { theme: Theme }): {} => ({
   label: "table",
+  width: "100%",
   minWidth: 320,
   borderCollapse: "collapse",
-  border: "1px solid",
-  borderColor: theme.colors.separator,
+  position: "relative",
+  tableLayout: "auto",
```
TejasQ: ? This is the default value for this property so I'm not sure why it's in here.

packages/components/src/Table.tsx
```diff
@@ -42,11 +43,27 @@ const Container = glamorous.table(({ theme }: { theme: Theme }): {} => ({
...
+
 const Table = (props: Props) => (
   <Container css={props.css} className={props.className}>
     <thead>
       <tr>{props.columns.map((column, index) => <th key={index}>{column}</th>)}</tr>
     </thead>
+    {props.rows.length === 0 ? (
+      <EmptyView>
+        <tr>
+          <glamorous.Td colSpan={9999}>There are no records available</glamorous.Td>
```
TejasQ: Is this `colSpan` prop legit? It looks fishy. 🐟

peterszerzo: Agreed :) I'm having a hard time styling this one because I wanted to avoid adding a wrapper `div` outside the `table` element, but I might just do that instead.

PR #378, opened 2018-04-09
Enhance input

### LINKED ISSUE #377: Input locked state, validation and hint
Inputs need some features:
* locked for editing with an unlock feature
* hint element in a form of a question mark icon with a tooltip explaining e.g. why the input field is locked
* validation error, giving it some red borders.

```
<Input
  error="Field must be unique"
  locked={this.state.locked}
  onUnlock={() => this.setState(prevState => ({ locked: false }))
  hint="You cannot edit this field because so and so"
/>
```

Can I get some thoughts on the API @TejasQ? Can these fields be condensed or restructured? Or, since I'm in a showboaty mood: can we make them more orthogonal 🤓?

peterszerzo:
Fixes #377. This goes a bit simpler with the API using the `disabled` attribute familiar from html inputs instead of locked, and `onToggle` to signal to the parent when the disabled state toggles. If `onToggle` is specified, a lock icon is displayed as per image below.

screen shot 2018-04-09 at 10 41 24 am


packages/components/src/Input.tsx
```diff
@@ -36,14 +52,60 @@ const InputField = glamorous.input(
...
+  color: theme.colors.error,
+  padding: `${theme.spacing / 6}px ${theme.spacing * 3 / 4}px`,
+  marginTop: theme.spacing / 6,
+  marginBottom: 0,
+  width: "100%",
+  borderRadius: 3,
+  position: "absolute",
+  backgroundColor: lighten(theme.colors.error, 45),
+  boxShadow: theme.shadows.card,
+  bottom: -28,
```
TejasQ: ?

PR #379, opened 2018-04-10
Table row click handler

peterszerzo:
This seems to be the syntax that fits most cleanly into the API, or should we also pass in the proper row and the index as a second argument?

```
const rows = [ row1, row2, row3 ]

<Table rows={rows} onRowClick={(rowIndex) => {
  const clickedRow = rows[rowIndex]
  console.log(clickedRow)
}} />
```


micha-f:
I'd say that the `(row, rowIndex) =>` would be more comfortable to work with. I don't really see a case where you would not want to do something with the row (data) in the handler.

TejasQ:
I agree with @micha-f

peterszerzo:
Awesome, it's done - thanks for your feedback!

PR #382, opened 2018-04-10
Select component fixes

peterszerzo:
Fixes a few issues on select that cause runtime errors, including search on missing labels (since they are optional), clearing values. Also a few visual improvements.


packages/components/src/Select.tsx
```diff
@@ -166,13 +166,13 @@ class Select extends React.Component<Props, State> {
               <OptionsList>
                 {options.map(
                   (option: IOption) =>
-                    option.label.match(new RegExp(search)) && (
+                    (option.label || String(option.value)).match(new RegExp(search)) && (
```
TejasQ: Are you sure the `new` keyword is essential here?

packages/components/src/Select/Select.style.ts
```diff
@@ -67,9 +68,11 @@ const DisplayValue: GlamorousComponent<IDisplayValueProps, {}> = glamorous.div(
 const Options: GlamorousComponent<{}, {}> = glamorous.div(
   {
     position: "absolute",
-    top: "calc(100% + 1px)",
+    top: "calc(100% + 6px)",
```
TejasQ: Why the magic 6?

peterszerzo: So it doesn't go overlap with the focus shadow. Using `theme.spacing` won't make it clearer (since the focus shadow is hard-coded), so I'll add a comment instead.

TejasQ: Wonderful! Thanks!

packages/components/src/Select.tsx
peterszerzo: No, I am not sure, and with a very good reason.

PR #399, opened 2018-04-16
Fix removal of exiting sunburst elements.

ImogenF:


packages/visualizations/src/utils/d3_utils.ts
```diff
@@ -19,6 +19,9 @@ export const onTransitionEnd = (selection: D3Transition, func?: () => void): D3T
   if (!func) {
     return
   }
+  if (selection.empty()) {
+    func()
```
peterszerzo: Seems like you can `return` after this execution because the rest of the code won't run anyway?

PR #412, opened 2018-04-25
Page enhancements

peterszerzo:
## Page

Repositioning the elements inside the `controls` prop and adding an opinionated back link feature as in the image below:

screen shot 2018-04-25 at 11 16 10 am

API like so:
```
<Page
  title="Title"
  breadcrumbs={/* as before */}
  controls={/* as before */}
  __experimentalBackLinkUrl="/abcd"
  __experimentalBackLinkLabel="Back to resources"
/>
```

Label defaults to `Back` if not specified, nothing is rendered if url isn't specified. Supports pushstate only.

## Breadcrumbs

Fixes an issue of all breadcrumbs elements being styled as links, and instead styling only the ones that are actual `a` tags (which are now expected to reside inside the breadcrumbs like so: `<Breadcrumb><a>Link</a></Breadcrumb>` as opposed to nested the other way around. This activates the breadcrumbs a bit better, showing where the closest relevant back link is. The last item in the breadcrumbs, denoting the current page, is now black (see image).


peterszerzo:
@micha-f it seemed a bit more up in the air to me, especially since I find the current visual outcome cluttered and confusing. For instance, the back link could go into the `controls` prop next to the edit link as per one of the sketches below.

screen shot 2018-04-24 at 5 12 04 pm

I understand you feel very strongly about having the back link and I would be happy to leave it at that and make it work, but at the same time I do not recall that we reached a definitive conclusion on the matter. Some writings on breadcrumbs like https://www.smashingmagazine.com/2009/03/breadcrumbs-in-web-design-examples-and-best-practices/#benefits-of-using-breadcrumbs and https://www.hongkiat.com/blog/breadcrumb-navigation-examined-best-practices-examples/ consider it a perfectly valid alternative to back links, which is why I suggested that we tested it with real users once the UI is ready.

PR #413, opened 2018-04-25
More flexible page back button api

peterszerzo:
Changing back button api to the following:
```
<Page
  {...bunchOfOtherProps}
  __experimentalBackButton={{
    url: "/abc",
    label: "ABC",
    onClick: () => {
      reactRouterHistory.pushState("/abc")
    }
  }}
/>
```
The previous implementation used the exposed window history, which react ignores :/.


packages/components/src/Page.tsx
```diff
@@ -62,19 +65,19 @@ const BackLinkContainer = glamorous.div(({ theme }: { theme: Theme }): {} => ({
 const Page = (props: Props) => (
   <Container>
     <TopBar>{props.breadcrumbs}</TopBar>
-    {props.__experimentalBackLinkUrl ? (
+    {props.__experimentalBackLink ? (
```
TejasQ: I have serious issues with `condition ? <Stuff /> : null`. Why not just use the logical and operator?

peterszerzo: I find it unsafe and tricky to read. What if the condition is `false` instead of `undefined`, or `""`? It probably works fine, but I don't want to keep in my head what React does in whichever case, nor do I want to keep my fingers crossed that a later version doesn't smack a stringified `"false"` in the middle of my UI. Also, explicit `null` is consistent with the expectation of a component that renders nothing (which is a case where `undefined` is not even allowed).

TejasQ: Peter, this has nothing to do with React and is basic javascript control flow. If, in case of an `else`, do `null`, why even handle the case? I find this really confusing. Let's discuss on Monday. Besides this, everything looks good.

PR #419, opened 2018-04-30
Upgrade build dependencies (merge 3rd)

peterszerzo:


peterszerzo:
@ImogenF I recommend rebasing on top of this as soon as it is merged, to avoid nasty rebase conflicts down the line. Also, because of upgrades in `tslint`, code like this will fail in the CI:
```
if (condition) {
  return 1
  // Need to remove the else block and have the `return 2` be out on its own
} else {
  return 2
}
```

TejasQ:
This PR is far too large to review...

peterszerzo:
All that's been changed by hand is dependency versions, and 3 lint errors in visualizations (of the type shown above). The rest is prettifier differences (trailing commas in multiline arrays mostly) and compiled files.

PR #429, opened 2018-05-07
Sidenav gradient

peterszerzo:
screen shot 2018-05-07 at 11 33 46 am

How does it look?


micha-f:
Sexy :)

TejasQ:
I severely dislike it because, to me, it looks like Windows 98. Can we make it more subtle?

peterszerzo:
I agree, @TejasQ, I'll take care of it!

peterszerzo:
This is the new version, I think it looks sufficiently subtle enough to remove Windows associations :)

screen shot 2018-05-09 at 8 14 14 am

TejasQ:
Can we have it point bottom-right to top-left in a 45º angle?

peterszerzo:
Agreed, looks better :)

image

TejasQ:
Incredible.

PR #431, opened 2018-05-08
Add components for Identification

TejasQ:
# In this PR
Two components: `Avatar` and `AvatarGroup`.

## Avatar
A component that is a simple `Avatar` that falls back to a `name` prop's initials if no `photo` is specified. Further, it has a `withName` prop that adds a name to it nicely.

### Default Settings
image

### withName
image

## AvatarGroup
A component that is a collection of `Avatar`s that takes in a group of people `{ name: string, photo?: string }` and renders a string of Avatars in a pretty way.

image


packages/components/lib/index.d.ts
```diff
@@ -1,4 +1,6 @@
 export { Theme, operational as operationalTheme } from "@operational/theme";
+export { default as Avatar } from "./Avatar";
+export { default as AvatarCollection } from "./AvatarCollection";
```
peterszerzo: Name this `AvatarGroup`, consistent with `ButtonGroup`

packages/components/src/Avatar.tsx
```diff
@@ -0,0 +1,66 @@
+import * as React from "react"
+import glamorous, { CSSProperties, Div } from "glamorous"
+import { Theme } from "@operational/theme"
+
+export type WithTheme = { theme: Theme }
+
+export interface AvatarProps {
+  name: string
+  withName?: boolean
```
peterszerzo: I think the name here needs a verb, like `displayName` or `showName`, especially since we have `withRouter` and `withTheme` methods doing very different things in this codebase.

packages/components/src/Avatar.tsx
```diff
@@ -0,0 +1,66 @@
+import * as React from "react"
+import glamorous, { CSSProperties, Div } from "glamorous"
+import { Theme } from "@operational/theme"
+
+export type WithTheme = { theme: Theme }
+
+export interface AvatarProps {
```
peterszerzo: Avatar background color logic should work the same way as in the `Button` component. Users should be able to tailor it, and get readable text colors for the initials automatically.

Alternatively, it could be interesting to do some random-seeming but deterministic color assignments like you had earlier, @TejasQ. Why don't we just add the ascii codes of all the letters in the name prop, take the remainder when divided by say 8, and pick one from the visualization palette? This feature could be on by default, and overridden when `color` prop is specified.

packages/components/src/Avatar.tsx
```diff
@@ -0,0 +1,66 @@
...
+  name: string
+  withName?: boolean
+  photo?: string
+  css?: CSSProperties | (<T>(props: T & WithTheme) => CSSProperties)
+  size?: number
+}
+
+const Avatar = ({ name, photo, css, withName = false, size = 50 }: AvatarProps): React.ReactElement<AvatarProps> => {
```
peterszerzo: Either decrease the default size or bump the font size of the initials for better proportions). I'm not sure if it makes sense to allow users to customize it, maybe have it hard-coded (or in`./constants.ts`) and see how our needs evolve?

packages/components/src/AvatarCollection.tsx
```diff
@@ -0,0 +1,33 @@
+import * as React from "react"
+import glamorous, { CSSProperties, Div } from "glamorous"
+import { Theme } from "@operational/theme"
+
+import { Avatar } from "./"
+
+export type WithTheme = { theme: Theme }
+
+export interface AvatarCollectionProps {
```
peterszerzo: Is there a reason we prefix the name of this interface with the name of the component? No other component does it. Of course, if you think that explicitly exporting all the interfaces from this library makes sense, then we should open a separate discussion for it.

packages/components/src/AvatarCollection.tsx
```diff
@@ -0,0 +1,33 @@
...
+  people: {
+    name: string
+    photo?: string
+  }[]
+  size?: number
+  css?: (props: WithTheme) => CSSProperties | CSSProperties
+}
+const AvatarCollection = ({ people = [], size = 32, css = {} }: AvatarCollectionProps) => (
```
peterszerzo: Same comment with the size - I think at first they can all be the same and not modifiable, and we can have a look later.

packages/components/src/Avatar.tsx
```diff
@@ -0,0 +1,66 @@
...
+import glamorous, { CSSProperties, Div } from "glamorous"
+import { Theme } from "@operational/theme"
+
+export type WithTheme = { theme: Theme }
+
+export interface AvatarProps {
+  name: string
+  withName?: boolean
+  photo?: string
+  css?: CSSProperties | (<T>(props: T & WithTheme) => CSSProperties)
```
peterszerzo: Does it make sense to support a `title` attribute? Then we can put something in the second line below the name in a more grayed out color, and have the component balance out a bit better (big circle next to small single-line text looks a bit strange)

packages/components/src/Avatar.tsx
TejasQ: I think that's a little over-engineered: ASCII/8 and so on.

I think what would be most idiomatic is a completely random color that is overridable by using a `color` prop that maps to theme colors (like `Button`) or takes a custom hex and handles contrast accordingly. 

What do you think?

packages/components/src/Avatar.tsx
TejasQ: This was extracted from the Labs UI which needs configurable sizes currently. Albeit there are just _two_ different sizes: one for the list view, another for detail; I think it's something that would be nice to have configurable, especially considering GitHub's own use of avatars: your avatar next to your comment is significantly smaller than the one [here](https://github.com/peterszerzo), for example.

packages/components/src/AvatarCollection.tsx
TejasQ: It was a copypasta from Labs. Will change.

packages/components/src/AvatarCollection.tsx
TejasQ: [Same response](https://github.com/contiamo/operational-ui/pull/431#discussion_r187006614)

packages/components/src/Avatar.tsx
TejasQ: While I agree that it makes sense to support it, currently, there is no need for this and I'm not sure it should block the deployment of this feature. We can always add it if one of our UIs need it.

packages/components/src/Avatar.tsx
peterszerzo: I agree that there is complexity, but I don't see another solution that works visually. If you want random colors, you'd have to make this into a class component so the randomly generated value is persisted on the instance and not re-generated on every render. Even if we did that, the same person would have a different color after a refresh, which would make things look untrustworthy (in my gmail app, I rely on my brother's avatar being colored yellow when I scanning a list for his e-mails, and I'm super thrown off when it changes).

At the same time, `O(n)` is not a scalability concern when names will always have n < 20 characters, and the operation would take way below 1ms anyways, even if you have dozens of avatars on the same page. You can always memoize the result with a one-liner if my performance estimate proves wrong.

packages/components/src/Avatar.tsx
peterszerzo: I wouldn't re-use the avatar component for a comment - that would couple font sizes and spacings in components with very different usages and contexts, making the UI difficult to adapt later.

If we have two different sizes, relying on a familiar prop like `condensed` to toggle between large and small seems like a better way to go. If we do end up needing more than two avatar sizes, we can certainly think about supporting it then.

packages/components/src/Avatar.tsx
peterszerzo: I would argue that they need it visually, but I am happy to leave it as an enhancement.

packages/components/src/Avatar.tsx
TejasQ: I should introduce you to my friend [Nagarajan Ganapathysubramanian](https://www.linkedin.com/in/nagarajan-ganapathysubramanian-a4920213/). I think he'd appreciate the complexity of this solution.

I get the point about a different color on every re-render not being ideal, _but_ I _must_ stress the complexity being a problem on a view like the `IdP` with tens or hundreds of accounts per page. Could you imagine doing the calculation on each render?

How about we leave the problem up to the user and say we support a `color` prop, but you handle your own random-but-not-so-random assignment logic and complexity that comes with it?

PR #435, opened 2018-05-08
Parcel builds for website and visual tests

peterszerzo:
This should take care of `@types/react` inconsistencies with `create-react-app-ts`.


TejasQ:
Can we please get the gradient out of this PR?

peterszerzo:
If we merge sidenav gradient (it is ready with comments addressed), then we can go ahead with this one without a rebase. Does that sound good?

PR #439, opened 2018-05-14
Enhance documentation

### LINKED ISSUE #424: <Page> doesn't deal with the height of the header
When I have a `<Page />` into the `<OperationalUI />`, the height of the page in not correct. In fact the `<Page />` have `heigth: 100%` without take care to the header height.

### Code

```jsx
<OperationalUI withBaseStyles>
    <Router>
      <Layout
        css={{ height: "100vh" }}
        sidenav={
          <Sidenav expanded>
                <SidenavHeader label="hello" icon="Star" />
          </Sidenav>
        }
        main={
          <div>
            <Header>Identity Provider</Header>
            <Page title="hi">
              my content
            </Page>
          </div>
        }
      />
    </Router>
  </OperationalUI>
```

### Result

image

(issue on bottom, when we scroll)

### LINKED ISSUE #433: Add anchor navigation to website
It would be quite nice to have https://operational-ui.js.org/#checkboxes take you straight to _that_ component's demo. 

Thoughts?

peterszerzo:
* [x] search bar
* [x] link to snippets
* [x] stateful components demo state changes
* [x] docs for page and layout (fix #424)
* [x] anchor navigation (fix #433)


packages/website/src/Sections/Components/Breadcrumbs.tsx
```diff
@@ -1,14 +1,15 @@
 import * as React from "react"
 import { Breadcrumbs, Breadcrumb } from "@operational/components"
+import * as constants from "../../constants"
 
 export const title = "Breadcrumbs"
 
-export const docsUrl = "https://github.com/contiamo/operational-ui/blob/master/docs/components/breadcrumb.md"
+export const docsUrl = `${constants.docsBaseUrl}/components/breadcrumb.md`
+
```
TejasQ: Do we need all this whitespace?

peterszerzo: I tend to add a whitespace between exports so they don't get lost. It is kind of treating simple exported constants like functions, where we usually add whitespaces - but I find it hard to form a ruleset that works in every situation. How would you approach this?

TejasQ: Makes sense.

PR #441, opened 2018-05-14
Visualisation styling

ImogenF:


packages/visualizations/src/utils/focus_utils.ts
```diff
@@ -68,34 +116,70 @@ const FocusUtils: Object<any> = {
...
+          newPosition = "toLeft"
+          arrowX = label.width
+        } else {
+          arrowX = 0
+        }
         break
       default:
         throw new Error(`Invalid label position '${position}'.`)
     }
     // Finally. Done.
```
peterszerzo: I'd clarify this comment a bit `Do operations if calculations were successful`, or remove it altogether

PR #442, opened 2018-05-15
[this will never be merged] Basic charts review

peterszerzo:


packages/visualizations/src/Chart/facade.ts
```diff
@@ -27,20 +27,29 @@ import {
 } from "./typings"
 
 const defaultConfig: ChartConfig = {
+  // R transitionDuration?
   duration: 1e3,
   height: 500,
+  // R what is hidden?
```
ImogenF: The chart. All visualisations have this option, where they can be hidden and shown on demand.

packages/visualizations/src/Chart/facade.ts
```diff
@@ -27,20 +27,29 @@ import {
...
   hidden: false,
+  // R innerBarSpacing?
   innerBarPadding: 2,
+  // R do we need this to be separate? Wouldn't the user already be aware of whether
```
ImogenF: The d3 categorical axis scale uses a fraction rather than a pixel value for spacing - rather annoying, but not really avoidable.

packages/visualizations/src/Chart/facade.ts
```diff
@@ -27,20 +27,29 @@ import {
...
+  //   they set up a categorical axis?
   innerBarPaddingCategorical: 0.2,
   legend: true,
   maxBarWidthRatio: 1 / 3,
   minBarWidth: 3,
   numberFormatter: (x: number): string => x.toString().replace(/\B(?=(\d{3})+(?!\d))/g, ","),
+  // R barSpacing? I feel like people will understand that this is the outer one
```
ImogenF: I'm not sure I agree that this would be clear. I would assume `barSpacing` to be an overriding value that sets spacing between all bars, whether on the same tick or not.

packages/visualizations/src/Chart/facade.ts
```diff
@@ -27,20 +27,29 @@ import {
...
   outerBarPadding: 10,
   palette: theme.colors.visualizationPalette,
   timeAxisPriority: ["x1", "x2", "y1", "y2"],
+  // R what do we need the uid for? Would a unique one be generated if we render multiple charts
+  //   on the same page?
```
ImogenF: Yup.

packages/visualizations/src/Chart/facade.ts
```diff
@@ -27,20 +27,29 @@ import {
...
   uid: uniqueId("chart"),
   visualizationName: "chart",
+  // R is there a way to set this to take up the width of the container?
```
ImogenF: Not sure I understand - width of which container? The chart container? The visualisation needs to be explicitly told what size to render...

packages/visualizations/src/Chart/facade.ts
```diff
@@ -60,7 +69,7 @@ const defaultSeriesAccessors: SeriesAccessors = {
   hide: (d: Object<any>): boolean => d.hide || false,
   hideInLegend: (d: Object<any>): boolean => d.hideInLegend || false,
   key: (d: Object<any>): string => d.key || uniqueId("key"),
-  legendColor: (d: Object<any>): string => initialColorAssigner(d.key),
+  legendColor: (d: Object<any>): string => initialColorAssigner(d.key), // R this can just be colorAssigner(defaultConfig.palette)(d.key) right? I find myself wondering why the two const declarations above are there
```
ImogenF: The solution you're suggesting would result in a *new* colorAssigner every time series.legendColor() is called - so would always return the first color from the palette.

packages/visualizations/src/Chart/series_manager.ts
```diff
@@ -85,6 +85,8 @@ class ChartSeriesManager implements SeriesManager {
     const barIndices: Object<number> = {}
     forEach((series: Object<any>): void => {
       const hasBars: boolean = !!find((renderOptions: RendererOptions<any>) => renderOptions.type === "bars")(
+        // R is there actual drawing happening here? The name `computeBarIndices` would be
+        //   confusing in that case.
```
ImogenF: No, no drawing happening here.

packages/visualizations/src/Chart/typings.ts
```diff
@@ -264,6 +264,7 @@ export interface DataForLegends {
 }
 
 // @TODO
+// R what is this TODO for?
```
ImogenF: Once focus labels are added, I'll add a proper type for the payload that is sent with the hover event.

packages/visualizations/src/Chart/typings.ts
```diff
@@ -274,6 +275,7 @@ export interface HoverPayload {}
 export type Focus = Focus<HoverPayload>
 
 // @TODO
+// R what is this TODO for?
```
ImogenF: Reminder to uncomment the `focus` part once focus labels are implemented.

packages/visualizations/src/Chart/facade.ts
micha-f: Not sure this is true actually. We might want to return the default config from a function to ensure a different uid property is computed every time. It's also kind of an internal thing - not something a user should be able to overwrite in a provided config value.

packages/visualizations/src/Chart/facade.ts
peterszerzo: Awesome, it all makes sense!

packages/visualizations/src/Chart/facade.ts
```diff
@@ -27,20 +27,29 @@ import {
 } from "./typings"
 
 const defaultConfig: ChartConfig = {
+  // R transitionDuration?
```
ImogenF: This is used in so many places, in all of the visualisations - I'd like to avoid making it a longer name, if possible.

ImogenF:
@peterszerzo Happy for me to delete this PR and the branch?

PR #444, opened 2018-05-15
Add charts to website.

ImogenF:


packages/website/src/Sections/Visualizations/Charts.tsx
```diff
@@ -0,0 +1,85 @@
+import * as React from "react"
+import { VisualizationWrapper, Chart } from "@operational/visualizations"
+import { Div } from "glamorous"
+
+export const title = "Charts"
+
+export const docsUrl = "https://github.com/contiamo/operational-ui/blob/master/docs/visualizations/chart.md"
+
```
peterszerzo: Add `snippetUrl` (see components).

ImogenF: Fixed.

PR #449, opened 2018-05-18
Make sunburst breadcrumbs configurable.

ImogenF:
Added 2 sunburst config options:

* breadcrumbItemWidth - width of each breadcrumb item (excluding arrow)
* maxBreadcrumbLength - maximum number of breadcrumb items to display without truncating the list


packages/visualizations/src/Sunburst/breadcrumb.ts
```diff
@@ -76,11 +86,13 @@ class Breadcrumb {
     trail.exit().remove()
 
     // Add breadcrumb and label for entering nodes.
+    const itemWidth = (d: any): number => (d === "hops" ? 40 : this.state.current.get("config").breadcrumbItemWidth)
```
TejasQ: What does the `40` do?

ImogenF: Have updated the code to make this clearer.

PR #452, opened 2018-05-22
Deprecation

peterszerzo:
This PR introduces a HOC that makes it easy to log deprecation warnings in a component based on the props it is given. This should allow gentler code changes, gradual upgrades, and help handling semver more easily. Logs are only seen in development, and only once per component, avoiding log pollution.

Usage:
```
const SomeComponent = (props: Props) => (
)

export default deprecate(props =>
  props.oldProp
    ?
    [ "`oldProp` is deprecated, use `newProp` instead" ]
    :
    []
)(SomeComponent)
```


TejasQ:
I think this is something that should've been a discussion before a PR appeared. 

Why do we need a HOC that logs deprecation warnings in a component that is slated for deprecation? Why don't we simply add the warnings on a _per-component_ basis since we'll remove them anyway in the next major version?

peterszerzo:
@TejasQ this PR is a basis for the discussion, referenced in the notion page. I wanted to see if it's doable.

Also, it's not mainly for a components that are staged for deprecation, but for individual props that are renamed or migrated. I'll explain today.

fabien0102:
Good idea, to improve a little bit this pattern, I think that we can build something like this.

```typescript
const MyComp = ({a: string}) => <div>{a}</div>

export default deprecated([
{oldProp: "b", newProp: "a" }
])(MyComp)

// Usage
<MyComp a="hello" /> // new way

<MyComp b="hello" /> // old way (with a deprecation warning)
```

And for most avanced case

```typescript
const Greet = ({name: string}) => <div>Hi {name}</div>

export default deprecated([
{oldProp: "firstname" },
{oldProp: "lastname"}
], (props) => ({
name: props.firstname + " " + props.lastname
}))(Greet);

// Usage
<Greet name="Fabien Bernard" /> // new way

<Greet firstname="Fabien" lastname="Bernard" /> // old way (with a deprecation warning)

```

fabien0102:
https://www.npmjs.com/package/react-deprecate

peterszerzo:
I think it's clearer to keep the renaming/adapting as part of the individual component's implementation, and keep responsibilities of the deprecator method super thin. It should only read the props and maybe do some logging, maybe not, but you wouldn't need to figure out how it works when you are trying to find a bug.

What I imagined was that the component would still use:
```
export interface Props {
  oldProp?: string
  newProp?: string
}
```

But rename it in its render method - making the interface clear at a first glance in the implementation.

Whether we add an abstraction that helps with auto-generating warnings for renamed props is a decision that I think is best left for the future. I think we'll have more complicated use-cases than just renaming (like checking if there is an `<a/>` tag as a direct child), where we might want to link to the docs or do crazier things. We'll know better once we know what the component is used most often for.

TejasQ:
Have we fully considered using a pre-made solution as @fabien0102 suggested as opposed to rolling our own? I can already see some immediate benefits here, the chief one being _it already exists_.

peterszerzo:
I have. I think that using the rename feature makes component implementations confusing to read, especially once we start having a lot of deprecation code around. The roll-our-own solution leaves component behavior completely unaffected (could use `react-deprecate` the same way, but then why use a library..), weighs 15 lines, and is already done.

packages/components/src/utils/deprecate.tsx
```diff
@@ -0,0 +1,19 @@
+import * as React from "react"
+
+type P = any
+
+const deprecate = (createWarning: (props: P) => string[]) => (
```
fabien0102: You can use a generic here instead of this ugly `any` :-p

peterszerzo: Care to give me a hand on it sometime today?

fabien0102: `<T>(createWarning: (props:T))` should works

fabien0102: After we need to try if everything follow as expected, we can works on it after ;)

packages/components/src/utils/deprecate.tsx
```diff
@@ -0,0 +1,32 @@
...
+ * export default deprecate<Props>(props => [ 
+ *   ...[ "Your component is deprecated" ],
+ *   ...(props.oldProp ? [ "oldProp is deprecated, use newProp instead" ] : [])
+ * ])(Comp)
+ *
+ */
+function deprecate<P>(createWarning: (props: P) => string[]) {
+  return (Component: React.ComponentType<P>): React.ComponentType<P> => {
+    let logCount = 0
+    return (props: P) => {
```
TejasQ: We'll use a stateful component here as discussed with @peterszerzo and @fabien0102.

peterszerzo:
The blocks package is also being removed because it is causing some typings errors due to outdated dependencies. Since it's not used anywhere in production, I'd like to take them out instead of spending time fixing them.

packages/components/src/utils/deprecate.tsx
```diff
@@ -14,17 +14,21 @@ import * as React from "react"
...
-    let logCount = 0
-    return (props: P) => {
-      const deprecation = createWarning(props)
-      if (process.env.NODE_ENV === "development" && deprecation.length > 0 && logCount === 0) {
-        console.log(`Operational UI warnings:\n\n  * ${deprecation.join("\n  * ")}`)
-        logCount = logCount + 1
+    return class extends React.Component<P> {
+      componentDidMount() {
+        const warnings = createWarning(this.props)
+        if (process.env.NODE_ENV === "development" && warnings.length > 0) {
```
TejasQ: Can we instead say `!== "production"` so that these show up in staging and all other envs?

peterszerzo: Good point!

PR #454, opened 2018-05-23
Select fixes

### LINKED ISSUE #438: Select component issues
* component doesn't close when the input is clicked the second time after opening (currently only works on outside click)
* search is case sensitive - should be case insensitive and attempt to match both label and value, if given
* search needs padding that makes it consistent with the rest

peterszerzo:
Fixes #438, with some visual improvements:

screen shot 2018-05-23 at 1 38 49 pm


packages/components/src/Select/Select.Option.tsx
```diff
@@ -12,45 +12,43 @@ export interface Props {
   selected?: boolean
   onClick?: () => void
   children?: React.ReactNode
-  color?: string
 }
 
-const Container = glamorous.div(
-  ({ theme, color, selected }: { theme: Theme; color?: string; selected: boolean }): any => {
-    const backgroundColor = expandColor(theme, color) || theme.colors.white
+const Container = glamorous.div(({ theme, selected }: { theme: Theme; selected: boolean }): any => {
```
TejasQ: Can we remove the `any` and allow TS to infer the type?

PR #455, opened 2018-05-24
Button spinner

### LINKED ISSUE #315: Button spinner
Most components have async interactions: buttons do something async on click (like requests), autocomplete/select components could do fetches on filter, going forward, _table_ blocks could use filtering, the login block, etc.

It might be a good idea to expose a component that creates and gives out a spinner to use inside of any old component in the event one is needed, to signal background activity.

Here's a [sample implementation](https://codesandbox.io/s/moojrr58mx) that I threw together of what this could look like. Thoughts?

## Use Case
There is a use case in the Labs UI for this or something like this, for example, when one clicks "Deploy" on a bundle: it fires off a request. It wouldbe nice to have the button disabled with a status indicator on the button.

image

peterszerzo:
Fixes #315, adding an optional `loading` boolean prop to the button component.

screen shot 2018-05-24 at 10 50 07 am


TejasQ:
It would be nice if the button's label stayed visible with a spinner added _beside_ it, perhaps even with a stretch animation so we don't lose context of what the button _is_.

peterszerzo:
Looks like we have two conflicting goals here:
* as you mentioned, it is better if the label is visible
* the button shouldn't have to change size when going into a loading state, otherwise it causes visual distraction and even more undesirable consequences such as sending some adjacent content to a new line. The ways in which this is possible with unchanged labels would be a small icon at the edges like in the legacy Contiamo UI. However, I find that very unappealing aesthetically (either forces huge paddings or an oddly small spinner).

I would vouch for keeping the current solution and mitigating the damage caused by the hidden label by e.g. a tooltip that shows up after a while, or just a plain title attribute. I don't think the loss of context is such an issue since you just clicked the button, and the extra measures should cover the case when a loading takes so long that a user already forgot what it was for.

packages/components/src/Button/Button.tsx
```diff
@@ -35,51 +30,65 @@ export interface Props {
...
+      backgroundColor: active ? activeBackgroundColor : backgroundColor,
+      color: loading ? "transparent" : foregroundColor,
+      opacity: disabled ? 0.6 : 1.0,
+      outline: "none",
+      position: "relative",
 
-    color: textColor,
-    opacity: disabled ? 0.6 : 1.0,
-    outline: "none",
+      "& .op_button-spinner": {
```
TejasQ: We use classnames now?

peterszerzo: Only in very rare cases when applying styles on children can't be done in an unambiguous way.

packages/components/src/Button/Button.tsx
```diff
@@ -35,51 +30,65 @@ export interface Props {
...
+        color: foregroundColor,
+      },
 
-    ...!disabled
-      ? {
-          ":hover": {
-            backgroundColor: activeBackgroundColor,
-            color: readableTextColor(activeBackgroundColor, ["white", "#222"]),
-          },
+      ...!disabled
```
fabien0102: Can switch the true/false case here, to avoid this strange `...!` pattern ^^ (so begin by the disabled case)

peterszerzo: This is intentional - most of the times buttons are enabled, and it is clearer to read the common case first. We're using the `disabled` instead of `enabled` variable because it is an html property and will avoid having to do conversions (although it might make the reading clearer here).

PR #468, opened 2018-05-30
`to` prop

### LINKED ISSUE #445: Render `as` prop
Components should be able to customize the html tag they render by, like so:

```
<Button
  as={Link}
  to="/somewhere"
  css={{ ":hover": {} }}
>
  Go somewhere
</Button>
```

### Challenges

For some components, it is tricky to reliably pass props along to the rendered element, and to make that compatible with internal rendering logic. For instance, in the example above, hover styles cannot be passed along because react router's link is not a glamorous component. One solution to this is to specify that we recommend that only glamorous components be passed into `as`, and if that's not the case, then dear user, don't be surprised. We can supply an alternative `Link` implementation that renders a `<glamorous.a />` (tap into context and redirect, set href, prevent default, all that fun stuff).

I would argue that it doesn't make sense to do this everywhere. Buttons, sidenav stuff, and typography elements have a very strong case for this feature to avoid unnecessary html tags and to enhance accessibility, but not so much any other component (correct me if I'm wrong :)). Restricting the scope this way can also reduce the amount of custom logic we'll have to maintain in the components.

peterszerzo:
Establishes a pattern to fix #445.

Adds pushstate routing to operational components via context - also known as `create an app with routing without react-router in about 25 lines`. Usage example here: https://github.com/Contiamo/operational-ui/blob/b1431858c8bc0249761df96a291cea2f6b9333d5/packages/components/scripts/dev-server/index.tsx


packages/components/src/Button/Button.tsx
```diff
@@ -3,19 +3,25 @@ import glamorous, { GlamorousComponent } from "glamorous"
...
   css?: Css
   className?: string
-  /** Invoked when you click on the button */
-  onClick?: React.HTMLProps<HTMLButtonElement>["onClick"]
+  /* 
+   * Invoked when you click on the button 
+   *
+   * @todo Adjust typings to be compatible with React.SyntheticEvent
+   */
+  onClick?: (ev: any) => void
```
peterszerzo: @fabien0102 the typing you added is incompatible with `React.SyntheticEvent` declaration in the `onClick` call below, and taking the type declaration out there doesn't work either. What is the right way to type this?

I would prefer not passing the event along at all until we have evidence that working with it is necessary in production (as in functionality can't be achieved through the other props, the new `to` prop included). This is what the other `onClick: () => void` declarations throughout the codebase do in order to keep the API simpler. Let me know if anything comes up.

packages/components/src/SidenavHeader/SidenavHeader.tsx
```diff
@@ -37,44 +43,54 @@ export interface State {
...
+  return {
+    label: "sidenavheader",
+    textDecoration: "none",
+    width: "100%",
+    position: "relative",
+    borderBottom: "1px solid",
+    borderLeft: "4px solid",
+    borderLeftColor: isActive ? stripColor : "transparent",
+    borderBottomColor: theme.colors.separator,
+    backgroundColor: isActive ? "#F8F8F8" : "transparent",
```
fabien0102: This color should be inside the theme

packages/components/src/SidenavItem/SidenavItem.tsx
```diff
@@ -4,70 +4,92 @@ import { Theme } from "@operational/theme"
...
+import { Icon, IconName, ContextConsumer, Context } from "../"
+import { isModifiedEvent } from "../utils"
 
 export interface Props {
   id?: string
   /** `css` prop as expected in a glamorous component */
   css?: Css
   className?: string
   onClick?: () => void
+  /** Navigation property a'la react-router <Link/> */
```
fabien0102: `a'la` ??

PR #471, opened 2018-05-31
General chart improvements

ImogenF:
Final chart features and fixes. 
Type improvements and unit tests will be in a separate PR.


ImogenF:
https://visual-tests-chart-improvements.fe.ctmo.io

micha-f:
image

@ImogenF There is something weird about the "Flag 3" / "Flag 4" labels here

ImogenF:
@micha-f good spot, I'll check what went wrong there...

micha-f:
image

@ImogenF This one is really difficult to understand. Is it correct?

ImogenF:
@micha-f The bar ranges one is correct, yes. The lines are there just for reference but can easily be removed. I'm not sure what the use case would be for rendering a range as bars, particularly with both positive and negative values, but I implemented it for consistency, so that the user can apply any of the the standard renderers to a range.

micha-f:
image

@ImogenF There seems to be something overlaying the bottom left of the chart (close to 0,0)

ImogenF:
@micha-f Have fixed the axis being cut off. Can't recreate the problem with the flags - yesterday I could only see the problem in the deployed version of the visual tests, not locally, and today it's not even appearing in the deployed tests. Could you check again please? Apart from that, am I ok to merge?

ImogenF:
@TejasQ Thanks for reviewing. Did you get a chance to look into the CI errors? Can I ignore them?

micha-f:
@ImogenF  Cannot reproduce either. Weird.

TejasQ:
For clarity, the CI fails because our test runner hangs. We're currently looking into it!

PR #472, opened 2018-05-31
Tune sidenav

peterszerzo:
Fine-tuning sidenav to treat headers more like organizing elements (they show their children even if they themselves are not links/activated). Related style tweaks.

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     "build": "cd packages/website && npm run build",
     "bootstrap": "lerna bootstrap",
     "postinstall": "npm run bootstrap",
-    "clean": "rm -rf node_modules && rm -rf packages/*/node_modules",
+    "clean": "rm -rf node_modules && rm -rf packages/*/node_modules && rm -f packages/*/package-lock.json && rm -f packages/*/yarn.lock && rm -f yarn.lock",
     "test": "lerna run test",
     "lint": "lerna run lint",
     "prettier": "lerna run prettier",
--- a/packages/components/src/SidenavHeader/SidenavHeader.tsx
+++ b/packages/components/src/SidenavHeader/SidenavHeader.tsx
@@ -64,6 +64,9 @@ const containerStyles = ({
     borderBottomColor: theme.colors.separator,
     /** @todo Add to theme once colors are updated across codebase */
     backgroundColor: isActive ? "#F8F8F8" : "transparent",
+    ":hover": {
+      backgroundColor: "#F8F8F8",
+    },
   }
 }
 
@@ -73,6 +76,7 @@ const ContainerLink = glamorous.a(containerStyles)
 
 const Content = glamorous.div(({ theme, isActive }: { theme: Theme; isActive: boolean }): CssStatic => ({
   textDecoration: "none",
+  cursor: "pointer",
   position: "relative",
   display: "flex",
   alignItems: "center",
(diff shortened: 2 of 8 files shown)


packages/components/src/SidenavHeader/SidenavHeader.tsx
```diff
@@ -135,10 +141,13 @@ class SidenavHeader extends React.Component<Props, State> {
               css={this.props.css}
               color={this.props.color}
               isActive={isActive}
-              className={this.props.className}
+              className={[this.props.className, "op_sidenavheader"].filter(cls => Boolean(cls)).join(" ")}
```
TejasQ: Why are we using classnames again?

packages/components/src/SidenavItem/SidenavItem.tsx
```diff
@@ -32,11 +32,12 @@ const containerStyles = ({ theme, isActive }: { theme: Theme; isActive: boolean
...
   justifyContent: "flex-start",
   whiteSpace: "nowrap",
-  fontSize: 12,
+  fontSize: 14,
+  fontWeight: 400,
   // Specificity is piled up here to override default styles
   "a:link&, a:visited&": {
     textDecoration: "none",
-    color: isActive ? theme.colors.linkText : "#666666",
+    color: isActive ? theme.colors.linkText : "#888",
```
TejasQ: `/** @todo */` for hex colors please.

packages/theme/src/index.ts
```diff
@@ -174,10 +174,10 @@ const operational: Theme = {
   shadows,
   colors,
   spacing: 16,
-  box: 56,
+  box: 72,
```
TejasQ: What's this?

packages/components/src/SidenavHeader/SidenavHeader.tsx
peterszerzo: In this particular case, a child of this container should only display when `.op_sidenavheader` is hovered. The alternative would be keeping track of hover in state.

In general, classnames are used in rare cases when selecting a parent or a child in a component is tricky or is likely to break. For example `& > div` will break when the tag name changes, and unfortunately cannot be solved with and `& > *` because that doesn't increase specificity, hence doesn't work as an override (we have the same problem in this particular case). Classnames should only show up for one or two special rules in any given component style object.

package.json
```diff
@@ -11,7 +11,7 @@
     "build": "cd packages/website && npm run build",
     "bootstrap": "lerna bootstrap",
     "postinstall": "npm run bootstrap",
-    "clean": "rm -rf node_modules && rm -rf packages/*/node_modules",
+    "clean": "rm -rf node_modules && rm -rf packages/*/node_modules && rm -f packages/*/package-lock.json && rm -f packages/*/yarn.lock && rm -f yarn.lock",
```
fabien0102: https://github.com/lerna/lerna#clean ?

peterszerzo: Good point, is it ok if we keep it for later? Don't really want to spend time testing it since it works.

packages/components/src/SidenavHeader/SidenavHeader.tsx
TejasQ: Let the record show that this is smelly to me.

packages/theme/src/index.ts
TejasQ: Can we at least add a `@todo` then?

PR #473, opened 2018-06-01
Sidenav headers open by default

peterszerzo:
Sidenav headers should be open by default - they can be closed either by clicking on the header text or the collapse button on the right.

screen shot 2018-06-01 at 1 03 19 pm


peterszerzo:
Abandoning this per design discussion with @micha-f

PR #479, opened 2018-06-05
Simplify and improve visualisation typings.

ImogenF:


TejasQ:
Let's try to never have a PR this big agian.

ImogenF:
@tejas sure. This one was just a lot of very minor, similar changes so didn't make much sense to me to split it up.

packages/visualizations/src/Chart/canvas.ts
```diff
@@ -273,13 +262,13 @@ class ChartCanvas implements Canvas {
...
   private margin(axis: AxisPosition): number {
-    const margins: { [key: string]: number } = this.state.current.get("computed").axes.margins || {}
+    const margins = this.state.current.get("computed").axes.margins || {}
     return margins[axis] || 0
   }
 
   private calculateDrawingContainerDims(): Dimensions {
     const config = this.state.current.get("config")
-    const dims: Dimensions = {
+    const dims = {
```
micha-f: I would argue that in this case it is nicer to use 
```
const dims: Dimensions = { ... } 
```
and avoid the return type in the method signature. That way, `dims` passed to `this.stateWriter` (and any other things you do with them inside `calculateDrawingContainerDims` all have the type `Dimensions`.

On another note - `calculateXYZ` does not sound like a method that has a side-effect (storing the value using the stateWriter). Is there a way to make this function pure and call the stateWriter somewhere else? Can we separate calculation and storing the value? I guess this investigation would start with examining from where this method is called.

Same comments apply for `calculateDrawingDims` below.

micha-f:
Nice 👍

PR #480, opened 2018-06-05
Fix d3 utils for path attrTween transitions.

ImogenF:


packages/visualizations/src/utils/d3_utils.ts
```diff
@@ -70,16 +70,19 @@ export const setPathAttributes: AttributeSetter<{
   fill: string | any
   stroke: string | any
   opacity: number | any
+  isTween: boolean
```
micha-f: Can we explain why `isTween` is required?

ImogenF: Have added a comment

micha-f:
Looks good 👍

PR #481, opened 2018-06-05
Chart renderer opacity

ImogenF:


packages/visual-tests/src/TestCases/Chart/symbol-1.ts
```diff
@@ -9,11 +9,12 @@ const createRenderer = (options: any) => {
       size: options.size || ((series: any, d: any) => 50),
       stroke: options.stroke || ((series: any, d: any) => series.legendColor()),
       symbol: options.symbol || ((series: any, d: any) => "circle"),
+      opacity: options.opacity || ((series: any, d: any) => 0.8),
     },
   }
 }
 
-const createData = (options: any) => {
+const createData = (options: any): any => {
```
micha-f: Why does this need to return `any`? Can we not create something that is compatible with what `viz.data` expects?

ImogenF: You're right, it doesn't need to - there was something weird going on as a result of the multiple branches I'm working on but I've sorted it now.

micha-f:
Nice 👍

PR #484, opened 2018-06-06
AvatarGroup improvement

fabien0102:
## Issue
#477 

## Sections to review
https://deploy-preview-484--operational-ui.netlify.com/docs/#avatar
https://deploy-preview-484--operational-ui.netlify.com/docs/#avatargroup

PR #485, opened 2018-06-07
Fix tslint errors

fabien0102:
Some tslint was in this file. I've just fixed these errors and improve a little bit type :angel:

Normally, Travis should passed properly after this PR


packages/visual-tests/src/TestCases/Chart/symbol-3.ts
```diff
@@ -1,44 +1,43 @@
...
+}
+
+type Country = keyof typeof generators
+
+const dataPoint = (country: Country) => {
   return generators[country]()
 }
 
-const generateData = (country, amount = 100) => {
-  let result = []
```
fabien0102: Even if I understand the intention, tslint was complaining about this line. Indeed `result` is never reassigned so it should be a `const` :wink: (`Array.push` is just a method that mutate the array)

fabien0102:
Pfff… out of memory randomly on Travis… 😕

packages/visual-tests/src/TestCases/Chart/symbol-3.ts
micha-f: looks good to me @fabien0102

fabien0102:
Part of #484 (rebased to be able to have success tests)

PR #487, opened 2018-06-11
Chart focus customisation

ImogenF:
https://visual-tests-focus-customisation.fe.ctmo.io

### DIFF
--- a/packages/visualizations/src/Chart/facade.ts
+++ b/packages/visualizations/src/Chart/facade.ts
@@ -37,7 +37,7 @@ const defaultConfig = (): ChartConfig => {
     maxBarWidthRatio: 1 / 3,
     maxFocusLabelWidth: 350,
     minBarWidth: 3,
-    numberFormatter: (x: number): string => x.toString().replace(/\B(?=(\d{3})+(?!\d))/g, ","),
+    numberFormatter: (x: number): string => (+x.toFixed(2)).toString().replace(/\B(?=(\d{3})+(?!\d))/g, ","),
     outerBarSpacing: 10,
     palette: theme.colors.visualizationPalette,
     showComponentFocus: false,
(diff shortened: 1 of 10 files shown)


ImogenF:
[Charts - customisable focus labels](https://trello.com/c/VB2HXoUe/188-charts-customisable-focus-labels)

micha-f:
screen shot 2018-06-13 at 10 41 34

@ImogenF  There seems to be a problem with the formatting of the numbers in the label.
Also - can we add the country to the label?

ImogenF:
@micha-f is that in the website example or in the visual specs? Either way, that should be fixed...  Have just updated the deployed visual tests. 

In terms of number formatting though, the default formatter basically assumes there won't be more than 3 decimal places. It formats everything after the decimal exactly the same as everything before. I'm not sure how much logic we want to have in the default number formatter...  should I add default rounding?

Will add the countries now.

packages/visualizations/src/Chart/facade.ts
```diff
@@ -37,7 +37,7 @@ const defaultConfig = (): ChartConfig => {
     maxBarWidthRatio: 1 / 3,
     maxFocusLabelWidth: 350,
     minBarWidth: 3,
-    numberFormatter: (x: number): string => x.toString().replace(/\B(?=(\d{3})+(?!\d))/g, ","),
+    numberFormatter: (x: number): string => (+x.toFixed(2)).toString().replace(/\B(?=(\d{3})+(?!\d))/g, ","),
```
micha-f: Is there a reason not to use `d3.format` ?

ImogenF: No, no reason

micha-f: It might be more robust / simpler than writing our own code here.

PR #491, opened 2018-06-12
[WIP] Proposed style constants implementation.

ImogenF:


packages/theme/src/index.ts
```diff
@@ -1,184 +1,142 @@
...
-  navBackground: string
-  text: string
-  lightText: string
-  linkText: string
-  border: string
-  inputBorder: string
-  separator: string
+// Default constants
+const buttons = {
+  grey: {
```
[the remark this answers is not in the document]
micha-f: That works 👍

packages/theme/src/index.ts
```diff
@@ -1,184 +1,142 @@
...
+  blue: {
+    active: {
+      borderRadius: "2px",
+      backgroundColor: "#1499CE",
+      color: "#fff",
+    },
+  },
+  ghost: {
+    active: {
+      borderRadius: "2px",
```
fabien0102: You can write `2` here (the basic number is in px by default)

packages/theme/src/index.ts
```diff
@@ -1,184 +1,142 @@
...
-  border: string
-  inputBorder: string
-  separator: string
+// Default constants
+const buttons = {
+  grey: {
+    active: {
+      borderRadius: "2px",
+      backgroundColor: "#e2e2e2",
+      color: "#1499CE",
```
fabien0102: Every colors over there should reference something into `fontColor` (to avoid any duplication) :wink:

micha-f: @fabien0102 I disagree. We do not use some of these colors as font colors. I think it's perfectly fine to define the button colors without referencing anything. They are such basic elements of the UI.

TejasQ:
Closing in favor of #492

PR #492, opened 2018-06-12
Update style constants

TejasQ:
In this PR, I've added a file that contains styling constants per the new design. I've removed _specific_ styles for buttons, cards, etc. and only included the things that are general and expected to be reused throughout Operational UI.

Additionally, I've re-added the theme things that were removed in order to prevent breakages throughout the Operational component library.


packages/theme/src/constants.ts
```diff
@@ -0,0 +1,89 @@
...
+ * allow for fine tuning and allow a more transparent way to
+ * observe color variation, verbosely describing the hue,
+ * saturation, and lightness of a color.
+ *
+ * hsla is used where alpha blending is involved.
+ */
+export const color = {
+  primary: {
+    main: "hsl(197, 82%, 44%)",
+    contrast: "hsl(0, 0, 100%)",
```
ImogenF: What are the `contrast` values used for?

packages/theme/src/constants.ts
```diff
@@ -0,0 +1,89 @@
...
+  grey: (lightness: 96 | 93 | 56 | 45 | 40 | 33 | 24 | 20) => ({
+    main: `hsl(0, 0, ${lightness}%)`,
+
+    /**
+     * This is a _really_ naïve contrast check.
+     * @todo reconsider this as needs arise.
+     */
+    contrast: lightness > 50 ? "hsl(0, 0, 0)" : "hsl(0, 0, 100%)",
+  }),
+}
```
ImogenF: Do we also want other standard colours, namely `white` and possibly `black`?

packages/theme/src/constants.ts
```diff
@@ -0,0 +1,89 @@
...
+    fineprint: 12,
+  },
+}
+
+export const space = {
+  wide: 28,
+  default: 20,
+  content: 16,
+  basic: 8,
+}
```
ImogenF: I've been looking through the components and a lot of them manipulate the theme spacing quite heavily - things like `theme.spacing * 0.75`. Are we trying to fully get rid of that now, and just use the closest of the above options?

packages/theme/src/constants.ts
ImogenF: For context, the `page` component for instance explicitly has a white background. Haven't checked how widely these colours were used from the theme though.

packages/theme/src/constants.ts
TejasQ: For buttons and anything involving a background color. So for `primary`, it's blue and its respective `contrast` is white, according to this snippet.

That means, when we have a `Button` with a blue background, the provided `contrast` value can be used for the text that _goes on_ the button (in this case, white) in order to tightly couple a harmonious complement.

This will allow us to design things like buttons, cards, etc. with a consideration for _"okay, the background is dark red... what color should the text be?"_ already taken care of.

packages/theme/src/constants.ts
TejasQ: That's a great idea. Absolutely. Will add these.

packages/theme/src/constants.ts
TejasQ: Yes. According to @micha-f, some things, design-wise, will need to be fine tuned and just plain old explicit magic numbers, but by and large, we'd like to use these (or multiples of 8) where possible if I understand correctly.

Do I understand correctly, @micha-f?

micha-f: That would be the idea - unless it looks bad. In that case we have to use custom fine-tuning. 

Also I would set 1-4px paddings / margins directly in the components. They are quite common and I don't think we should define those as a calculation of any base spacing. They look good the way they are and would not need to adjust if the spacing is changed.

ImogenF: Ok, nice. So for components that currently have large calculated margins or padding, we choose one of the available options (i.e. 24 could go to either `default` or `big`, depending on what it looks like), and for small margins or paddings we set them explicitly. That works.

micha-f: @TejasQ I'm not sure it's clear what the difference between `default` and `basic` is.

How about we use these?
`base`: 4
`small`: 8
`content`: 16
`element`: 20
`big`: 28

micha-f: image

As an example of the "1-4px rule" above, I don't think it would make sense to define the spacing between the pagination buttons as a fraction of the base spacing. We don't want this spacing to change if the spacing is changed. It looks good the way it is, regardless of what the theme spacing is.

TejasQ: The 1-4px rule is clear.

That said, I don't quite understand the difference [here](https://github.com/contiamo/operational-ui/pull/492#discussion_r194801151) between `content` and `element`.

micha-f: Was thinking that `content` would mean "inside of an `element`". Happy to use a better naming!

micha-f:
@TejasQ This is nice. How about we combine this approach with one aspect of the other approach?
It'd be nice to have an idea about which grays one is supposed to use for fonts, backgrounds and separators. For example, a `shades` function could limit to a subset of the grays.

packages/theme/src/constants.ts
micha-f: That was the idea - not sure it's clear though.

micha-f:
These would give us a name for all the greys we use and avoid hardcoding a number into our components. 
```
type Grey =
    96 |    // #f6f6f6 
    93 |    // #ececec
    91 |    // #e8e8e8
    75 |    // #c0c0c0
    56 |    // #909090
    45 |    // #747474
    40 |    // #666
    33 |    // #545454
    24 |    // #3e3e3e
    20      // #333

const colors = {
    grey: (lightness: Grey) => ({
        main: `hsl(0, 0, ${lightness}%)`,

        /**
         * This is a _really_ naïve contrast check.
         * @todo reconsider this as needs arise.
         */
        contrast: lightness > 50 ? "hsl(0, 0, 0)" : "hsl(0, 0, 100%)",
    })
}

const shades = {
    dark: colors.grey(24),
    light: colors.grey(93),
    lighter: colors.grey(96)
}

const separators = {
    "default": colors.grey(91),
    light: colors.grey(93)
}

const fonts = {
    dark: colors.grey(20),
    "default": colors.grey(33),
    light: colors.grey(40),
    lighter: colors.grey(45),
    lightest: colors.grey(56)
}

const inputBorder = {
    "default": colors.grey(75),
    disabled: colors.grey(91)
}
```

packages/theme/src/constants.ts
```diff
@@ -68,6 +83,47 @@ export const color = {
   }),
 }
 
+/**
+ * A specialized color palette for shadows.
+ */
+export const shadowColor = {
```
micha-f: These would be use for backgrounds. Should we call them `backgrounds`?

micha-f:
Looks good!

TejasQ:
[Remove themability of Operational](https://trello.com/c/06WFxrhp/196-remove-themability-of-operational)

packages/components/src/OperationalUI/OperationalUI.tsx
```diff
@@ -1,11 +1,11 @@
 import * as React from "react"
+
+/** @todo rename this import once we're fully in emotion */
+import { ThemeProvider as EmotionThemeProvider } from "emotion-theming"
+
+/** @todo remove this once we're fully in emotion */
 import { ThemeProvider } from "glamorous"
```
fabien0102: You can rename this to `GlamorousThemeProvider`

packages/components/src/OperationalUI/OperationalUI.tsx
```diff
@@ -25,21 +25,22 @@ export interface Props {
...
+          <React.Fragment>
+            {props.withBaseStyles ? (
+              <style
+                dangerouslySetInnerHTML={{
+                  __html: baseStylesheet(props.theme || operational),
+                }}
+              />
+            ) : null}
+            {props.children}
+          </React.Fragment>
```
fabien0102: Why not the classic and beautiful `<>` instead of `<React.Fragment>`?

packages/components/src/Status/Status.tsx
```diff
@@ -27,7 +26,11 @@ const Status = styled("div")`
   width: 8px;
   height: 8px;
   border-radius: 50%;
-  box-shadow: 0 0 4px ${getColorFromProps}99;
+  box-shadow: 0 0 4px
+    ${(props: Props) =>
+      tinycolor(getColorFromProps(props))
```
fabien0102: 😍

packages/components/src/OperationalUI/OperationalUI.tsx
TejasQ: Great idea!

PR #494, opened 2018-06-13
Update scatterplot example with axis titles and rotated axis labels.

micha-f:
Chart bottom padding seems to be computed incorrectly after data update.

image


ImogenF:
@micha-f Should be fixed now, have tested pretty thoroughly.

micha-f:
@ImogenF Nice work! Not able to break it anymore 👍

PR #495, opened 2018-06-13
Add Status component 

TejasQ:
This PR contains a Status component to indicate whether a component is online or offline or whatever. It's supposed to look like this from the design:

Status

## Important
This PR also is the _start_ of what will be an eventual migration to [👩‍🎤emotion](https://emotion.sh) since [💄 glamorous](https://github.com/paypal/glamorous) has been [deprecated in favor of it](https://github.com/paypal/glamorous/issues/419) by PayPal et al. 

We'd like to do this in the _most_ backwards compatible, scalable way, and so I've proposed to start with it for this trivial component.

I suggest we discuss further (if required) on Monday during the team meeting.

### DIFF
--- a/packages/components/src/types.ts
+++ b/packages/components/src/types.ts
@@ -1,9 +1,9 @@
 import { Theme } from "@operational/theme"
 import { CSSProperties } from "glamorous"
-import { Interpolation } from "create-emotion"
 
 export type WithTheme = { theme: Theme }
 
+/** @todo Rethink this line with Emotion */
 export type Css = any
 
 // TODO: make this stricter to disallow { margin: 20, fruit: "much much" }
--- a/packages/components/src/types.ts
+++ b/packages/components/src/types.ts
@@ -1,9 +1,10 @@
 import { Theme } from "@operational/theme"
 import { CSSProperties } from "glamorous"
+import { Interpolation } from "create-emotion"
 
 export type WithTheme = { theme: Theme }
 
-export type Css = CSSProperties | (<T>(props: T & WithTheme) => CSSProperties)
+export type Css = any
 
 // TODO: make this stricter to disallow { margin: 20, fruit: "much much" }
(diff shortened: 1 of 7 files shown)


packages/components/src/types.ts
```diff
@@ -1,9 +1,10 @@
 import { Theme } from "@operational/theme"
 import { CSSProperties } from "glamorous"
+import { Interpolation } from "create-emotion"
 
 export type WithTheme = { theme: Theme }
 
-export type Css = CSSProperties | (<T>(props: T & WithTheme) => CSSProperties)
+export type Css = any
```
TejasQ: We'll have to rethink this as emotion gains more ground in the codebase.

TejasQ:
[Add status indicator component](https://trello.com/c/fXDC6d1W/192-add-status-indicator-component)

ImogenF:
@TejasQ I assume you've seen this thread about deprecating glamorous in favour of emotion? https://github.com/paypal/glamorous/issues/419. Some good tips in there for making the migration easier. 

From the looks of things, it wouldn't actually be that big a job to migrate everything at once. I'm wondering if that would make more sense than introducing it gradually and having a mix of glamorous and emotion in the codebase?

TejasQ:
@ImogenF totally. I actually cited and linked to that same issue in my initial comment at the opening of this PR. 

I agree about the size of the job. I think it should be fairly straightforward to move over, _but_, we have some complex things going on in Operational with themes and things. Currently, I am not as aware of Emotion's best practices and theming functionality as I should be. 

I think once we _know_ Emotion well enough, we'll be more equipped to do it. I'm currently looking to read up a bit on it before jumping in. I'd suggest we look into it, its features and API together as a team (cc @fabien0102) and then discuss and decide on a migration strategy on Monday at the JF? 

This way, we'll only have a mix of glamorous and emotion for today and tomorrow.

What do you think?

PR #496, opened 2018-06-14
Update default number formatter using d3-format.

ImogenF:
Using the d3`~` format option with `".2f"` to strip trailing zeros doesn't work, because `".2f"` defines a certain number of decimal places, which `~` would then potentially remove. 

So now the default number formatter rounds the given number to 2dp, converts it from a string back to a number (hence removing any trailing zeros) and then adds commas between the thousands.


micha-f:
Can we test that this number -> string -> number conversion works correctly in all cases?
Since this is used in all charts, I believe it makes sense to avoid some unnecessary function calls in this case:

```
const defaultNumberFormatter = compose(
  format(","),
  (val) => +val,
  format(".2f")
)

{
  (...)
  numberFormatter: defaultNumberFormatter
}
```

micha-f:
great 👍

PR #497, opened 2018-06-14
Feature/resource list

fabien0102:
## Table component

Redesign of the component using the classic `<table>` for accessibility.

The design was updated following the contiamo design spec but without breaking the public API of the component.

## Docs link

https://deploy-preview-497--operational-ui.netlify.com/docs/#table

## Notes

This component is using emotion and the new constants file/system. So I'm waiting for your feedback on this new kind of implementation.

I also update the theme wrapper for the unit test, to have access to the theme variable even in our test :wink:


packages/components/src/Table/Table.tsx
```diff
@@ -1,156 +1,113 @@
...
+  },
+}))
 
-const TableCell = glamorous.div(
-  ({ theme }: WithTheme): {} => ({
-    ...theme.typography.body,
-    flex: 1,
-  }),
-)
+const Action = styled(Td)(({ theme }: CompProps) => ({
```
TejasQ: Can we name this `TextAction`?

packages/components/src/Table/Table.tsx
```diff
@@ -1,156 +1,113 @@
...
-    top: theme.spacing,
-    textAlign: "center",
-    backgroundColor: theme.colors.background,
-    ...theme.typography.body,
-  }),
-)
+const Actions = styled(Td)(({ theme }: CompProps) => ({
+  textAlign: "right",
+  paddingRight: theme.space.small,
+  opacity: 0,
```
TejasQ: Can we leave a comment here about _why_ opacity?

packages/components/src/Table/Table.tsx
```diff
@@ -1,156 +1,113 @@
...
-    width: "fit-content",
-    position: "absolute",
-    top: "50%",
-    right: theme.spacing / 2,
-    transform: "translate3d(0, -50%, 0)",
-    display: "none",
-  }),
-)
+const EmptyView = styled("div")(({ theme }: CompProps) => ({
+  color: theme.color.text.default || "red",
```
TejasQ: Or red? :D

packages/components/src/Table/Table.tsx
```diff
@@ -1,156 +1,113 @@
...
-          </TableBodyRow>
-        ))}
-      </TableBody>
-    </TableElement>
-    {props.rows.length === 0 ? <EmptyView>There are no records available</EmptyView> : null}
-  </Container>
-)
+const Table: React.SFC<Props> = ({ rows, columns, onRowClick, rowActionName, __experimentalRowActions, ...props }) => {
+  return (
+    <>
```
TejasQ: We can get rid of this and handle the empty dataset logic better.

PR #498, opened 2018-06-14
Visualizations tests

ImogenF:
Add/improve visual tests and fix any issues that become apparent.

See Trello for list of visual tests added.

Fixed:

- Axis positioning with titles, rotated labels, data updates etc. Should now be working in all cases.
- Axis rules: for categorical and time axes, if the data is rendered as bars, the rules are drawn *between* the ticks. If there is no bars renderer (i.e. anything else: line, area, symbols etc), the rules are drawn *on* the ticks.
- Rendering of bars as ranges.
- Area renderer displays vertical line in case of only a single data point.

Looks like quite a big PR but most of it is just visual test cases, i.e. no complicated code and can be quickly checked visually. 

https://visual-tests-498.fe.ctmo.io


ImogenF:
[Improve and expand visual tests](https://trello.com/c/DjcDgrT1/195-improve-and-expand-visual-tests)

ImogenF:
[Adapt line and area renderers to render single points](https://trello.com/c/XpY8vz5V/197-adapt-line-and-area-renderers-to-render-single-points)

micha-f:
Awesome to have all these visual tests in there. And as expected we discovered some small issues :)

PR #503, opened 2018-06-19
Fix internal style adjustments

peterszerzo:
This PR replaces

```
<Input css={{ width: "100%" }} />
```

with 

```
const FullWidthInput =styled(Input)`
  width: 100%;
`

<FullWidthInput />
```

To fix all internal broken styles from the emotion upgrade (8 locations total).

### DIFF
--- a/packages/components/src/Breakdown/Breakdown.tsx
+++ b/packages/components/src/Breakdown/Breakdown.tsx
@@ -164,7 +164,7 @@ const Number = styled("div")(
   }),
 )
 
-const Span = styled("span")()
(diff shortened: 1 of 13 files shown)


packages/components/src/Breakdown/Breakdown.tsx
```diff
@@ -164,7 +164,7 @@ const Number = styled("div")(
   }),
 )
 
-const Span = styled("span")()
+const Span = styled("span")({ fontFeatureSettings: "'tnum'" })
```
TejasQ: Object syntax here but string literal syntax elsewhere? 🤔

PR #505, opened 2018-06-19
Chip redesign.

ImogenF:
- [x] `props.color` can be used to assign a color directly, using a name of a color defined in the constants (e.g. `primary`), or a hex string.
- [x] General style improvements have been made, including removing all references to the deprecated theme. 
- [x] Update website

### DIFF
--- a/packages/components/src/Chip/Chip.tsx
+++ b/packages/components/src/Chip/Chip.tsx
@@ -1,10 +1,9 @@
 import * as React from "react"
 import styled from "react-emotion"
-import { darken, readableTextColor, expandColor } from "@operational/utils"
-import { OperationalStyleConstants, Theme } from "@operational/theme"
+import { readableTextColor, expandColor } from "@operational/utils"
+import { OperationalStyleConstants } from "@operational/theme"
 import { Icon, IconName } from "../"
-import { isWhite } from "../utils/color"
-import { WithTheme, Css, CssStatic } from "../types"
+import { WithTheme, Css } from "../types"
 
 export interface Props {
   /** Id */
--- a/packages/components/src/Chip/Chip.tsx
+++ b/packages/components/src/Chip/Chip.tsx
@@ -1,7 +1,7 @@
 import * as React from "react"
 import styled from "react-emotion"
-import { darken, readableTextColor } from "@operational/utils"
-import { OperationalStyleConstants, Theme, expandColor } from "@operational/theme"
+import { darken, readableTextColor, expandColor } from "@operational/utils"
+import { OperationalStyleConstants, Theme } from "@operational/theme"
 import { Icon, IconName } from "../"
 import { isWhite } from "../utils/color"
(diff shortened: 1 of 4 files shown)


micha-f:
@TejasQ I think that there should be two ways to use Chips:
1) Assigning a specific color
2) Not assigning a color -> a color is selected based on the content of the chip, so we need a mapping function that maps a certain content to the same color every time. Here is the color assignment in avatars as a reference: https://github.com/contiamo/operational-ui/blob/master/packages/components/src/Avatar/Avatar.tsx#L130

We will not offer the option to choose a color to users of our platform initially. I believe 2) will help us a lot in this case.

micha-f:
@ImogenF There should also be a way to map any string to a limited number of colors in a stable way.

Let's say we hash the content. Then take the first character of that hash and map 36 letters and numbers (a-z, 0-9) to 12 colors (three characters map to each color).

ImogenF:
@micha-f Sure, we can do that. But if we have 5 different tags, which happen to all start with the letter "a", they would all be the same colour. Does that matter? It may be that the tags are so similar that the user would like different colours to be able to distinguish more easily (e.g. "project1", "project2")

micha-f:
@ImogenF if we hash them then they should not start with the same letter:
```
micha@beach | ~ $md5 -s "test"
MD5 ("test") = 098f6bcd4621d373cade4e832627b4f6
micha@beach | ~ $md5 -s "test1"
MD5 ("test1") = 5a105e8b9d40e1329780d62ea2265d8a
micha@beach | ~ $md5 -s "test2"
MD5 ("test2") = ad0234829205b9033196ba818f7a872b
```

ImogenF:
@micha-f I see - I'm not super familiar with hashes. In that case, that looks like a really nice solution. I'll do that now!

micha-f:
I know that there are JS implementations of md5. But there might be simpler hash algorithms, too. Might be worth doing a few mins of research.

TejasQ:
I'm having a hard time believing what I just read.

We're including a dependency on a hashing algorithm in order to calculate a color that _ought to be provided by an API in the first place_.

Does this not seem like overkill to anyone else here? For tag colors? Seriously?
@fabien0102 @peterszerzo thoughts? Surely I can't be the only one who thinks this is a bit much. _And_ we want to move fast and get a refreshed design out there? Really?

If no color is given, I'd say use the primary as a default. If consumers want custom colors, they can implement their own hashing algorithms or whatever they need to determine colors and pass them through the color prop.

I strongly believe that this ought not be the concern of this component and that we'd be making unnecessary engineering choices to include such logic.

micha-f:
@TejasQ I agree 👍

fabien0102:
Totally agree, just keep it simple, for now the "contiamo blue" is a good default for me, and for the future we can add the avaibility add custom color to tags, so just take the user color.

And btw, the md5 is not really secured, I will prefer sha512! 😇 (I'm joking, if it's not obvious for everyone ^^)

Issue #506, opened 2018-06-19
Paginator


ImogenF:
![image](https://user-images.githubusercontent.com/14272822/41592375-a51c7f86-73bc-11e8-8380-97c8fbd955c2.png)

## Proposed API
```jsx
<Paginator
  itemCount={258}
  page={this.state.page}
  itemsPerPage={50} 
  onChange={page => this.handleChange(page)} 
/>
```

The "first"/"prev" and "next"/"last" buttons only show up if the first/last items are not being displayed.

TejasQ:
I believe you have the wrong screenshot...

ImogenF:
https://trello.com/c/Fku5UWiB/191-pagination-component

TejasQ:
Great API! I think conditionally rendered buttons based on props could be nice: only show a next button if given an `onNext` prop and so on.

ImogenF:
@TejasQ You mean replace `onChange` with `onNext`, `onFirst`, `onPrev` and `onLast`? That would be nice, but I think the user shouldn't have to determine whether page 5 is the last page or not and only pass an `onLast` prop accordingly. How about only showing a next button if an `onNext` prop is provided *AND* if `props.itemCount` < `props.page` * `props.itemsPerPage` ?

ImogenF:
#507

ImogenF:
![](https://github.trello.services/images/mini-trello-icon.png) [Pagination component](https://trello.com/c/Fku5UWiB/191-pagination-component)

fabien0102:
Regarding the `onChange` vs `onNext/onPrev`, I also think that the `onNext` can be interesting, maybe keep the both options for now (it's really depends how we deal the pagination with the backend, so both options can be nice here)

But yes, what even we choose, we need to disable button regarding the state (disable first if we are of first page, etc) :wink:

ImogenF:
@fabien0102 The disabling is in there already.

I really think `onNext/onPrev` would overcomplicate things and I'm not sure I see any benefit to it - in all cases, we just want to change `state.page`, so either we keep the logic in the `Paginator` component, as it is now, or we move it out into the `onNext` etc methods and trust the user to set it up correctly.

TejasQ:
Merged.

PR #507, opened 2018-06-19
Paginator redesign.

ImogenF:
- Navigation through buttons with `first`, `prev`, `next` and `last` labels rather than clickable icons.
- Navigation buttons only displayed if they will actually change the items displayed (i.e. if items 1 - 50 are being displayed, the `first` and `prev` will not be rendered)
- Displays range of item numbers that are being displayed, rather than page numbers

### DIFF
--- a/packages/components/src/Paginator/Paginator.tsx
+++ b/packages/components/src/Paginator/Paginator.tsx
@@ -74,30 +74,40 @@ const PaginatorControl = ({ children, itemCount, itemsPerPage, page, onChange, t
     onChange && onChange(Math.ceil(itemCount / itemsPerPage))
   }
 
-  let handler
-
   switch (type) {
     case "previous":
-      handler = handlePrevious
-      break
+      return (
+        <NavigationButton condensed onClick={handlePrevious}>
+          {children}
+        </NavigationButton>
+      )
 
     case "first":
-      handler = handleFirst
-      break
+      return (
+        <NavigationButton condensed onClick={handleFirst}>
+          {children}
+        </NavigationButton>
+      )
 
     case "next":
-      handler = handleNext
-      break
+      return (
+        <NavigationButton condensed onClick={handleNext}>
+          {children}
+        </NavigationButton>
+      )
+
+    case "last":
+      return (
+        <NavigationButton condensed onClick={handleLast}>
+          {children}
+        </NavigationButton>
+      )
 
     default:
-      handler = handleLast
+      throw new Error(
+        "No handler specified for NavigationButton in Paginator component\nPlease refer to the docs: http://operational-ui.js.org/docs/#paginator",
+      )
   }
-
-  return (
-    <NavigationButton condensed onClick={handler}>
-      {children}
-    </NavigationButton>
-  )
 }
 
 const getRange = ({ page, itemCount, itemsPerPage }: Props) => {
--- a/packages/components/src/Paginator/Paginator.tsx
+++ b/packages/components/src/Paginator/Paginator.tsx
@@ -49,10 +49,12 @@ interface ControlProps {
 const NavigationButton = styled(Button)(({ theme }: { theme?: OperationalStyleConstants }) => ({
   width: 56,
   marginRight: 3,
-  padding: `0 ${theme.space.base}px`,
+  padding: 0,
   "& svg": {
     verticalAlign: "middle",
-    margin: "0 0 0 3px",
+  },
+  "& span": {
+    padding: "0 3px",
   },
 }))
 
@@ -106,7 +108,8 @@ const Paginator: React.SFC<Props> = props => {
         first
       </PaginatorControl>
       <PaginatorControl type="previous" {...controlProps} isDisabled={isFirstDisabled}>
-        <Icon.ChevronsLeft size="11" /> <span>prev</span>
+        <Icon.ChevronsLeft size="11" />
+        <span>prev</span>
       </PaginatorControl>
       <PaginatorSpan key={props.page}>
         <span>{getRange({ page: props.page, itemCount: props.itemCount, itemsPerPage: props.itemsPerPage })}</span> of{" "}
--- a/packages/components/src/Paginator/Paginator.tsx
+++ b/packages/components/src/Paginator/Paginator.tsx
@@ -3,25 +3,20 @@ import * as Icon from "react-feather"
 import styled from "react-emotion"
 import { OperationalStyleConstants } from "@operational/theme"
 import { Css, CssStatic } from "../types"
-import { Button } from "../"
+import Button from "../Button/Button"
 
 export interface Props {
   id?: string
   /** `css` prop as expected in a glamorous component */
-
   css?: Css
   className?: string
   /** Function to be executed after changing page */
-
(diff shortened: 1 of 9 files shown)


ImogenF:
#506

ImogenF:
@TejasQ I've changed my mind about getting rid of the `onChange` method. I think it's much easier for the user to only have to pass one method to the `Paginator` component, rather than 4.

ImogenF:
[Pagination component](https://trello.com/c/Fku5UWiB/191-pagination-component)

TejasQ:
[The docs website for this PR](https://deploy-preview-507--operational-ui.netlify.com/docs/) is broken.

ImogenF:
@TejasQ That's why it's still WIP 😉 Working on that now.

ImogenF:
@fabien0102 can you check I implemented the `defaultProps` correctly? I'd like to get this merged soon and out of the way, if possible.

packages/components/src/Paginator/Paginator.tsx
```diff
@@ -58,36 +46,34 @@ interface ControlProps {
...
   page: number
-  pageCount: number
+  itemCount: number
+  itemsPerPage: number
   type: "first" | "previous" | "next" | "last"
 }
 
-const PaginatorControl = ({ children, onChange, pageCount, page, type }: ControlProps) => {
+const NavigationButton = styled(Button)(({ theme }: { theme?: OperationalStyleConstants }) => ({
+  width: "56px",
```
TejasQ: Nit: `56` as a numeric value works here too.

packages/components/src/Paginator/Paginator.tsx
```diff
@@ -181,34 +116,47 @@ const Container = styled("div")({
...
   const controlProps = {
-    pageCount,
-    page,
-    onChange,
+    itemCount: props.itemCount,
+    itemsPerPage: props.itemsPerPage,
+    page: props.page,
+    onChange: props.onChange,
   }
+  const displayFirst = props.page !== 1
```
TejasQ: Can we call these `shouldDisplayFirst`? They would help newer developers to infer that they're conditionally set variables, I think.

ImogenF:
@TejasQ nits are fixed.

TejasQ:
Thanks! The [live preview](https://deploy-preview-507--operational-ui.netlify.com/docs/#headerbar) also seems to be broken so we can't review the design...

packages/components/src/Paginator/Paginator.tsx
```diff
@@ -16,40 +17,27 @@ export interface Props {
...
-  maxVisible?: number
   /** Index of the current selected page */
 
   page?: number
-  /** Total number of pages */
+  /** Total number of items */
 
-  pageCount: number
+  itemCount: number
+  /** Number of items per page */
```
TejasQ: This seems to be a great opportunity to fix the weird line spacing.

packages/components/src/Paginator/Paginator.tsx
```diff
@@ -181,34 +116,47 @@ const Container = styled("div")({
...
+          first
+        </PaginatorControl>
+      )}
+      {shouldDisplayFirst && (
+        <PaginatorControl type="previous" {...controlProps}>
+          <Icon.ChevronsLeft size="11" /> prev
+        </PaginatorControl>
+      )}
+      <PaginatorSpan key={props.page}>
+        <span>{getRange({ page: props.page, itemCount: props.itemCount, itemsPerPage: props.itemsPerPage })}</span> of{" "}
```
TejasQ: Do we need this span?

packages/components/src/Paginator/Paginator.tsx
```diff
@@ -16,40 +17,27 @@ export interface Props {
...
+  /** Total number of items */
 
-  pageCount: number
+  itemCount: number
+  /** Number of items per page */
+
+  itemsPerPage: number
 }
 
 const PaginatorSpan = styled("div")(
```
TejasQ: It's called PaginatorSpan but it's a div...

packages/components/src/Paginator/README.md
```diff
@@ -7,7 +7,7 @@ class ComponentWithPaginator extends React.Component {
   constructor(props) {
     super(props)
     this.state = {
-      page: 1
+      page: undefined,
```
TejasQ: Can we omit this altogether? 🤔

packages/components/src/Paginator/Paginator.tsx
ImogenF: Yes - the content of the span is formatted differently to the rest of the content (darker font)

TejasQ:
After visually reviewing it locally, I find that the _jump_ of the buttons when the first and previous buttons show up makes it fairly hard to use and muddies the experience for me.

Can we, instead of _removing_ the buttons, simply disable them when it doesn't make logical sense to use them?

ImogenF:
@TejasQ - disable, or hide completely (but keep the spacing)?

TejasQ:
I think disable would look better. I can foresee that hiding them but keeping the spacing might cause things to look misaligned. Partially hidden (=> lower opacity => disabled) could work better in this case.

micha-f:
The netlify preview of this seems broken:
```
Uncaught TypeError: Cannot read property '__emotion_forwardProp' of undefined
    at createStyled (bundle.16fa1a08.js:1)
    at Module.<anonymous> (bundle.16fa1a08.js:23)
    ...
```

fabien0102:
image
I think that it should be center ;)

micha-f:
The border colors look a bit dark for my taste (and are different to XD).

TejasQ:
@ImogenF FYI I updated the border colors

TejasQ:
@ImogenF those are the only examples of white buttons we have, so I'd say yes unless @micha-f disagrees.

micha-f:
It looks good!

Just the spacing **after** `prev` and **before** `next` could be a bit bigger?

TejasQ:
@micha-f that'll affect buttons everywhere. Did you want paginator buttons to look different?

ImogenF:
@micha-f Do you mean the spacing after/before the text _inside_ the buttons, or after/before the buttons themselves?

And @TejasQ no, the button heights are wrong - it's a side-effect of taking out the borders. I'll fix those.

packages/components/src/Paginator/Paginator.tsx
TejasQ: How? is there a CSS rule or something for it?

micha-f:
@ImogenF inside of the buttons (padding)

ImogenF:
@micha-f Fixed already 😉

packages/components/src/Paginator/Paginator.tsx
ImogenF: https://github.com/contiamo/operational-ui/pull/507/files#diff-767e9e0a6e00c4eb364a255ae0beb446R56

ImogenF: I could put the span and the arrows icon inside separate styled components but that seemed a bit overkill

Issue #475, opened 2018-06-04
Top bar


TejasQ:
We'd like to redesign the top bar for our Contiamo app frame to look like:

![image](https://user-images.githubusercontent.com/9947422/40923513-46d7f4da-6815-11e8-8974-58207ac8e95c.png)

### Proposed API
```jsx
<HeaderBar
  start={Logo}
  center={props => <ProjectSelector projects={props.projects} />}
  end={UserMenu}
/>
```

The rationale behind `start`, `center`, and `end` props allow for greater i18n in right-to-left settings while maintaining clear descriptions of _what_ content goes where.

## `<ProjectSelector />`
This component allows us to select projects. A project looks like,

### Project Interface
```ts
interface Project {
  id: string // UUID
  name: string
}
```

### Proposed API
```jsx
<ProjectSelector
   projects={[
    { id: "awFAE", name: "hahah" }
   ]}
  onChange={() => {}}
  activeProjectIndex={4}
  {/* Or... */}
  activeProjectId="heaf-aeg-aeg"
/>
```

## `<UserMenu />`
This component has two states: `active` and `inactive`. When `active`, the user is presented with a context menu showing possible user actions: logout, change avatar, etc. When `inactive`, the user just sees his/her Avatar.

### `User`
```ts
interface User {
  id: string
  image: string
  name: string
  email?: string
}
```

### Proposed API
```jsx
<UserMenu
  user={user}
  onLogout={() => {}}
  {/* onEvent for each possible user event */}
/>
```

PR #508, opened 2018-06-19
Add HeaderBar

TejasQ:
Added HeaderBar according to #475


TejasQ:
[Top bar](https://trello.com/c/IEhjqSYn/184-top-bar)

packages/components/src/ContiamoLogo/ContiamoLogo.tsx
```diff
@@ -0,0 +1,83 @@
+import * as React from "react"
+import styled from "react-emotion"
+import { OperationalStyleConstants } from "@operational/theme"
+import { expandColor } from "@operational/utils"
+
+type WithTheme = { theme?: OperationalStyleConstants }
+
+export interface LogoProps {
```
ImogenF: Looks like this is just call `Props` in all of the other components - any reason why it's called `LogoProps` here? If not, I'd rename to `Props`.

packages/components/src/Select/Select.Option.tsx
```diff
@@ -34,39 +33,32 @@ const Container = styled("div")(
...
-      width: theme.deprecated.spacing * 0.75,
-      height: theme.deprecated.spacing * 0.75,
-    },
-  }),
-)
+const IconContainer = styled("div")(({ theme }: { theme?: OperationalStyleConstants & { deprecated: Theme } }) => ({
+  display: "flex",
+  alignItems: "center",
+  justifyContent: "center",
+  width: theme.deprecated.spacing * 1.25,
```
ImogenF: Are we not trying to move away from the deprecated them in favour of the constants? Why not use theme.space.xxx here?

packages/components/src/Select/Select.Option.tsx
```diff
@@ -1,9 +1,8 @@
 import * as React from "react"
 import styled from "react-emotion"
 import { OperationalStyleConstants, Theme, expandColor } from "@operational/theme"
```
ImogenF: Can't see where `expandColor` is being used here but the method of this name in `@operational/theme` is being deprecated in favour of the new version in `@operational/utils`, which has the correct typing.

packages/components/src/Select/Select.tsx
```diff
@@ -1,10 +1,10 @@
 import * as React from "react"
 import styled from "react-emotion"
-import { WithTheme, Css } from "../types"
+
```
ImogenF: Unnecessary blank line 😉

packages/components/src/Select/Select.tsx
```diff
@@ -1,10 +1,10 @@
...
 import { Label, LabelText, inputFocus } from "../utils/mixins"
 import SelectOption from "./Select.Option"
 import SelectFilter from "./Select.Filter"
-import { readableTextColor, spin, fadeIn, resetTransform } from "@operational/utils"
+import { readableTextColor, fadeIn, resetTransform } from "@operational/utils"
 import { OperationalStyleConstants, Theme, expandColor } from "@operational/theme"
```
ImogenF: Same comment as above re: `expandColor`

packages/components/src/ContiamoLogo/ContiamoLogo.tsx
TejasQ: The interface `Props` _really helps_ with things like [intellisense](https://code.visualstudio.com/docs/editor/intellisense). I gave it a more ambiguous name in order to help developers using _this component's_ props differentiate between the other hundred `Props` exports we have throughout the library.

packages/components/src/Select/Select.tsx
TejasQ: Actually, it's a recommended best practice to separate package imports from local imports with a blank line. 

For me, it helps know where importing libraries/frameworks end and where importing _our own stuff_ starts.

ImogenF: Hmmmm ok. Good to know!

TejasQ:
@micha-f care to recheck the design?

micha-f:
I believe there is still a little bit more space below the C than above? Might be just a pixel or two.

packages/components/src/HeaderBar/HeaderBar.tsx
```diff
@@ -0,0 +1,58 @@
...
+   * leftmost thing.
+   */
+  start: React.ReactNode
+
+  /** The center of the HeaderBar. */
+  center?: React.ReactNode
+
+  /** The "end" element of the HeaderBar. */
+  end?: React.ReactNode
+}
```
peterszerzo: I don't understand the i18n argument in naming these props - they will never show up in the output html, and you can translate actual content if needed (and in terms of code readability, `left` is pretty clear regardless of the direction in which one language or another writes). I feel like we're some sacrificing important clarity in where these components go. `center` is not actually centered, it is left-aligned in the main section of the bar.

I'd vouch for `start -> corner`, `center -> left` and `end -> right`.

fabien0102: If you take one of the last css property: `flex`, they also use `flex-end` and `flex-start` as value to define the classic and old school `right` and `left`. For me it's just become a good practice to not use `left` and `right` as `start` and `end` are more semantic and future proof :wink:

TejasQ: But `end` is _also_ a corner, no? Also, `start` is more of a `side` so I'm not sure the `corner` semantic would work...

fabien0102: or `logo`, `start`, `end`?

peterszerzo: The way I imagine it is that there is a `corner` section of the component and a main section, technically, you could do `<HeaderBar corner={</>} main={{ start: </>, end: </> }} />`, and flattening that would lead to the last thing I suggested.

TejasQ: How's about:

```jsx
<HeaderBar
  logo={<Logo />}
  main={<ProjectSelector />} {/* this is nice because _anything_ can go in main and it's not the "center" */}
  end={<UserMenu />}
/>
```

If we agree, I'll commit this and release things asap.

peterszerzo: I love it!

packages/components/src/ContiamoLogo/ContiamoLogo.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+const LogoType = styled("svg")(({ theme, color }: LogoProps & WithTheme) => ({
+  fill: expandColor(theme, color) || theme.color.white,
+}))
+
+const ContiamoLogo: React.SFC<LogoProps> = ({ size, color }) => (
+  <LogoContainer>
+    <Icon left name="Contiamo" size={size} color={color} />
+    <LogoType color={color} width={size * 3.5} viewBox="650 255 1800 215" fill="currentColor">
+      <g transform="matrix(1.3333333,0,0,-1.3333333,0,737.64133)">
```
peterszerzo: Add a `title` tag here for accessibility, since the contiamo logo isn't spelled out in words.

packages/components/src/Avatar/Avatar.tsx
```diff
@@ -42,70 +35,53 @@ const Container = styled("div")({
...
       ? theme.deprecated.colors.visualizationPalette[
           colorAssignment % theme.deprecated.colors.visualizationPalette.length
         ]
       : null
     const backgroundColor = assignedBackgroundColor || fixedBackgroundColor
-    const textColor = readableTextColor(backgroundColor, [theme.deprecated.colors.text, "white"])
+    const textColor = readableTextColor(backgroundColor, [theme.color.text.default, "white"])
     return {
-      fontSize: 9,
+      fontSize: size === "medium" ? 13 : 9,
```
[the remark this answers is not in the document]
TejasQ: I disagree. I think the spread is harder to notice at a glance, whereas this style would read more along the rhythm of "fontSize -> is it medium? 13. Otherwise, 9."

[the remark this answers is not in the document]
peterszerzo: The spread is at the very beginning of the line, and makes it very clear that `hey, here is a consolidated set of styles for medium, and here is a different set otherwise`. We use the spread + single-branching pattern across the codebase (`Breakdown`, `Input`, `SidenavHeader`), and if the readability is coming in close to a tie in this particular case, it becomes a lot trickier to read once you have fields that are set in one case and not the other like so:
```
{
  border: propField ? "1px solid" : undefined,
  margin: propField ? undefined : 10px
}
```
as opposed to
`...propField ? { border: "1px solid" } : { margin: 10 }`
I'd vouch for consistency here.

PR #513, opened 2018-06-20
Code component

ImogenF:
Code component with pale blue background for nicely displaying code snippets. Uses a code font and applies syntax specific highlighting (currently broken in website, but working in docs)


packages/components/src/Code/Code.tsx
```diff
@@ -0,0 +1,46 @@
...
+import { OperationalStyleConstants } from "@operational/theme"
+import { Css } from "../types"
+import * as hljs from "highlight.js"
+
+export interface Props {
+  /** Id */
+  id?: string
+
+  /** `css` prop as expected in a glamorous component */
+  css?: Css
```
fabien0102: Don't need this anymore ;)

ImogenF:
@micha-f this is still a WIP, but would appreciate any feedback you might have about the design: https://deploy-preview-513--operational-ui.netlify.com/docs/#code. 

I'm not sure how to change the colours used in the highlighting to match the screenshot you just added to trello but I'll look into that if you think it's important.

ImogenF:
[Code element](https://trello.com/c/B0MTyxTP/200-code-element)

TejasQ:
This is really strange. Updating the code in the playground breaks the component:

image

ImogenF:
@TejasQ I'm aware of that bug - it's not just when you update the code in the playground either. Trying to figure out what's going on / find a way to fix it.

micha-f:
Looks great!

packages/components/package.json
```diff
@@ -37,15 +37,18 @@
     "emotion-theming": "^9.2.3",
     "glamor": "^2.20.40",
     "glamorous": "^4.6.0",
+    "highlight.js": "^9.12.0",
```
peterszerzo: Do we need this dependency? `react-highlight` should use depend on it internally.

packages/components/package.json
```diff
@@ -37,15 +37,18 @@
...
     "react": "^16.2.0",
     "react-dom": "^16.2.0",
     "react-emotion": "^9.2.3",
     "react-feather": "^1.1.0",
+    "react-highlight": "^0.12.0",
     "react-icons": "^2.2.5"
   },
   "devDependencies": {
     "@babel/core": "^7.0.0-beta.51",
+    "@types/highlight.js": "^9.12.3",
```
peterszerzo: See comment above

packages/components/src/__tests__/Code.test.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { Code as ThemelessCode } from "../index"
+import wrapDefaultTheme from "../utils/wrap-default-theme"
+
+const Code = wrapDefaultTheme(ThemelessCode)
+
+describe("Code", () => {
+  it("Should render", () => {
+    expect(
+      render(
+        <Code syntax="json">
```
peterszerzo: Update to the `code` prop to make sure component renders with that also.

packages/components/src/Code/Code.tsx
```diff
@@ -2,18 +2,96 @@ import * as React from "react"
 import styled from "react-emotion"
 import { OperationalStyleConstants } from "@operational/theme"
 import Highlight from "react-highlight"
+import { injectGlobal } from "emotion"
+
+injectGlobal`
```
TejasQ: Can we move this to the bottom of the file, please?

ImogenF:
@TejasQ Is that really nicer than how it was originally? I personally find it easier to understand if `export default Code` comes directly after the definition of `Code`

packages/components/src/Code/Code.tsx
```diff
@@ -0,0 +1,43 @@
...
+  id?: string
+  /** Language for syntax highlighting */
+  syntax?: string
+  /** Code string to highlight */
+  code?: string
+  children?: any
+}
+
+const Container = styled("div")(
+  ({ theme }: { theme?: OperationalStyleConstants }): {} => {
```
TejasQ: We don't need to annotate the return type here.

packages/components/src/Code/Code.tsx
```diff
@@ -0,0 +1,43 @@
...
+      backgroundColor: "rgba(20, 153, 206, 0.05)",
+      border: "1px solid rgba(20, 153, 206, 0.1)",
+      borderRadius: theme.borderRadius,
+      padding: `0 ${theme.space.small}px`,
+    }
+  },
+)
+
+const StyledHighlight = styled(Highlight)(
+  (): {} => {
```
TejasQ: We don't need to annotate a return type here.

packages/components/src/Code/Code.tsx
```diff
@@ -0,0 +1,43 @@
...
+    return {
+      background: "transparent",
+    }
+  },
+)
+
+const Code = (props: Props) => (
```
TejasQ: Maybe we can destructure props here for conciseness.

packages/components/src/__tests__/Code.test.tsx
```diff
@@ -0,0 +1,30 @@
+import * as React from "react"
+import { render, mount } from "enzyme"
+import { Code as ThemelessCode } from "../index"
+import wrapDefaultTheme from "../utils/wrap-default-theme"
```
TejasQ: This can be `import { withTheme } from "emotion-theming"`

peterszerzo: If you use a built-in method for injecting context into the component, you will also need to import our theme from a different package and make sure it's available in context, making tests more verbose. That was the reason `wrapDefaultTheme` was created.

packages/components/src/Code/Code.tsx
peterszerzo: I would like to discourage destructuring props - for components with lots of props, it is disorienting to keep track of so many loose variables running around, and there is nothing distinguishing them visually from other local variables that don't come from props.

PR #514, opened 2018-06-20
Enhance sidenav

peterszerzo:
Adds missing features to the side nav:
* [x] truncated summary items in `SidenavHeader`
* [x] `SidenavHeader` icon
* [x] condensed `SidenavHeader`
* [x] active state fully controlled with `active` and `onToggle` props
* [x] polish + adjust spacing to exactly match designs

screen shot 2018-06-20 at 4 03 21 pm

### DIFF
--- a/packages/components/src/SidenavHeader/SidenavHeader.tsx
+++ b/packages/components/src/SidenavHeader/SidenavHeader.tsx
@@ -22,6 +22,8 @@ export interface Props {
   condensed?: boolean
   /** Active state - renders colored strip on the left */
   active?: boolean
+  /** Callback called when the active state changes */
+  onToggle?: (newActiveState: boolean) => void
   /**
    * Expanded state
    *
@@ -35,9 +37,7 @@ export interface Props {
   children?: React.ReactNode
 }
 
-export interface State {
-  isOpen: boolean
-}
+export interface State {}
 
 const containerStyles = ({
   theme,
@@ -183,18 +183,15 @@ const truncate = (maxLength: number) => (text: string) => {
 }
 
 export class SidenavHeader extends React.Component<Props, State> {
-  state = {
-    isOpen: false,
-  }
+  state = {}
 
   render() {
     const hasChildren = Boolean(this.props.children)
     const hasChildLinks = React.Children.toArray(this.props.children).some(child =>
       Boolean((child as { props: SidenavItemProps }).props.to),
     )
-    const isActive =
-      Boolean(this.state.isOpen || (this.props.to && window.location.pathname.match(`^${this.props.to}`))) &&
-      hasChildren
+
+    const isActive = Boolean(this.props.active)
 
     // Actual `to` prop should invalidate if the element has sublinks and is active
     const to = isActive && hasChildLinks ? undefined : this.props.to
@@ -212,9 +209,7 @@ export class SidenavHeader extends React.Component<Props, State> {
               className={[this.props.className, "op_sidenavheader"].filter(cls => Boolean(cls)).join(" ")}
               onClick={(ev: React.SyntheticEvent<Node>) => {
                 this.props.onClick && this.props.onClick()
-                this.setState(prevState => ({
-                  isOpen: !prevState.isOpen,
-                }))
+                this.props.onToggle && this.props.onToggle(!this.props.active)
 
                 if (!isModifiedEvent(ev) && ctx.pushState && this.props.to) {
                   ev.preventDefault()
@@ -242,12 +237,10 @@ export class SidenavHeader extends React.Component<Props, State> {
                   onClick={(ev: React.SyntheticEvent<Node>) => {
                     // Prevent clicks on parent in order to avoid conflicting behavior
                     ev.stopPropagation()
-                    this.setState(prevState => ({
-                      isOpen: !prevState.isOpen,
-                    }))
+                    this.props.onToggle && this.props.onToggle(!this.props.active)
                   }}
                 >
-                  <Icon name={this.state.isOpen ? "ChevronUp" : "ChevronDown"} />
+                  <Icon name={this.props.active ? "ChevronUp" : "ChevronDown"} />
                 </CloseButton>
               )}
               {isActive && <ItemsContainer>{this.props.children}</ItemsContainer>}
--- a/packages/components/src/SidenavHeader/SidenavHeader.tsx
+++ b/packages/components/src/SidenavHeader/SidenavHeader.tsx
@@ -58,12 +58,6 @@ const containerStyles = ({
     borderLeft: "4px solid",
     borderLeftColor: isActive ? stripColor : "transparent",
     borderBottomColor: theme.deprecated.colors.separator,
-
-    /** @todo Add to theme once colors are updated across codebase */
-    backgroundColor: isActive ? "#F8F8F8" : "transparent",
-    ":hover": {
-      backgroundColor: "#F8F8F8",
-    },
   }
 }
 
@@ -116,6 +110,7 @@ const ItemsContainer = styled("div")(
       deprecated: Theme
     }
   }): {} => ({
+    animation: `${fadeIn} .15s forwards ease`,
     position: "relative",
     top: -16,
     marginTop: -10,
--- a/packages/components/src/SidenavHeader/SidenavHeader.tsx
+++ b/packages/components/src/SidenavHeader/SidenavHeader.tsx
@@ -37,8 +37,6 @@ export interface Props {
   children?: React.ReactNode
 }
 
-export interface State {}
-
 const containerStyles = ({
   theme,
   color,
@@ -102,10 +100,11 @@ const LabelText = styled("p")`
   letter-spacing: 0.25;
   text-transform: uppercase;
   white-space: nowrap;
+  user-select: none;
   margin: 0;
   ${({ isActive, theme }: { isActive: boolean; theme?: OperationalStyleConstants }) => `
-    color: ${theme.color.text.dark},
-    font-size: ${theme.font.size.body}px,
+    color: ${theme.color.text.dark};
+    font-size: ${theme.font.size.body}px;
   `};
 `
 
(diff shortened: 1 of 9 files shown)


packages/components/src/SidenavHeader/SidenavHeader.tsx
```diff
@@ -87,32 +76,46 @@ const ContainerLink = styled("a")(containerStyles)
...
-    letterSpacing: 0.25,
-    fontSize: 14,
-    textTransform: "uppercase",
-    whiteSpace: "nowrap",
   }),
 )
 
+const LabelText = styled("p")`
+  position: relative;
+  color: #333333;
```
TejasQ: `theme.color.text.dark`

packages/components/src/SidenavHeader/SidenavHeader.tsx
```diff
@@ -87,32 +76,46 @@ const ContainerLink = styled("a")(containerStyles)
...
+  font-weight: 500;
+  letter-spacing: 0.25;
+  font-size: 14;
```
TejasQ: `theme.font.size.body`

packages/components/src/SidenavHeader/SidenavHeader.tsx
```diff
@@ -87,32 +76,46 @@ const ContainerLink = styled("a")(containerStyles)
...
+  text-transform: uppercase;
+  white-space: nowrap;
+  margin: 0;
+  ${({ isActive }: { isActive: boolean }) =>
+    isActive
+      ? `
+    top: -5px;
```
TejasQ: Why `margin-top` in one case and `top` in another?

packages/components/src/SidenavHeader/SidenavHeader.tsx
```diff
@@ -122,7 +125,7 @@ const ItemsContainer = styled("div")(
     }
   }): {} => ({
     position: "relative",
-    top: -theme.deprecated.spacing,
+    top: -theme.space.content - 6,
```
TejasQ: If these values (and values on lines 145-148) are specific to this component, let's skip manipulating `theme.space` for them.

packages/components/src/SidenavHeader/SidenavHeader.tsx
```diff
@@ -154,20 +157,53 @@ const CloseButton = styled("div")(
   }),
 )
 
+const IconContainer = styled("p")`
+  display: inline-block;
+  vertical-align: middle;
+  margin: 0 0 0 8px;
```
TejasQ: We can probably use `theme.space.small` here.

packages/components/src/SidenavHeader/SidenavHeader.tsx
```diff
@@ -154,20 +157,53 @@ const CloseButton = styled("div")(
...
+    position: relative;
+    top: -2px;
+  }
+`
+
+const Summary = styled("div")`
+  display: block;
+  font-weight: normal;
+  text-transform: none;
+  margin-top: 4px;
```
TejasQ: And `theme.space.base` here.

packages/components/src/SidenavHeader/SidenavHeader.tsx
```diff
@@ -154,20 +157,53 @@ const CloseButton = styled("div")(
...
+  return text.slice(0, maxLength) + "..."
+}
+
 export class SidenavHeader extends React.Component<Props, State> {
   state = {
     isOpen: false,
   }
 
   render() {
+    const hasChildren = React.Children.count(this.props.children) > 0
```
TejasQ: or `hasChildren = Boolean(props.children)`?

packages/components/src/SidenavHeader/SidenavHeader.tsx
```diff
@@ -154,20 +157,53 @@ const CloseButton = styled("div")(
...
     const hasChildLinks = React.Children.toArray(this.props.children).some(child => (child as any).props.to)
```
TejasQ: This is sick. Can we improve the typing of the child somehow?

packages/components/src/SidenavHeader/SidenavHeader.tsx
peterszerzo: Found a better solution for this one, in which the gray summary list is not removed but hidden, making hacks that maintain the position of the label unnecessary.

packages/components/src/SidenavHeader/SidenavHeader.tsx
peterszerzo: This is a fine-tuning value (as in that go-the-extra-mile step that nudges things to the right place after the layout was put in place). These should not be encoded as variables. I can mention this in a comment.

packages/components/src/SidenavHeader/SidenavHeader.tsx
peterszerzo: A bit I'd say - I can cast it as an object that has a props field typed as `SidenavItem`'s props. `ReactChild` does not take generics, so it is not possible to know whether it even has props (as raw strings are also valid).

packages/components/src/SidenavHeader/SidenavHeader.tsx
```diff
@@ -5,51 +5,41 @@ import { fadeIn } from "@operational/utils"
...
   /** Close handler (via chevron button on the top right) */
-
   onClose?: () => void
   children?: React.ReactNode
 }
 
-export interface State {
-  isOpen: boolean
-}
+export interface State {}
```
TejasQ: Empty state?

packages/components/src/SidenavHeader/SidenavHeader.tsx
```diff
@@ -139,10 +138,10 @@ const CloseButton = styled("div")(
     display: "none",
     alignItems: "center",
     justifyContent: "center",
-    width: theme.deprecated.spacing * 1.5,
-    height: theme.deprecated.spacing * 1.5,
-    top: theme.deprecated.spacing * 1.5,
-    right: theme.deprecated.spacing,
+    width: theme.space.content * 1.5,
```
TejasQ: We can just hardcode numbers here I think since they're specific to this close button.

packages/components/src/SidenavHeader/SidenavHeader.tsx
```diff
@@ -139,10 +138,10 @@ const CloseButton = styled("div")(
...
+    height: theme.space.content * 1.5,
+    top: theme.space.content * 1,
```
TejasQ: `* 1`?

packages/components/src/SidenavHeader/SidenavHeader.tsx
TejasQ: Sure.

packages/components/src/SidenavHeader/SidenavHeader.tsx
```diff
@@ -154,20 +153,52 @@ const CloseButton = styled("div")(
...
+    const hasChildLinks = React.Children.toArray(this.props.children).some(child =>
+      Boolean((child as { props: SidenavItemProps }).props.to),
     )
 
+    const isActive = Boolean(this.props.active)
+
     // Actual `to` prop should invalidate if the element has sublinks and is active
     const to = isActive && hasChildLinks ? undefined : this.props.to
     const ContainerComponent = to ? ContainerLink : Container
+    const childLabels: string[] = React.Children.map(this.props.children, child => (child as any).props.label) || []
```
TejasQ: Can we better type this `child`?

packages/components/src/SidenavHeader/SidenavHeader.tsx
```diff
@@ -181,9 +212,7 @@ export class SidenavHeader extends React.Component<Props, State> {
               className={[this.props.className, "op_sidenavheader"].filter(cls => Boolean(cls)).join(" ")}
```
TejasQ: Can we not rely on classnames please?

peterszerzo: Will be addressed in a different PR, see #517

TejasQ:
Can we update the `Layout` component in the docs to use the new Sidenav as well?

TejasQ:
Can we make [the demo](https://deploy-preview-514--operational-ui.netlify.com/docs/#!/Layout) stateful so we can try expand/collapse functionality?

Also, looks like we need a `cursor: pointer` on the items.

TejasQ:
Also, 

#### This
image

#### And this
image

seem sketchy. @micha-f care to comment on the design?

TejasQ:
[Sidebar Component](https://trello.com/c/8ZcQ30Pr/172-sidebar-component)

packages/components/src/SidenavItem/SidenavItem.tsx
```diff
@@ -1,8 +1,6 @@
 import * as React from "react"
-import styled from "react-emotion"
-import { OperationalStyleConstants, Theme } from "@operational/theme"
-import { lighten } from "@operational/utils"
-import { WithTheme, Css, CssStatic } from "../types"
+import styled, { Interpolation } from "react-emotion"
```
fabien0102: Interpolation?

packages/components/src/SidenavHeader/SidenavHeader.tsx
```diff
@@ -68,153 +50,169 @@ const containerStyles = ({
...
+  return (
+    <ContextConsumer>
+      {(ctx: Context) => {
+        return (
+          <ContainerComponent
+            id={props.id}
+            href={to}
+            color={props.color}
+            isActive={isActive}
+            className={[props.className, "op_sidenavheader"].filter(cls => Boolean(cls)).join(" ")}
```
fabien0102: Why do we need this magic here?

fabien0102:
I have 2 more questions about this PR (even if it's close)

* How can I set an `SidenavItem` active?
* Is it make sense to deal with the state our self here?

peterszerzo:
The fix for the `Sidenavitem` active state is coming up in a different branch.

I don't think it's too much bother to have active state this controlled from the outside, especially because there will be other custom needs such as preventing a `SidenavHeader` close if the element in question is active based on the current URL. At that point, local state will just be fighting outside state, which I would argue will set us up for bugs.

PR #515, opened 2018-06-20
Fix UI regressions for inclusion in our products

TejasQ:
- Updated Button Icon positioning
- Updated Cards with text overflows
- Updated CardHeader links with icons
- Updated CardItems to include children
- Updated HeaderBar to have 100% width instead of 100vw
- Updated Icon to be a styled component
- Updated Layout component to more closely resemble our use cases
- Updated Page component to have sticky headers

### DIFF
--- a/packages/components/src/Button/Button.tsx
+++ b/packages/components/src/Button/Button.tsx
@@ -58,7 +58,8 @@ const containerStyles: Interpolation<Props> = ({
     label: "button",
     fontSize: theme.font.size.small,
     fontFamily: theme.font.family.main,
-    display: "inline-block",
+    display: "inline-flex",
+    alignItems: "center",
     padding: `0 ${condensed ? theme.space.small : theme.space.content}px`,
     borderRadius: theme.borderRadius,
     border: 0,
@@ -87,11 +88,9 @@ const Container = styled("button")(containerStyles)
 const ContainerLink = styled("a")(containerStyles)
 
 const IconContainer = styled("div")(({ theme }: Css) => ({
+  display: "flex",
+  alignItems: "center",
   marginLeft: theme.space.small,
-  float: "right",
-  "& svg": {
-    verticalAlign: "text-bottom",
-  },
 }))
 
 const ButtonSpinner = styled(Spinner)(
--- a/packages/components/src/Button/Button.tsx
+++ b/packages/components/src/Button/Button.tsx
@@ -90,7 +90,6 @@ const ContainerLink = styled("a")(containerStyles)
 const IconContainer = styled("div")(({ theme }: Css) => ({
   display: "flex",
   alignItems: "center",
-  marginLeft: theme.space.small,
 }))
 
 const ButtonSpinner = styled(Spinner)(
@@ -136,7 +135,7 @@ const Button = (props: Props) => {
           {props.children}
           {props.icon && (
             <IconContainer condensed={props.condensed}>
-              {typeof props.icon === "string" ? <Icon name={props.icon as IconName} size={16} /> : props.icon}
+              {typeof props.icon === "string" ? <Icon right name={props.icon as IconName} size={16} /> : props.icon}
             </IconContainer>
           )}
           {props.loading && <ButtonSpinner containerColor={props.color} />}
--- a/packages/components/src/CardItem/CardItem.tsx
+++ b/packages/components/src/CardItem/CardItem.tsx
@@ -24,10 +24,10 @@ const CardItemBody = styled("div")(({ theme }: { theme?: OperationalStyleConstan
   marginBottom: theme.space.content,
 }))
 
-const CardItem: React.SFC<Props> = ({ title, value }) => (
+const CardItem: React.SFC<Props> = ({ title, value, children }) => (
   <div>
     <CardItemTitle>{title}</CardItemTitle>
-    <CardItemBody>{value}</CardItemBody>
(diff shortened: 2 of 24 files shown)


packages/components/src/Button/Button.tsx
```diff
@@ -89,11 +90,9 @@ const Container = styled("button")(containerStyles)
 const ContainerLink = styled("a")(containerStyles)
 
 const IconContainer = styled("div")(({ theme }: Css) => ({
+  display: "flex",
+  alignItems: "center",
   marginLeft: theme.space.small,
```
peterszerzo: Might it make sense to use `<Icon left` down below instead of adding a margin here? That way the value of the margin doesn't need to be duplicated and guarantees consistent icon spacing across the UI.

I know this doesn't work for custom icons specified by the user (the `props.icon` case), but I think we should aim for removing custom icons altogether and just do strings to guarantee a bit more consistency.

packages/components/src/CardItem/CardItem.tsx
```diff
@@ -24,10 +24,10 @@ const CardItemBody = styled("div")(({ theme }: { theme?: OperationalStyleConstan
   marginBottom: theme.space.content,
 }))
 
-const CardItem: React.SFC<Props> = ({ title, value }) => (
+const CardItem: React.SFC<Props> = ({ title, value, children }) => (
   <div>
     <CardItemTitle>{title}</CardItemTitle>
-    <CardItemBody>{value}</CardItemBody>
+    <CardItemBody>{children || value}</CardItemBody>
```
peterszerzo: Why both `children` and `value`?

packages/components/src/Layout/Layout.tsx
```diff
@@ -7,79 +7,54 @@ import { Props as SidenavProps } from "../Sidenav/Sidenav"
...
 const Layout = (props: Props) => {
   return (
-    <Container css={props.css} className={props.className}>
-      {props.loading ? <Progress /> : <CssPlaceholder />}
-      {props.sidenav}
-      <Main>{props.main}</Main>
+    <Container>
+      {props.header}
+      <Content>
+        {/* {props.loading ? <Progress /> : <CssPlaceholder />} */}
```
peterszerzo: Is the intent here to remove the loading functionality altogether or move it to some other component? If so, what is the rationale? I wonder whether we could instead uncomment this line and pull it one node up next to `{props.header}`?

packages/components/src/Page/Page.tsx
```diff
@@ -79,6 +88,7 @@ const Grid = styled("div")(
       maxWidth: props.fill ? "none" : 1150,
```
peterszerzo: Are we sure that the `Page` component is the right place to restrict max width? Relying on the `fill` prop to do this seems to muddle up concerns, and therefore hard to track - can we set it on the page's content instead?

packages/components/src/Button/Button.tsx
TejasQ: `<Icon right`*, you mean. Will do.

packages/components/src/CardItem/CardItem.tsx
TejasQ: It's just syntax sugar for people who prefer more concise JSX.

packages/components/src/Layout/Layout.tsx
TejasQ: Oooo, this was an accident. Good catch!

packages/components/src/Page/Page.tsx
TejasQ: I'll try it.

fabien0102: it make sense, because we have this kind of structure:

```tsx
const Tab = n => () => (
  <Page>
    <Card title={`${n} Tab`} />
  </Page>
)
;<Page
  title="Bundle detail"
  tabs={[
    { name: "overview", component: Tab("overview") },
    { name: "jobs", component: Tab("jobs") },
    { name: "functions", component: Tab("functions") },
  ]}
/>
```

So Page into Page (and sometime fullwidth, sometime not, depending of the context)

peterszerzo:
@ImogenF thanks for reporting, I'll fix up the website early next week.

packages/components/src/Page/Page.tsx
TejasQ: Turns out, the max width is set on the `Grid`, which _is_ the page's content.

packages/components/src/PageArea/PageArea.tsx
```diff
@@ -3,7 +3,7 @@ import styled from "react-emotion"
 
 export interface PageAreaProps {
   /** Name of the area */
-  name?: string
+  name?: "main" | "side"
```
peterszerzo: This is awesome!

packages/components/src/Layout/README.md
```diff
@@ -1,43 +1,103 @@
 This component lays out an opinionated application frame with side navigation, a main section and some optional conveniences. It sits at the top level of the page.
 
-## Usage
+### Basic Layout
 
 ```jsx
-const  Breadcrumb = require("../Breadcrumb/Breadcrumb").default
-
 const sidebar = (
```
peterszerzo: This will be fixed in the sidenav branch (make sure to keep `Sidenav` examples as opposed to `Sidebar` ones when resolving merge conflicts)

PR #520, opened 2018-06-22
Spring cleaning

peterszerzo:
* [x] sidebar is removed - sidebar components have never really been used, they are too close in name to `Sidenav` components. I suggest removing them altogether - if we need them back, we can look at the implementation out of version control, but maybe start from scratch with real business needs in mind.
* [x] visual tests move to `emotion`.
* [x] visual tests can move to `Sidenav` components.
* [x] visualizations docs move to their respective folder.
* [x] remove website, instead, add splash screen to `styleguidist`.
* [x] fix minor breaks in components (`Breadcrumb`, `ButtonGroup`)


packages/visual-tests/src/App.tsx
```diff
@@ -98,96 +98,89 @@ class App extends React.Component<{}, State> {
...
+                      }}
+                      label={test.title}
+                    />
+                  ))}
+                </SidenavHeader>
+              ))}
+            </Sidenav>
+          }
+          main={
+            <div style={{ height: "100%" }}>
```
TejasQ: Is this necessary? Should Operational handle this out of the box?

packages/visual-tests/src/MarathonRenderer.tsx
```diff
@@ -30,35 +30,31 @@ const Item = glamorous.li({
...
+  display: "inline-block",
+  marginLeft: 8,
+  "&::before": {
+    content: " → ",
+  },
+}))
 
 const MarathonRendererComponent = ({ results, ref }: MarathonRenderer) => (
-  <Div>
+  <div>
```
TejasQ: Können wir es als Fragment schreiben?

packages/visual-tests/src/App.tsx
peterszerzo: It used to, but now I'm having a hard time getting it to work. Care to lend me a hand?

packages/visual-tests/src/MarathonRenderer.tsx
peterszerzo: Da hast du ja wohl Recht. Danke für den Vorschlag :)

packages/components/styleguide/Logo.tsx
```diff
@@ -0,0 +1,148 @@
+import * as React from "react"
```
TejasQ: Should this be called Logo?

packages/components/styleguide/OperationalLogo.tsx
```diff
@@ -0,0 +1,30 @@
+import * as React from "react"
```
TejasQ: Can we use `<Icon name="Operational" />` instead?

peterszerzo: This one is a slightly different format to accommodate the rotation.

packages/components/styleguide/Logo.tsx
peterszerzo: You're right, we should call this LogoAndBanner for clarity.

Granted, the current way of adding the banner screen onto the website is a bit strange (going through a portal that is controlled through the logo component on the upper left corner), but everything else I've tried was hackier and less maintainable (copy-paste the implementation of styleguidist's page wrapper component and fix the typings manually).

packages/components/styleguide/OperationalLogo.tsx
peterszerzo: Not if we want the rotation.

packages/visual-tests/src/App.tsx
fabien0102: Any update about this?

packages/visual-tests/src/MarathonRenderer.tsx
fabien0102: No fragment?

packages/website/src/favicons/browserconfig.xml
```diff
@@ -1,2 +0,0 @@
-<?xml version="1.0" encoding="utf-8"?>
-<browserconfig><msapplication><tile><square70x70logo src="/ms-icon-70x70.png"/><square150x150logo src="/ms-icon-150x150.png"/><square310x310logo src="/ms-icon-310x310.png"/><TileColor>#ffffff</TileColor></tile></msapplication></browserconfig>
```
fabien0102: It can be nice to add a favicon to our styleguist ;)

PR #524, opened 2018-06-22
Update Grid logic for Page

### LINKED ISSUE #521: `Page` does not handle single column grid elements well
## Problem

image

Operational UI, when laying out single-column Page contents, the rows scale far too big because the Grid assumes each immediate child is a row with an `auto` height.

## Solution

Wrap children in a `<div>` to signal to CSS Grid that it's ONE row with stacked children.

TejasQ:
This PR fixes #521 and adds a convenience script `yarn test:components:update` to update snapshots root-level.

### DIFF
--- a/packages/components/src/Page/Page.tsx
+++ b/packages/components/src/Page/Page.tsx
@@ -85,6 +85,11 @@ const initialState = {
 }
 
 class Page extends React.Component<Props, Readonly<typeof initialState>> {
+  static defaultProps = {
+    areas: "main",
+    fill: false,
+  }
+
   constructor(props: Props) {
     super(props)
     if (props.activeTabName && props.tabs) {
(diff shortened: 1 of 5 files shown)


packages/components/src/Page/Page.tsx
```diff
@@ -85,6 +85,11 @@ const initialState = {
 }
 
 class Page extends React.Component<Props, Readonly<typeof initialState>> {
+  static defaultProps = {
+    areas: "main",
```
peterszerzo: This prop should be an array, or better yet, extracted automatically from the props of the children.

TejasQ: Why should it be an array? Are you thinking `["main", "side"]`? That doesn't quite make sense to me. It's just a string that describes _"how should I lay out my children?"_. I'm not sure I understand your angle here.

TejasQ: It's not as easy as examining the children. You could have the following structure:

```jsx
<Page>
  <Child1>Hello!</Child1>
  <Child2>Hola!</Child2>
</Page>
```

What would you expect? Them to stack like HTML blocks? I would. But what if we want to signal to operational that we want them to be a main area with a smaller sidebar next to it?

```diff
- <Page>
+ <Page areas="main side">
  <Child1>Hello!</Child1>
  <Child2>Hola!</Child2>
</Page>
```

...done. It's magic

What if you wanted `Child1` as the sidebar and `Child2` as the main content area?

```diff
- <Page areas="main side">
+ <Page areas="side main">
  <Child1>Hello!</Child1>
  <Child2>Hola!</Child2>
</Page>
```

omg stop the presses we have a magical layout to rule them all.

Does that make sense as to why it requires a little more than examining the children?

Issue #525, opened 2018-06-22
Create a Sha component


TejasQ:
[Inspired by this comment](https://github.com/contiamo/labs-ui/pull/76#discussion_r197387826), we're going to be working with a lot of SHAs in the Labs project, so let's create a convenience component for it!

PR #528, opened 2018-06-25
Add Sha component

fabien0102:
### Issue

#525 

Very simple component, I assume for now that the text style is mainly given by the parent.


packages/components/src/Sha/Sha.tsx
```diff
@@ -0,0 +1,16 @@
+import * as React from "react"
+import styled from "react-emotion"
+import { OperationalStyleConstants } from "@operational/theme"
+
+interface ShaProps {
+  short?: boolean
+  children: string
```
peterszerzo: I think we should be opinionated with font-sizes here - if we leave the responsibility to the outer context to determine this, the results are both more difficult to track and open us up to inconsistencies where this is sized 11px in one part of the UI and 13px in the other.

Could the `size: "small" | "medium"` prop work here? Seems like a solution that we can extend if we really need a third kind of size.

fabien0102: And I don't think so, I don't want dealing with the size here, just want a conveniant way to display a Sha in a specific context, so just override what is really necessary. Putting a `size` props here is useless for me.

peterszerzo: It's not exactly fair to call it useless - having a few number of consistent font sizes is a clear win for design, and an API that enforces that is desirable, especially after we write many thousands of lines of code that'll be hard to sift through.

I would also pin down font weight while we're at it - deviations won't look untrustworthy in my opinion.

TejasQ: SHAs are typically shortened because they're looooooooooooooooong. I think the purpose of this component _should be_ to improve the coding experience of a developer by removing the overhead of _"how short is a short SHA?"_ 🤔 each time he wants a short SHA in a UI. We can do this by specifying the `short` prop as a simple convenience boolean value, while abstracting away the mental overhead.

Can we improve the developer experience by setting font size/weight? I don't think so. I don't think so because there are different font sizes according to the design: some texts are `13px`, some are `14px`. 

As a developer, I'd just want my SHA to be short, and then look like whatever's around it (See GitHub). 

I'd say let's merge this for now, and if the event arises where we have ugly, inconsistent SHAs everywhere, we pin down styles going forward.

PR #529, opened 2018-06-25
Deal with Page and router

fabien0102:
As we have some sub-pages in our product, we need a way to inject router logics into `Page` component.

For this, I have the following proposition:

```jsx
<Page
  title="With a router"
  onTabChange={this.onChange}
  activeTabName={/* router -> tabs matching */}
  tabs={["overview", "jobs", "functions"]}
>
  <Switch>
    <Route path="overview" render={() => <Card title="overview" /> } />
    <Route path="jobs" render={() => <Card title="jobs" /> } />
    <Route path="functions" render={() => <Card title="functions" /> } />
    <Route path="functions/:id" render={() => <Card title="functions/:id" /> } />
  </Switch>
</Page>
```

So it's not link internally to the router, but we can play with the `Page` and `react-router` together to deal with the most advanced routing patterns.

### DIFF
--- a/packages/components/src/Page/Page.tsx
+++ b/packages/components/src/Page/Page.tsx
@@ -5,6 +5,10 @@ import { Title } from ".."
 import PageArea from "../PageArea/PageArea"
 import PageContent from "../PageContent/PageContent"
 
+const isStringArray = (value: any): value is string[] => {
+  return value && typeof value[0] === "string"
+}
+
 export interface Props {
   /** Page title */
   title?: string
@@ -20,7 +24,7 @@ export interface Props {
    * List of tabs
    * This will disable any children to render `tabs[i].component` instead
    */
-  tabs?: { name: string; component: React.ComponentType }[]
+  tabs?: { name: string; component: React.ComponentType }[] | string[]
   /**
    * Active tab name
    *
@@ -93,7 +97,9 @@ class Page extends React.Component<Props, Readonly<typeof initialState>> {
   constructor(props: Props) {
     super(props)
     if (props.activeTabName && props.tabs) {
-      const index = props.tabs.findIndex(({ name }) => name.toLowerCase() === props.activeTabName.toLowerCase())
+      const index = isStringArray(props.tabs)
+        ? props.tabs.findIndex(name => name.toLowerCase() === props.activeTabName.toLowerCase())
+        : props.tabs.findIndex(({ name }) => name.toLowerCase() === props.activeTabName.toLowerCase())
       this.state = {
         ...initialState,
         activeTab: index === -1 ? 0 : index,
@@ -105,14 +111,18 @@ class Page extends React.Component<Props, Readonly<typeof initialState>> {
 
   onTabClick(index: number) {
     this.setState({ activeTab: index })
-    this.props.onTabChange && this.props.onTabChange(this.props.tabs[index].name.toLowerCase())
+    if (isStringArray(this.props.tabs)) {
+      this.props.onTabChange && this.props.onTabChange(this.props.tabs[index].toLowerCase())
+    } else {
+      this.props.onTabChange && this.props.onTabChange(this.props.tabs[index].name.toLowerCase())
+    }
   }
 
   render() {
     const { children, title, actions, tabs, areas, fill } = this.props
     const { activeTab } = this.state
     const grid = React.Children.count(children) > 1 ? "main side" : "main"
-    const CurrentTab = tabs && tabs[activeTab].component
+    const CurrentTab = tabs && !isStringArray(tabs) && tabs[activeTab].component
 
     return (
       <Container>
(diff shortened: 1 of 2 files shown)


peterszerzo:
Looks like you're duplicating route matching logic: once inside the `activeTabName` prop, and once inside the `<Switch>`, delegating it to `react-router`. But if you're aware of the active tab name at the time you render this piece of JSX, then you can keep the children exactly where they used to be in the old API and get the same outcome. To optimize, you can conditionally not render the ones that don't match the path if you're keep on keeping the virtual dom tree light.

packages/components/src/Page/Page.tsx
```diff
@@ -20,11 +24,11 @@ export interface Props {
    * List of tabs
    * This will disable any children to render `tabs[i].component` instead
    */
-  tabs?: { name: string; component: React.ComponentType }[]
+  tabs?: { name: string; component: React.ComponentType }[] | string[]
   /**
    * Active tab name
    *
-   * Useful for easy router mapping
+   * If set, the component become controlled
```
TejasQ: 👍

packages/components/src/Page/Page.tsx
```diff
@@ -90,29 +94,23 @@ class Page extends React.Component<Props, Readonly<typeof initialState>> {
...
-      const index = props.tabs.findIndex(({ name }) => name.toLowerCase() === props.activeTabName.toLowerCase())
-      this.state = {
-        ...initialState,
-        activeTab: index === -1 ? 0 : index,
-      }
-    } else {
-      this.state = initialState
-    }
-  }
+  state = initialState
```
TejasQ: `readonly`

packages/components/src/Page/Page.tsx
```diff
@@ -90,29 +94,23 @@ class Page extends React.Component<Props, Readonly<typeof initialState>> {
...
 
   onTabClick(index: number) {
     this.setState({ activeTab: index })
-    this.props.onTabChange && this.props.onTabChange(this.props.tabs[index].name.toLowerCase())
+    if (isStringArray(this.props.tabs)) {
```
TejasQ: This can be refactored to be more DRY:

```js
const tabName = isStringArray(this.props.tabs) ? this.props.onTabChange(this.props.tabs[index].toLowerCase() : this.props.onTabChange(this.props.tabs[index].name.toLowerCase()
```

And use `tabName` later.

You could also early return if there is no `onTabChange`.

TejasQ:
I like this approach because it allows tabs to handle routes _inside_ of their tab content blocks. A big pain point with the current API is that things get messy if there is further routing to be done _inside_ of the tabs. 

Consider a case where a tab's content has to display different components for different subroutes. While we _can_ have a `Switch` with `Route` components _inside_ a subroute component, passing data becomes more complicated (really, it becomes data drilling) and state gets really easy to lose track of: I can already foresee us opening a subview and going _"why's there a `Switch` in here?"_ months from now.

I appreciate this API because it keeps things a bit more transparent.

peterszerzo:
Can you elaborate on the data drilling issue with the current API, maybe with a code sample? I don't understand that or the transparency argument.

On a different note, how do we know that the logic inside this method:
```jsx
activeTabName={/* router -> tabs matching */}
```

is consistent with whatever component wins here:
```jsx
<Switch>
  <Route path="overview" render={() => <Card title="overview" /> } />
  <Route path="jobs" render={() => <Card title="jobs" /> } />
  <Route path="functions" render={() => <Card title="functions" /> } />
  <Route path="functions/:id" render={() => <Card title="functions/:id" /> } />
</Switch>
```
and not wind up showing the overview page content with the `jobs` tab highlighted? This scenario can and should be made impossible by the design of the component. I don't see a reason why one would want route matches calculated in two different ways when you only need one, making it more difficult to debug not just the client code as well as the implementation that needs to support the two kinds of API's. We best not open up that Pandora's box in the first place.

How about simply pulling the switch statement out?
```
<Switch>
  <Route
    path="overview"
    render={() => <Page activeTab="overview" tabs={[/* tabs with the old API */]}>
  />
</Switch>
```
You're easier on the calculations and safe against route matching inconsistencies, and as far as I can tell, you can have nested routes just fine.

packages/components/src/Page/Page.tsx
```diff
@@ -20,11 +24,11 @@ export interface Props {
    * List of tabs
    * This will disable any children to render `tabs[i].component` instead
    */
-  tabs?: { name: string; component: React.ComponentType }[]
+  tabs?: { name: string; component: React.ComponentType }[] | string[]
```
peterszerzo: Not keeping tab `name` and `component` together is a magnet for bugs in my opinion. There is a strong guarantee in this API that tab name and tab contents will always be displayed together, and giving up on that is really not worth it. See alternative solution to the routing issue in the main thread of this PR.

peterszerzo:
Abandoning this PR after a verbal discussion about the concerns raised in this thread.

Some ideas for a potential future re-open in case the need arises:
* make the dangers of the API suggestion in this PR explicit by expecting a `dangerousTabNames: string[]` prop that works alongside the current `tabs` prop, which would keep its API as it currently is in master.
* adding a `to` prop to the the tab object which would take care of routing the way `<Button to="/" />` works.

PR #532, opened 2018-06-26
Close select when selecting single item

peterszerzo:
Fixes an issue where the `select` popup doesn't close when selecting a single item. This feature is enabled automatically when the select is not multiselect.

select


peterszerzo:
@fabien0102 examples and explanation added!

TejasQ:
time to rebase!

PR #533, opened 2018-06-26
Copy CNAME file to gh-pages deploys

peterszerzo:
This PR fixes https://operational-ui.js.org, which is broken because the `CNAME` file is only present on the `master` branch. It is coped into the `dist` folder before the `gh-pages` binary does the deploy.


fabien0102:
Can we use the netlify build instead? https://operational-ui.netlify.com so we don't need to update the gh-pages  anymore.

The good point is that we don't need to rebuild something when we merge a PR to master, netlify just redirect the DNS to the PR build :wink:

peterszerzo:
I don't mind where we host it, I'd simply like it for https://operational-ui.js.org to work as a nice snappy link to show off to the outside.

PR #536, opened 2018-06-27
Remove `@operational/theme` package

peterszerzo:
This PR removes the `theme` package in preparation for breaking down the monorepo into two repos.

* [x] remove from `visualizations` - parts of the theme that are used is copied over and used internally as `deprecatedTheme`. This package is now free to organize style constants however it pleases, eventually replacing this object.
* [x] remove from `components`.


packages/components/src/Breakdown/Breakdown.tsx
```diff
@@ -116,18 +102,8 @@ const Bar = styled("div")(
...
-  }: {
-    theme?: OperationalStyleConstants & {
-      deprecated: Theme
-    }
-    fill: number
-    color: string
-  }): CssStatic => {
-    const backgroundColor: string = expandColor(theme.deprecated, color) || theme.deprecated.colors.info
+  ({ theme, fill, color }: { theme?: OperationalStyleConstants; fill: number; color: string }): CssStatic => {
+    const backgroundColor: string = deprecatedExpandColor(theme.deprecated, color) || theme.deprecated.colors.info
```
ImogenF: Is there a non-deprecated version of `expandColor`? Looks like there is one being used in Chip.tsx

peterszerzo: Yes, both versions are around for the time being

PR #538, opened 2018-06-27
Feature/input copy

fabien0102:
347a9f9d-b68c-428b-b699-a1b9ed105c04

### Notes

- it's typesafe, we can't put `icon` and `copy` together
- it's magic, just add `copy` and it's works ✨


fabien0102:
My rebase failed… I need to find my commit inside ref-log…

PR #539, opened 2018-06-27
Add `right` prop to `CardColumn`

fabien0102:
image

Simple property to push the content on the right

### DIFF
--- a/packages/components/src/CardColumn/CardColumn.tsx
+++ b/packages/components/src/CardColumn/CardColumn.tsx
@@ -5,17 +5,22 @@ import { OperationalStyleConstants } from "@operational/theme"
 export interface Props {
   /** Column title */
   title?: string
+  /** Align the content to the right */
+  contentRight?: boolean
 }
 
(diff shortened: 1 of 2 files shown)


packages/components/src/CardColumn/CardColumn.tsx
```diff
@@ -5,16 +5,19 @@ import { OperationalStyleConstants } from "@operational/theme"
 export interface Props {
   /** Column title */
   title?: string
+  /** Align the content to the right */
+  right?: boolean
```
TejasQ: Possible alternatives that indicate the _intent_ of this prop could be:
- `alignContent`?
- `justify`? This would be very familiar to designers/CSS folks.
- `justifyContent`? 

I think currently, `right` would cause someone to read UI code as if the `CardColumn` _itself_ was on the `right`. 

Consider,

```jsx
<CardColumns>
  <CardColumn right>hi!</CardColumn>
  <CardColumn>hola</CardColumn>
</CardColumns>
```

I would expect the column with `hi!` to be on the right. But instead, _it's contents_ are on the right! 🤔 

Strange.

I suggest renaming this prop because of this.

PR #540, opened 2018-06-27
Fix layout on `<Layout/>`

peterszerzo:
Robust height sizing with independent scrolling between side and main areas

### DIFF
--- a/packages/components/src/Layout/Layout.tsx
+++ b/packages/components/src/Layout/Layout.tsx
@@ -17,58 +17,64 @@ export interface Props {
 const Container = styled("div")({
   overflow: "hidden",
   height: "100%",
+  // Positioned relative to display the absolutely positioned
+  // loading bar element correctly.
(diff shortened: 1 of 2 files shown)


packages/components/src/Layout/Layout.tsx
```diff
@@ -14,45 +14,62 @@ export interface Props {
...
+
 const Content = styled("div")(
   {
+    position: "relative",
     display: "grid",
-    overflow: "auto",
+    width: "100%",
+    height: "100%",
+    overflow: "hidden",
+    "& > *:nth-child(1)": {
```
TejasQ: What is all this?

peterszerzo: It is specifying strict column/row positions for the individual children. Arguably more readable/safer if we selected them as `[Main]` and `[Side]` with the babel-emotion magic. I'll refactor!

PR #544, opened 2018-06-29
Add custom icons

peterszerzo:
Custom icons live in a separate file that is auto-generated from Affinity (~Illustrator-equivalent tool used to draw them). They are 18x18, with a typical line weight of 2px. 

screen shot 2018-07-02 at 10 17 20 am


peterszerzo:
@TejasQ regarding the button with labs: it looks like brand icons are standardized at 32px and regular ones at 18px. Adding `Labs` as icon name to a button will render the standard 32px size, but since I don't think anyone will want to use the button this way, I don't see a need to protect against it further in the code. I'll fix the example to reflect this.

packages/components/src/Button/README.md
```diff
@@ -14,9 +14,9 @@ Using buttons is as simple as including the component with a text node as a chil
   <Button condensed>Condensed</Button>
 </div>
 <div style={{ marginBottom: 10 }}>
-  <Button color="success" icon="ExternalLink">Icon</Button>
-  <Button color="success" icon="Labs">Icon</Button>
-  <Button condensed icon="ExternalLink">Icon</Button>
+  <Button color="success" icon="Open">Icon</Button>
+  <Button color="success" icon="Open">Icon</Button>
```
TejasQ: This is duplication of the previous line. The reason I put the `Labs` icon in there is to include a check for brand icons in buttons visually, which this removes.

packages/components/src/ContextMenu/README.md
```diff
@@ -3,16 +3,12 @@ Context menus are nested menus that can expand from anywhere on a page. Their us
...
 
-<ContextMenu>
-  <Icon name="MoreHorizontal" size={16} />
-  <ContextMenuItem
-    onClick={() => alert("clicked")}
-  >
-    Menu 1
-  </ContextMenuItem>
+;<ContextMenu>
+  <span>Click here to trigger the context menu</span>
```
TejasQ: No more icon. Nice.

packages/components/src/Button/README.md
peterszerzo: I'll force them to be 18px then and add the `Labs` example back.

micha-f:
I think we should do away with Pantheon and Labs logos in operational for now. We don't have a use case for them.

micha-f:
@TejasQ Ok, will create a ticket to change that.

Issue #456, opened 2018-05-24
withBaseStyles should be assumed


TejasQ:
After implementing Operational UI in a number of UIs, `withBaseStyles` is being used extensively and seems to be a seemingly unnecessary piece of boilerplate. I propose we use base styles by default, and instead have a `noBaseStyling` prop or similar to allow for customization.

Thoughts?

peterszerzo:
I would say that since conflicts between injected pieces of CSS can be so frustrating to debug, this is good to have around explicitly. It also makes it clear that you're opting into the whole Operational thing, and not just an isolated snippet or a single visualization.

TejasQ:
I hear you. In contrast though, in the beginning, we said (and still adhere to) the idea that it should be an opinionated way to build operational UIs. I would argue that opting into base styles breaks that opinionation.

fabien0102:
As a new member in this project, I'm totally agree with @TejasQ , the basic bootstrap should be with a base style assumed. It's a ui kit. And for customization, normally the `theme` props should do the job, no?

peterszerzo:
Good points, fellas, let's do it! It'll be the first use of the `deprecate` HOC.

TejasQ:
Incredible.

peterszerzo:
Fixed with #553

PR #546, opened 2018-07-02
Message management inside OperationalUI

peterszerzo:
Enhances `OperationalUI` to manage messages internally.

### Usage

```tsx
import { OperationalUI, ContextConsumer } from "@operational/components"

<OperationalUI hideMessageAfter={6000}>
  <ContextConsumer>
    {ctx => (
      <Button onClick={() => {
        ctx.pushMessage({ body: "Message", type: "info" })
      }}>Generate a message</Button>
    )}
  </ContextConsumer>
</OperationalUI>
```

Notes:
* `type` may be `info`, `success` and `error`.
* state is handled internally, so messages can be created and dismissed without rendering anything or reacting to any events.
* `hideMessageAfter` defaults to 10000 (10 seconds), and only hides messages that are not of type error.

### Progress

* [x] message creation and rendering. Correct colors.
* [x] hide after implementation.
* [x] do not hide if `hideMessageAfter` is explicitly set to `0`.

screen shot 2018-07-03 at 8 55 04 am


packages/components/package.json
```diff
@@ -19,6 +19,7 @@
   },
   "scripts": {
     "start": "npm run docs:start",
+    "start:standalone": "parcel scripts/dev-server/index.html --no-hmr",
```
TejasQ: What's this for?

packages/components/package.json
```diff
@@ -35,8 +36,6 @@
     "@operational/utils": "^2.0.10",
     "emotion": "^9.2.3",
     "emotion-theming": "^9.2.3",
-    "glamor": "^2.20.40",
-    "glamorous": "^4.6.0",
```
TejasQ: Awesome!

packages/components/src/Message/Message.tsx
```diff
@@ -8,19 +8,19 @@ import { WithTheme, Css, CssStatic } from "../types"
...
   /** Called when close icon is clicked. Icon is not rendered at all if this prop is not specified. */
-
   onClose?: () => void
 }
 
 const Container = styled("div")(
   ({ theme, color }: { theme?: OperationalStyleConstants; color?: string }): CssStatic => {
-    const backgroundColor = deprecatedExpandColor(theme.deprecated, color) || theme.deprecated.colors.info
+    const backgroundColor =
+      (color && color.slice(0, 3) === "rgb" ? color : undefined) ||
```
TejasQ: What does this do? `rgb`?

packages/components/src/OperationalUI/OperationalUI.tsx
```diff
@@ -50,22 +94,72 @@ a:hover: {
...
+class OperationalUI extends React.Component<Props, State> {
+  state: State = {
+    messages: [],
+    now: new Date().getTime(),
+  }
+
+  timer: number
+
   componentDidMount() {
+    if (this.props.__log) {
```
TejasQ: Might be a useful feature actually, `this.props.debug`. 🤔

packages/components/src/OperationalUI/OperationalUI.tsx
```diff
@@ -50,22 +94,72 @@ a:hover: {
...
+      console.log("mounted")
+    }
     this.props.withBaseStyles && injectGlobal(baseStylesheet(constants))
+    this.timer = window.setInterval(() => {
```
TejasQ: Updating the `now` every second? 🤔

TejasQ:
It would be nice if we had a similar API:

```jsx
import { OperationalContext } from "@operational/components"

const MyThing = () => <OperationalContext>
{({ pushMessage }) => <div onClick={() => pushMessage(SOME_API)}>PUSH!</div>
</OperationalContext>

export default MyThing
```

TejasQ:
[Undeploy does not seem to register. Nothing happends once button is clicked.](https://trello.com/c/fpvSnfqF/239-undeploy-does-not-seem-to-register-nothing-happends-once-button-is-clicked)

packages/components/package.json
peterszerzo: The issue with styleguidist's dev server is that it re-creates instances of components inside the README.md files - which means that every time the outside `OperationalUI` checks for the time, the inside one loses its state, along with the messages. Bottom line is, the message pushing feature cannot be tested on styleguidist's server.

packages/components/src/Message/Message.tsx
peterszerzo: The expand color method doesn't pick up on `rgba(` colors and triggers the default defined inside this component. Never mind this code, though, I'm remaking it entirely.

packages/components/src/OperationalUI/OperationalUI.tsx
peterszerzo: Yes, I feel like it's nicer than managing timeouts to know when one message or another needs to hide. And if we're worried about performance, we can always stick a `shouldComponentuUpdate` returning `false` on the child.

peterszerzo:
@TejasQ regarding the trello-linked issue: see comment above about styleguidist's server destroying `OperationalUI` instances inside markdown files, causing them to lose their state (this is why I added parcel back to test the feature in isolation).

packages/components/src/OperationalUI/OperationalUI.tsx
peterszerzo: This is what I used to discover the styleguidist issue - I'll remove it in the next commit :)

packages/components/src/OperationalUI/OperationalUI.tsx
peterszerzo: @TejasQ I found a better solution that does run a check every second, but doesn't update state unless there is a message that is going away. This also fixes the styleguidist server, and makes the standalone parcel one unnecessary.

TejasQ:
[Tweak message component](https://trello.com/c/czTLc7G7/220-tweak-message-component)

peterszerzo:
@TejasQ cleverness all over!

packages/components/src/OperationalUI/OperationalUI.tsx
```diff
@@ -14,6 +18,44 @@ export interface Props {
...
+   * A time interval after which a non-error message is automatically cleared, measured in milliseconds. If the value set is `0`, no message will disappear.
+   *
+   * @default 10000 (10s)
+   */
+  hideMessageAfter?: number
+}
+
+export interface State {
+  messages: {
+    message: {
```
TejasQ: Consider an interface for this because the same shape is used on line 43.

packages/components/src/OperationalUI/OperationalUI.tsx
```diff
@@ -14,6 +18,44 @@ export interface Props {
...
+      body: string
+      type: MessageType
+    }
+    addedAt: number
+  }[]
+}
+
+export interface Context {
```
TejasQ: This already exists in `types.ts` 🤔

packages/components/src/OperationalUI/OperationalUI.tsx
```diff
@@ -14,6 +18,44 @@ export interface Props {
...
+    case "info":
+      return "primary"
+    case "success":
+      return "success"
+    case "error":
+      return "error"
+  }
 }
 
 const { Provider, Consumer } = React.createContext({})
```
TejasQ: If context is initialized with default values, you don't even need typedefs because they're inferred for context throughout your application.

packages/components/src/OperationalUI/OperationalUI.tsx
```diff
@@ -50,9 +92,39 @@ a:hover: {
...
+  }
+
+  timer: number
+
+  removeOutdatedMessages() {
+    if (this.props.hideMessageAfter === 0) {
+      return
+    }
+    this.setState(prevState => {
+      const now = new Date().getTime()
```
TejasQ: I'm not as comfortable with this component managing all this state. Can we add a `timeout` prop to `Message` and have them self destruct accordingly? That would give users more control too.

packages/components/src/OperationalUI/OperationalUI.tsx
```diff
@@ -63,9 +135,33 @@ class OperationalUI extends React.Component<Props, {}> {
...
+            },
           }}
         >
-          {children}
+          <>
+            <Messages>
+              {this.state.messages.map(({ message, addedAt }, index) => (
+                <Message
+                  key={index}
+                  color={colorByMessageType(message.type)}
```
TejasQ: Maybe we don't even need this function, we can just say use `primary` as your `message.type` if you want to display informative messages.

packages/components/src/OperationalUI/OperationalUI.tsx
```diff
@@ -63,9 +135,33 @@ class OperationalUI extends React.Component<Props, {}> {
...
+                  onClose={() => {
+                    this.setState(prevState => ({
+                      messages: prevState.messages.filter(
+                        (message, filteredMessageIndex) => filteredMessageIndex !== index,
```
TejasQ: `message` is unused. Consider `_` to better describe intent to other developers.

packages/components/src/types.ts
```diff
@@ -1,15 +1,22 @@
...
 
-export type CssStatic = CSSProperties
+export type CssStatic = {}
+
+export type MessageType = "info" | "success" | "error"
 
 export interface Context {
-  pushState?: (path: string) => void
-  replaceState?: (path: string) => void
+  pushState?: (url: string) => void
```
TejasQ: I smell duplication.

packages/components/src/OperationalUI/OperationalUI.tsx
```diff
@@ -50,9 +92,39 @@ a:hover: {
...
+      if (prevState.messages.length > filteredMessages.length) {
+        return {
+          messages: filteredMessages,
+        }
+      }
+    })
+  }
+
   componentDidMount() {
     this.props.withBaseStyles && injectGlobal(baseStylesheet(constants))
```
TejasQ: `withBaseStyles` is unused.

packages/components/src/OperationalUI/README.md
```diff
@@ -1,47 +1,51 @@
 Main provider for Operational UI. Should need to wrap all your application with this component.
 
 ### Classic example
+
 ```jsx static
 <OperationalUI withBaseStyles>
```
TejasQ: `withBaseStyles` is unused.

packages/components/src/OperationalUI/OperationalUI.tsx
```diff
@@ -63,9 +135,33 @@ class OperationalUI extends React.Component<Props, {}> {
...
+              this.setState(prevState => ({
+                messages: [...prevState.messages, { message, addedAt: new Date().getTime() }],
+              }))
+            },
           }}
         >
-          {children}
+          <>
+            <Messages>
+              {this.state.messages.map(({ message, addedAt }, index) => (
```
TejasQ: `addedAt` is unused.

TejasQ:
To follow up an offline conversation about this implementation: I had an issue with the unnecessary interval: if there are no messages that need to be garbage collected, every UI that implements Operational will have a ticking timer for no apparent reason.

We resolved this by deciding to _only_ start an interval _when_ we have messages, and clear it when the message queue is empty.

packages/components/src/OperationalUI/OperationalUI.tsx
peterszerzo: `primary` feels like a misleading way to refer to a type of message, especially one that is probably less important than a `success` or an `error` type. Let's keep it `info` I'd say.

packages/components/src/types.ts
peterszerzo: Where exactly?

packages/components/src/OperationalUI/OperationalUI.tsx
peterszerzo: It is the same shape inside the `message` field - it's just that the `body + type` combo needs to be packed inside something additionally so we can have `addedAt` be there as well.

packages/components/src/OperationalUI/OperationalUI.tsx
peterszerzo: This is resolved per discussion summary below in the newer commits.

packages/components/src/OperationalUI/README.md
peterszerzo: `withBaseStyles` is actually not deprecated yet - it is still necessary to render the global styles (#456 still open). Will take care of that in a separate PR.

micha-f:
To document some thoughts here so I don't forget about them:
- 99% of the time there will be no (zero) current messages displayed
- When messages are displayed, 99% of the time it will be a single message

=> I don't think we necessarily need a global message queue and queue clearing interval.

Why don't we just create a simple timeout for each message? Once it's done, the message is removed. Nice and self-contained.

On another note, is there any way to make something like this work inside any component (using context and or portals)? That would allow us to completely manage the messages from the components that created them (even remove them at any time if still active).

```
render() => {
  return(
  	<div>
  	  <p>Amazing content</p>  	  
  	  <GlobalMessage duration="15"> {/* Displayed through portal outside of component DOM*/}
  	  	<Message color="error">
  		  Oh no… An error occured :(
		</Message>
  	  </GlobalMessage>
  	</div>
  )
}
```

peterszerzo:
@micha-f adding intervals to `Message` components will require setting up child-parent communication (so they can unset themselves in the state of the parent) and additional checks, and will make animating exits in the future trickier. It would split coherent logic into two places and make child components less presentational.

I feel like handling this globally makes sense because the messages are global - you feed them into a funnel and then not think about them anymore. We already have error message elements for form fields and cards that can be used locally.

micha-f:
@peterszerzo Regardless of how we decide on the interval / timeout - do you think there is a way to implement something that allows for the use of a `GlobalMessage` component as in the example above? It would keep all the implementation hidden and you could treat the GlobalMessage like a normal child. I was thinking that internally the `GlobalMessage` would then use `React.createPortal`.

peterszerzo:
A summary of what we discussed on the `GlobalMessages` topic:
- the current approach takes care of both state management and rendering, and comes with the least amount of code. @micha-f and I agreed that this is the preferred approach.
- the one drawback of this implementation is that an undo or retry functionality cannot be implemented because even if we set a retry method in the messages state on `OperationalUI`, there is no guarantee that the component in question will still be alive at the time the retry is triggered. The current recommendation is that if this is needed, the library user should opt out of this global message management and handle state themselves (example on the website under the `Messages` presentational component).

packages/components/src/OperationalUI/README.md
```diff
@@ -51,5 +55,63 @@ class RoutingComponent extends React.Component {
...
+You can use `OperationalUI`'s internal flash message feature to automatically render and manage messages by simply using the `pushMessage` method provided in context, as shown in the code snippet below:
+
+```jsx
+const ContextConsumer = require("../").ContextConsumer
+
+class MessageHandler extends React.Component {
+  render() {
+    return (
+      <OperationalUI>
+        <ContextConsumer>
```
TejasQ: Can we rename this export to `OperationalContext` or something similar?

packages/components/src/types.ts
TejasQ: It was removed. 🤔

TejasQ:
I am still not comfortable with this implementation and still think we can do better. 

Since this isn’t a blocker, I suggest we take some time and think it through and have discussions a bit more.

micha-f:
@TejasQ I believe that this is a good initial solution - keeping a single 2s interval running globally to remove expired messages. Why:

- Very simple to reason about
- No measurable performance penalty. A single interval timer is negligible.
- Zero risk of race conditions due to starting and stopping timers

We can always come back to this later but I feel like the current solution does not have obvious flaws that need to be fixed.

TejasQ:
I disagree. All of my concerns and counter arguments are described in #558.

Further, I would argue that a 2 second interval continuously running through all `OperationalUI` apps on all open tabs irrespective of whether it is need or not is not easier to reason about to the user. Having code that does that is _also_ not easier to reason about to newer teammates. The logic in `setState` here is _also_ not easy to reason about.

I agree about performance, but I do not see that as a reason to merge this. _"It works so it's okay"_ is not good enough, I'm sorry.

All I'm asking is that we consider possible alternatives and not rush to get this merged. I do not believe we are in a hurry with this feature.

peterszerzo:
@TejasQ you are certainly welcome to branch off of this PR and present an alternative solution, although I would like to ask you to be mindful of our time. I would love to give the approach a spin inside an application and see how it works with realistic needs both on a public API and a UX level. It might even help us make a more informed decision on the optimizations we want to add to this solution. Or at the very least, save the need for all those rebases.

peterszerzo:
@TejasQ your worked is merged in and it is all rebased and ready to go as far as I'm concerned.

packages/components/src/OperationalUI/OperationalUI.tsx
```diff
@@ -14,9 +17,53 @@ export interface Props {
...
+
+const colorByMessageType = (type: MessageType): string => {
+  switch (type) {
+    case "info":
+      return "primary"
+    case "success":
+      return "success"
+    case "error":
+      return "error"
+  }
```
TejasQ: We really should handle a `default` case here. Maybe even return `type` as the default. This would allow us to chop the preceding 4 lines with cases.

packages/components/src/OperationalUI/OperationalUI.tsx
```diff
@@ -50,22 +97,84 @@ a:hover: {
...
   render() {
-    const { withBaseStyles, pushState, replaceState, children } = this.props
+    const { pushState, replaceState, children } = this.props
     return (
       <ThemeProvider theme={constants}>
         <Provider
           value={{
             pushState,
             replaceState,
+            pushMessage: (message: Message) => {
```
TejasQ: IMO this is more readable if the function body lives elsewhere in this class. Just my opinion. If this is more readable to you, @peterszerzo, alles gut.

packages/components/src/OperationalUI/OperationalUI.tsx
```diff
@@ -50,22 +97,84 @@ a:hover: {
...
         >
-          {children}
+          <>
+            <Messages>
+              {this.state.messages.map(({ message }, index) => (
+                <MessageComponent
+                  key={index}
+                  color={colorByMessageType(message.type)}
+                  onClose={() =>
+                    this.setState(prevState => ({
```
TejasQ: Same comment as on line 148.

packages/components/src/Message/Message.tsx
```diff
@@ -1,37 +1,34 @@
 import * as React from "react"
 import styled from "react-emotion"
-import { OperationalStyleConstants, deprecatedExpandColor } from "../utils/constants"
+import { OperationalStyleConstants, expandColor } from "../utils/constants"
 import { readableTextColor, darken } from "@operational/utils"
```
TejasQ: afaik these aren't used.

packages/components/src/Message/Message.tsx
```diff
@@ -1,37 +1,34 @@
...
 import { Icon } from "../"
+import colorCalculator from "tinycolor2"
 import { WithTheme, Css, CssStatic } from "../types"
```
TejasQ: `Css` isn't used either.

packages/components/src/OperationalUI/OperationalUI.tsx
peterszerzo: This is important for clarity in my opinion - in a sense, the one `"success"` and the other `"success"` belong to different types (one to `Position`, one to `keyof OperationalColors` or whatever that interface is called). I'm looking to make that clear by writing this out explicitly.

packages/components/src/OperationalUI/OperationalUI.tsx
peterszerzo: I don't see it as a readability issue, especially since we have much more complicated components. I've worked on codebases where defining instance methods for everything was set as a convention, and I found it super hard to follow + so much more naming to do.

packages/components/src/OperationalUI/OperationalUI.tsx
TejasQ: Still, what do you think about having a default case?

peterszerzo: `MessageType` is defined as `"info" | "success" | "error"`, so you shouldn't need one in addition to the cases you have. Could add a `let default: never` at the end, but the TS compiler yelled at me when I tried to do that.

PR #548, opened 2018-07-03
Convert color constants to hex.

ImogenF:
- All colors are now in HEX format, except `ghost` which remains in hsla.
- Greys are now hardcoded.
- The `success` color has been updated to `"#0C991D"`


ImogenF:
@TejasQ I wasn't sure about browser support for 8-digit hex codes - but also, I feel like it's less clear at a glance what `#ffffff33` means, whereas it's instantly obvious that `hsla(0, 0%, 100%, 0.2)` is white with an opacity of 0.2...

TejasQ:
Good point. We can keep the HSLA/RGBA as is.

PR #550, opened 2018-07-03
Tooltip enhancements

peterszerzo:
Fine-tunes tooltip component as per https://trello.com/c/6t2zTG0M/221-fine-tune-tooltip-component. Also fixes a safari bug where the tooltip can't shrink to very short texts, using off-screen rendering.

screen shot 2018-07-03 at 3 19 40 pm

### DIFF
--- a/packages/components/src/Tooltip/Tooltip.Container.tsx
+++ b/packages/components/src/Tooltip/Tooltip.Container.tsx
@@ -1,0 +1,144 @@
+import * as React from "react"
+import styled from "react-emotion"
+import colorCalculator from "tinycolor2"
+
+import { OperationalStyleConstants } from "../utils/constants"
+import { Css } from "../types"
+
+export type Position = "top" | "left" | "right" | "bottom"
+
+const makeContainerPositionStyles = (position: Position) => {
+  switch (position) {
+    case "top":
+      return {
+        left: "50%",
+        top: -6,
+        transform: "translate3d(-50%, -100%, 0)",
+      }
+    case "bottom":
+      return {
+        left: "50%",
+        top: "100%",
+        transform: "translate3d(-50%, 6px, 0)",
+      }
+    case "left":
+      return {
+        top: "50%",
+        left: -6,
+        transform: "translate3d(-100%, -50%, 0)",
+      }
+    case "right":
+      return {
+        top: "50%",
+        right: -6,
+        transform: "translate3d(100%, -50%, 0)",
+      }
+  }
+}
+
+const makeCaretPositionStyles = (position: Position, backgroundColor: string): {} => {
+  switch (position) {
+    case "top":
+      return {
+        bottom: -4,
+        left: `calc(50% - 6px)`,
+        borderLeft: "6px solid transparent",
+        borderRight: "6px solid transparent",
+        borderTop: `6px solid ${backgroundColor}`,
+      }
+    case "bottom":
+      return {
+        top: -4,
+        left: `calc(50% - 6px)`,
+        borderLeft: "6px solid transparent",
+        borderRight: "6px solid transparent",
+        borderBottom: `6px solid ${backgroundColor}`,
+      }
+    case "left":
+      return {
+        right: -4,
+        top: `calc(50% - 6px)`,
+        borderTop: "6px solid transparent",
+        borderBottom: "6px solid transparent",
+        borderLeft: `6px solid ${backgroundColor}`,
+      }
+    case "right":
+      return {
+        left: -4,
+        top: `calc(50% - 6px)`,
+        borderTop: "6px solid transparent",
+        borderBottom: "6px solid transparent",
+        borderRight: `6px solid ${backgroundColor}`,
+      }
+  }
+}
+
+export const Container = styled("div")(
+  ({
+    position,
+    offScreenWidthTest,
+    singleLineTextWidth,
+    theme,
+  }: {
+    position: Position
+    offScreenWidthTest?: boolean
+    singleLineTextWidth: number
+    theme?: OperationalStyleConstants
+  }) => {
+    const backgroundColor = colorCalculator(theme.color.black)
+      .setAlpha(0.9)
+      .toString()
+    return {
+      backgroundColor,
+      label: "tooltip",
+      color: theme.color.white,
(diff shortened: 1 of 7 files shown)


packages/components/src/Tooltip/Tooltip.tsx
```diff
@@ -1,179 +1,216 @@
...
+      .toString()
+    return {
+      backgroundColor,
+      label: "tooltip",
+      color: theme.color.white,
       position: "absolute",
-      zIndex: theme.deprecated.baseZIndex - 1,
-      width: 0,
-      height: 0,
+      zIndex: theme.deprecated.baseZIndex + 101,
```
TejasQ: In the new theme constants, there are `zIndex`es for various chunks of UI.

packages/components/src/Tooltip/Tooltip.tsx
```diff
@@ -1,179 +1,216 @@
...
+      borderRadius: 2,
+      boxShadow: theme.deprecated.shadows.popup,
```
TejasQ: I think this can be a hardcoded shadow value.

packages/components/src/Tooltip/Tooltip.tsx
```diff
@@ -1,179 +1,216 @@
...
+            top: -200,
+            left: -200,
+          }
+        : {
+            // If there was an issue determining singleLineTextWidth, default to the 150px width
+            // Otherwise, honor the single line text width unless greater than 150px.
+            width: singleLineTextWidth === 0 ? 150 : Math.min(singleLineTextWidth + 4, 150),
+          }),
       ...(() => {
         if (position === "top") {
```
TejasQ: Can we get styles _outside_ of an IIFE arbitrarily in an object and probably use `switch` over `if` since it more clearly communicates the intent that the condition relies on _one_ discriminant (`position`)?

packages/components/src/Tooltip/Tooltip.tsx
```diff
@@ -1,179 +1,216 @@
...
+        left: theme.deprecated.spacing * -2,
+        display: "block",
+        width: theme.deprecated.spacing * 2,
+        height: "100%",
+      },
+      // They say behind every great tooltip is a great caret.
+      "&::before": {
+        content: "''",
+        position: "absolute",
+        zIndex: theme.deprecated.baseZIndex - 1,
```
TejasQ: 🤔

packages/components/src/Tooltip/Tooltip.tsx
```diff
@@ -1,179 +1,216 @@
...
+        width: 0,
+        height: 0,
+        ...(() => {
```
TejasQ: _Another_ IIFE that conditionally returns styles?

ImogenF:
Is there a reason why the tooltip that is *above* the element (in this case the one that says "I uncover them all") is more offset than the others?

peterszerzo:
@ImogenF good point, I'll look into it!

packages/components/src/Tooltip/Tooltip.tsx
```diff
@@ -1,179 +1,206 @@
...
-    padding: `${theme.deprecated.spacing / 3}px ${(theme.deprecated.spacing * 2) / 3}px`,
-    borderRadius: 2,
-    boxShadow: theme.deprecated.shadows.popup,
-    ...(() => {
-      if (position === "top") {
-        return {
-          left: "50%",
-          transform: "translate3d(-50%, calc(-100% - 6px), 0)",
-        }
+const containerPositionStyles = (position: Position): {} => {
```
TejasQ: Is the type annotation necessary?

TejasQ: Since these are factory functions, can we name them `makeContainerPositionStyles` or something similar?

packages/components/src/utils/mixins.ts
```diff
@@ -46,18 +48,17 @@ export const FormFieldControl = styled("div")(
     "& svg": {
       opacity: 0.4,
       position: "relative",
-      top: -1,
     },
     // :nth-child(2) refers to the tooltip
-    "& > :nth-child(2)": {
-      display: "none",
+    [formFieldControlTooltipSelector]: {
```
TejasQ: But this doesn't really solve anything. Can we not create styled components with these styles? 🤔

peterszerzo:
@ImogenF looks like top-positioned tooltip styles were broken - fixed and looking like the updated screenshot at the top of this page. Appreciate the catch!

packages/components/src/Tooltip/Tooltip.tsx
```diff
@@ -1,181 +1,209 @@
...
- * <Tooltip top/>, as opposed to <Tooltip position="top"/>
- * It gets translated into the Position type inside the component, so it allows for a more
- * straightforward implementation.
+ * In order to allow for tooltips that have a sensible max-width that adjusts its width for shorter text,
+ * and in order to have that working reliably across browsers, this implementation renders the tooltip offscreen
+ * in order to determine how wide it would be were it to not do line breaks at any width.
+ * The actual tooltip is rendered with this information extracted from the DOM node.
  */
+
 export interface Props {
```
TejasQ: Can we add a `/** @todo Strongly type this interface */` here?

packages/components/src/utils/mixins.ts
peterszerzo: I was having an issue exporting the `Container = styled.div("")` from the tooltip component - which would make this bulletproof - because of styleguidist yelling not finding the actual tooltip component export. I'll see if there's a way to fix this up in an idiomatic way.

packages/components/src/Tooltip/Tooltip.Container.tsx
```diff
@@ -0,0 +1,144 @@
...
+          }),
+      ...makeContainerPositionStyles(position),
+      // This pseudo-element extends the clickable area of a tooltip extending enough to disappear as the mouse moves over to the caret.
+      "&::after": {
+        content: "''",
+        position: "absolute",
+        top: 0,
+        left: -32,
+        display: "block",
+        width: -32,
```
TejasQ: I am so embarrassed that I approved this. 😕

PR #554, opened 2018-07-03
Added cases to components that we'll need for other UIs

TejasQ:
This PR adds the possibility to have a full-width login button, and a stacked Contiamo logo in order to support the following screenshot:

image

Also updated `Input` to support a `fullWidth` prop.

### DIFF
--- a/packages/components/src/Button/Button.tsx
+++ b/packages/components/src/Button/Button.tsx
@@ -1,10 +1,9 @@
 import * as React from "react"
 import styled, { Interpolation } from "react-emotion"
-import { readableTextColor, darken, lighten } from "@operational/utils"
+import { readableTextColor, darken } from "@operational/utils"
 
 import { OperationalStyleConstants, expandColor } from "../utils/constants"
 import { isWhite, isModifiedEvent } from "../utils"
-import { Css } from "../types"
 import { ContextConsumer, Context, Icon, IconName } from "../"
 import Spinner from "../Spinner/Spinner"
 
@@ -26,6 +25,8 @@ export interface Props {
   disabled?: boolean
   /** Condensed option */
   condensed?: boolean
+  /** Should the button fill its container? */
+  fill?: boolean
   children?: React.ReactNode
 }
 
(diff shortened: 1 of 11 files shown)


TejasQ:
[Integrate new design into IdP Login Screen](https://trello.com/c/vxJGwOdE/210-integrate-new-design-into-idp-login-screen)

packages/components/src/Button/Button.tsx
```diff
@@ -26,6 +25,8 @@ export interface Props {
   disabled?: boolean
   /** Condensed option */
   condensed?: boolean
+  /** Should the button fill its container? */
+  fill?: boolean
```
peterszerzo: I think `fillContainer` would make this clearer - otherwise I'd be thinking of fill colors.

packages/components/src/Button/Button.tsx
```diff
@@ -128,16 +127,11 @@ const Button = (props: Props) => {
...
-          loading={props.loading}
-          disabled={props.disabled}
-          condensed={props.condensed}
-          icon={props.icon}
           title={props.loading && props.children === String(props.children) ? String(props.children) : undefined}
         >
           {props.children}
           {props.icon && (
-            <IconContainer condensed={props.condensed}>
+            <IconContainer>
```
peterszerzo: Nice legacy catch :)

packages/components/src/Button/Button.tsx
```diff
@@ -128,16 +127,11 @@ const Button = (props: Props) => {
...
               {typeof props.icon === "string" ? <Icon right name={props.icon as IconName} size={18} /> : props.icon}
```
peterszerzo: I think we should remove the flexibility here and expect icon names only - if we find more needs for icons in buttons in the future, we can just add more named icons to the list. Would it make sense to take care of that while we're at it?

packages/components/src/Button/Button.tsx
TejasQ: Would `fullWidth` be even better?

PR #555, opened 2018-07-04
Enhance spinner component

peterszerzo:
This PR addresses two issues of the current spinner (criticizing my own design here):
- it doesn't spin.
- it has Microsoft associations.
- it is a bit out of place in character for an interface that is trying to be straightforward and approachable.

Current spinner:

screen shot 2018-07-04 at 2 03 43 pm

The replacement (click the netlify link to see it spinning):

contiamo-spinners

We have two spinners, `<Spinner />` and `<Spinner bounce />` - the first one is the general-purpose regular spinning circular shape, the other is made up of bouncing squares that are meant to indicate that data already available on the screen is being updated.

This PR also fixes the bundle, sync and unlock icons:

screen shot 2018-07-05 at 2 41 40 pm
screen shot 2018-07-05 at 2 41 47 pm
screen shot 2018-07-05 at 2 41 53 pm

### Progress

* [x] add approximate shape/rotation/API
* [x] redraw so it matches with other icons
* [x] fix Bundle, Sync and Unlock icons
* [x] refine API according to needs


micha-f:
@peterszerzo They look really nice 👍 

I believe we will require two different types of spinners:
1) as a general symbol for fetching data / executing an async operation
2) as a symbol for an ongoing live-polling / subscription

I think that your spinners work well for 1).

Maybe we can add something like a moving wave (left to right) for 2) ?

peterszerzo:
@micha-f something like this maybe?

screen shot 2018-07-04 at 2 33 50 pm

packages/components/src/Spinner/Spinner.tsx
```diff
@@ -5,63 +5,38 @@ import { WithTheme, Css, CssStatic } from "../types"
...
-      animationDelay: `${-2 * f * animationTimeUnit}s`,
-    },
-    "&:nth-child(3)": {
-      animationDelay: `${-3 * f * animationTimeUnit}s`,
-    },
-  }
-})
+const Container = styled("div")(
+  ({ size, color, theme }: { size?: number; color?: string; theme?: OperationalStyleConstants }) => ({
+    label: "spinner",
```
TejasQ: We don't need no explicit labels 🎶 
We have automatic labels now. :D

packages/components/src/Spinner/Spinner.tsx
```diff
@@ -86,12 +80,36 @@ const RegularSpinner = () => (
...
+  height: "100%",
+  position: "relative",
+})
+
+/**
+ * The bouncing spinner is constructed out of 3 80x80 boxes spaced out horizontally on a 360x360 grid (these units don't refer to pixels, they simply mimic the grid of the icons.
+ * The math used in here lays these boxes out so they're vertically centered and spaced equally on the horizontal axis without any gutter.
+ */
+const BouncingSpinnerBox = styled("div")(({ no }: { no: number }) => ({
+  width: `${(80 / 360) * 100}%`,
```
TejasQ: Can we not just hardcode `22.22%` in here?

TejasQ:
@micha-f thoughts? [Link for convenience](https://deploy-preview-555--operational-ui.netlify.com/#!/Spinner)

packages/components/src/Spinner/Spinner.tsx
peterszerzo: I'd rather not - this expresses how the number was worked out from the geometric properties of the boxes, and will make things easier to update.

Performance-wise, the impact isn't measurable, and even if it were, I think it is safe to assume that a tool like https://prepack.io or the JIT will have our back eventually.

PR #556, opened 2018-07-04
NameTag component

peterszerzo:
Adds the square-shaped avatar variations as a separate component called `NameTag`. I decided to go for a different component since avatars, icons and nametags are used in different contexts and I surmize that they will evolve independently in terms of both style and API.

screen shot 2018-07-04 at 4 53 23 pm


packages/components/src/NameTag/NameTag.tsx
```diff
@@ -0,0 +1,54 @@
...
+export interface Props {
+  color?: string
+  /**
+   * Adds a right margin to create proper spacing between this element and neighboring content.
+   */
+  left?: boolean
+  /**
+   * Adds a left margin to create proper spacing between this element and neighboring content.
+   */
+  right?: boolean
```
ImogenF: Surely `left` should add a left margin?

TejasQ: Actually, the way I see it, `left` says _”hey, I’m on the left!!!”_ and puts a space between it and whatever is to its _right_.

ImogenF: In that case, maybe the comments could make it a bit clearer that `left` means positioning the name tag to the left by adding a margin on the right. As it is, it's a bit confusing

TejasQ:
@peterszerzo my apologies! Tapped the close button by accident.

packages/components/src/NameTag/NameTag.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+const Container = styled("div")(
+  ({
+    theme,
+    color,
+    left,
+    right,
+  }: {
+    theme?: OperationalStyleConstants
+    color?: string
```
TejasQ: Consider `Props["color"]` here since its a subset of `Props`.

PR #558, opened 2018-07-04
Message Management Improvement Proposal

TejasQ:
I tried to sleep but I couldn't sleep thinking about this feature.

I'd like to propose these changes to the controversial PR #546 because it solves the following issues that I have with the current message management PR (#546):

- The current PR starts a timer on mount and leaves it there for no reason. Operational UI just keeps on ticking. I do not think this is good behavior, especially for a library. What happens when multiple tabs of Operational UIs are open? We have multiple timers. _But why?_ It is not memory intensive, I get that, _but why?_ After considering @peterszerzo's counter arguments about readability and the potential for introducing bugs, I propose [this simple solution](https://github.com/contiamo/operational-ui/commit/049d78e676481816b2b34cd95ac4422e0faa7709) that does nothing on mount, but only handles timers depending on whether there are messages in the queue or not. I have tested it exhaustively and it works. I would be very interested to hear any concerns around this approach.

- The current PR imports an interface `Context` from `../types` and uses it internally, [while all dependents on this interface in the codebase `import { Context } from "../"`](https://github.com/contiamo/operational-ui/pull/546/commits/1adb7e99693058e88707da35294111c3bc41363f#diff-7ad5f5c33bf8dabc4ec80287dbae9758R5). Upon working with this, I had trouble tracking down where this interface was coming from and I think it [would make more sense to place it in `OperationalUI.tsx`](https://github.com/contiamo/operational-ui/commit/57843cb50e35823ce88d8cf6fe033d343c816c13) for this reason.

- The current PR has a [_large_ amount of logic in `this.setState()`](https://github.com/contiamo/operational-ui/pull/546/files#diff-82e2290f26217200385396199d6f6c6cR107). I do not find this easy to reason about and quite frankly, could see how others on the team could get lost in this due to the fact that setState calls are typically concise, describing relevant state updates. [I have moved this logic around in order to make it more readable.](https://github.com/contiamo/operational-ui/commit/93e87b4b2b028b4e09b9f14c0a7b66709fd4a63f)

- The current PR has multiple lines of `//` comments. This PR uses multiline comments.

- The current PR defines an interface `Message` [but continues to use a primitive `{ body: string, type: MessageType }` throughout its implementation](https://github.com/contiamo/operational-ui/pull/546/files#diff-82e2290f26217200385396199d6f6c6cR142). I find this unnecessary and propose we [update it to use the interface throughout the implementation](https://github.com/contiamo/operational-ui/commit/93e87b4b2b028b4e09b9f14c0a7b66709fd4a63f#diff-82e2290f26217200385396199d6f6c6cR148).

## A Visual Representation

Perhaps this PR is justified better visually. Consider the following examples.

### Status of #546 
ezgif com-resize

### Status of #558 (this PR)
ezgif com-resize 1

## Important
image
Please not that I have not/do not want to re-implement this feature. That is _not_ what this PR is for. It is just to make my points clear and, if we agree that they are valid, merge them into the current message management PR (#546).


peterszerzo:
Looking good! I see this is similar to where we left off our conversation on Tuesday, and what I had implemented two commits back in the original branch: https://github.com/contiamo/operational-ui/pull/546/commits/39d9d4518f2f43d6078646f9fa9e8d083c65f621. We dismissed it after a brief chat with @micha-f yesterday on conceptual grounds that are perhaps tricky to argue, but I am not fundamentally opposed.

I'd say we merge :)

PR #560, opened 2018-07-05
Enhance visual tests website

peterszerzo:
* [x] fix compile errors after icons upgrade etc.
* [x] add different visualization color schemes.
* [x] move visual test cases project into visualizations.

screen shot 2018-07-05 at 4 21 40 pm


ImogenF:
While you're working on this, can we discuss the colour assignment issue? 
<blockquote class="trello-card"><a href="https://trello.com/c/h6sBO7P2/252-fix-automatic-colour-assignment">Fix automatic colour assignment.</a></blockquote><script src="https://p.trellocdn.com/embed.min.js"></script>

I've figured out that the colour assignment issue is caused by old instances of the visualisation `colorAssigner` being used, which shouldn't be the case since each visual test creates a new `Chart` instance, and each `Chart` instance should use a fresh colorAssigner. So not really sure what's going on here, but it's certainly not a problem I've had when running visualisations on my local dev server, so seems linked to the visual tests setup somehow.

ImogenF:
Looks great, but is the `View Code` button meant to be working?

peterszerzo:
@ImogenF thanks for the catch - this is going to be fixed in #586.

package.json
```diff
@@ -37,7 +35,7 @@
     "deploy": "gh-pages -d dist"
   },
   "lint-staged": {
-    "*.(js|jsx|json|html|md)": [
+    "*.(js|jsx|json|md)": [
```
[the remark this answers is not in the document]
fabien0102: sure?

ImogenF:
Noticed another potential bug while working on the visual tests - `warning.js:33 Warning: validateDOMNesting(...): <div> cannot appear as a descendant of <p>`. Looks like it's something to do with the spinner....?

packages/visualizations/package.json
```diff
@@ -12,6 +12,7 @@
   },
   "scripts": {
     "start": "parcel scripts/dev-server/index.html",
+    "start:visual:tests": "parcel visual-tests/index.html",
```
fabien0102: This key is strange, it should be `visual-tests` or `test:visual` or `start:visual-tests` (the `:` means a command subdomain)

fabien0102: Can you add `"visual-test:build": "rimraf dist && parcel build visual-tests/index.html --public-url ./"` and try to add the visual-tests in the netlify build?

PR #562, opened 2018-07-06
Hint component

peterszerzo:
The `Hint` component simply renders a question icon + a hover tooltip. It matches in color with table headings so they can be used in there seamlessly as `<Hint right>I help with your column</Help>`.

screen shot 2018-07-06 at 10 07 28 am


packages/components/src/utils/mixins.ts
```diff
@@ -35,6 +35,19 @@ export const FormFieldControls = styled("div")({
   right: 0,
 })
 
+export const hoverTooltip: {} = {
+  [TooltipContainer as any]: {
+    opacity: 0.01,
```
TejasQ: I'm curious. Why this opacity rule?

TejasQ:
[Hint component](https://trello.com/c/11XpudIW/244-hint-component)

peterszerzo:
Makes sense, @TejasQ and @ImogenF, thanks for the feedback. Stay tuned for the update!

packages/components/src/utils/mixins.ts
peterszerzo: So it's always rendered - otherwise the length can't be measured.

packages/components/src/Hint/Hint.tsx
```diff
@@ -0,0 +1,39 @@
...
+  right?: boolean
+}
+
+const Container = styled("div")(
+  ({ left, right, theme }: { left?: boolean; right?: boolean; theme?: OperationalStyleConstants }) => ({
+    position: "relative",
+    display: "inline-block",
+    verticalAlign: "middle",
+    color: theme.color.text.lightest,
+    ...(left ? { marginRight: theme.space.base } : {}),
```
TejasQ: [This](https://github.com/contiamo/operational-ui/pull/569#discussion_r200903530)

packages/components/src/Input/README.md
```diff
@@ -33,6 +33,12 @@ class StatefulInput extends React.Component {
 <Input value="12" label="Phone number" hint="Please use country code" />
 ```
 
+### With toggle state
+
+```jsx
+<Input value="Database Name" label="Name" onToggle={() => {}} disabled={true} hint="This value cannot be changed" />
```
TejasQ: is `onToggle` needed?

packages/components/src/utils/mixins.ts
TejasQ: How about `visibility: hidden`? I think that seems to more clearly describe the intent.

packages/components/src/Input/README.md
peterszerzo: It is - it toggles the disabled state for form fields where we want to make it just a little bit harder to change sensitive fields such as schema name (which may break psl code etc.)

TejasQ: I mean in this example.

packages/components/src/Hint/Hint.tsx
```diff
@@ -0,0 +1,39 @@
...
+import { Icon, Tooltip } from "../"
+import { hoverTooltip } from "../utils/mixins"
+
+export interface Props {
+  className?: string
+  children?: React.ReactNode
+  /**
+   * Indicates that this component is left of other content, and adds an appropriate right margin.
+   */
+  left?: boolean
```
TejasQ: Since we're spreading props in, why not `Props["left"]` and so on?

packages/components/src/utils/mixins.ts
```diff
@@ -2,7 +2,7 @@ import React from "react"
 import styled from "react-emotion"
 import { lighten } from "@operational/utils"
 import { OperationalStyleConstants } from "./constants"
-import TooltipContainer from "../Tooltip/Tooltip.Container"
+import { dangerousContainerClassName as dangerousTooltipContainerClassName } from "../Tooltip/Tooltip"
```
TejasQ: Why not just `dangerousTooltipContainerClassName`?

packages/components/src/utils/mixins.ts
```diff
@@ -35,26 +35,37 @@ export const FormFieldControls = styled("div")({
   right: 0,
 })
 
+export const hoverTooltip: {} = {
```
TejasQ: Is this mixin used anywhere except here? 🤔 We might be able to do away with the entire computed-property-with-classname thing completely.

peterszerzo: Not at the moment, but it is highly likely it will be reused. How would it help with doing away with the classname?

TejasQ: The styles can be self-contained (and therefore, components) in `Hint.tsx`. I suggest we don't run on the assumption that it'll be reused for now and put this logic with `Hint`, the only place it's used.

packages/components/src/Hint/Hint.tsx
```diff
@@ -0,0 +1,39 @@
+import * as React from "react"
+import styled from "react-emotion"
+import { OperationalStyleConstants } from "../utils/constants"
```
TejasQ: Can we separate imported files vs. packages with a line?

packages/components/src/utils/mixins.ts
peterszerzo: Hovering tooltips are a very common use-case (arguably the only way in which tooltips are used) and I'd like to encourage this mixin to be re-used as opposed to giving contributors an impression that this pattern needs to be re-implemented. It might make sense to move them into a better organizational structure - just opened https://github.com/contiamo/operational-ui/issues/578 for that, or together with tooltips as a commonly used utility.

PR #565, opened 2018-07-06
Add colorMapper util

micha-f:
A simple way to assign colors to strings deterministically without ending up with the same colors for similar names / names starting with the same initials. To be used with avatars and name tags.

### DIFF
--- a/packages/components/src/NameTag/NameTag.tsx
+++ b/packages/components/src/NameTag/NameTag.tsx
@@ -3,9 +3,11 @@ import styled from "react-emotion"
 import { readableTextColor } from "@operational/utils"
 
 import { OperationalStyleConstants, expandColor } from "../utils/constants"
+import { colorMapper } from "../utils/color"
 import { CardHeader, CardItem } from "../"
 
 export interface Props {
+  /** Background color */
   color?: string
   /**
    * Indicates that this component is left of other content, and adds an appropriate right margin.
@@ -15,7 +17,10 @@ export interface Props {
    * Indicates that this component is right of other content, and adds an appropriate left margin.
    */
   right?: boolean
-  children?: React.ReactNode
+  /**
+   * Children to this component are expected to be a plain string
+   */
+  children?: string
 }
 
 const Container = styled("div")(
@@ -24,13 +29,19 @@ const Container = styled("div")(
     color,
     left,
     right,
+    children,
+    assignColor,
   }: {
     theme?: OperationalStyleConstants
     color?: Props["color"]
     left?: Props["left"]
     right?: Props["right"]
+    children: string
+    assignColor: boolean
   }) => {
-    const backgroundColor = expandColor(theme, color) || theme.color.primary
+    const backgroundColor = assignColor
+      ? colorMapper(theme.color.palette)(children)
+      : expandColor(theme, color) || theme.color.primary
     const textColor = readableTextColor(backgroundColor, [theme.color.white, theme.color.black])
     return {
       backgroundColor,
@@ -49,6 +60,14 @@ const Container = styled("div")(
   },
 )
 
-const NameTag: React.SFC<Props> = props => <Container {...props}>{props.children}</Container>
+const NameTag: React.SFC<Props> = props => (
+  <Container {...props} assignColor={Boolean(!props.color)}>
+    {props.children}
+  </Container>
+)
+
+NameTag.defaultProps = {
+  children: "",
+}
 
 export default NameTag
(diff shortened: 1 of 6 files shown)


TejasQ:
How does it work?

micha-f:
It uses a hashing function (FNV 1a - https://en.wikipedia.org/wiki/Fowler%E2%80%93Noll%E2%80%93Vo_hash_function) to turn a UTF8 string into a positive int in the 32-bit space (0 to 2147483647*2). It then assigns this number to one of `n` colors using `mod n`. Small changes in the input string will result in a different hash (int) and thus (most likely) in a different color.

TejasQ:
And there's no simpler/better solution to do this?

micha-f:
If you want to 
- assign strings deterministically to the same color
- assign a different color even for very similar strings (those that only differ in one character)
then you need to somehow turn a string into a hash with a uniform distribution and map that to a color.

I don't think there is a simpler / better solution - otherwise I would have proposed it 😉 .

peterszerzo:
@micha-f I like the idea of doing this properly. If you don't mind, I'll fix up the branch so tests are passing myself, and see how we integrate it to the rest of the codebase.

packages/components/src/utils/color.ts
```diff
@@ -1,2 +1,84 @@
...
+/*
+ * Hashing algorithm used to pair strings with colors (see `colorMapper` above), chosen
+ * as a compromise between speed and low risk of having identical colors assigned for two
+ * strings that are nearly identical.
+ *
+ * Copied from https://github.com/tjwebb/fnv-plus (MIT licensed), author Travis Webb <me@traviswebb.com>
+ * This FNV-1a hash algorithm, often simply called "fnv", disperses hashes throughout
+ * the 32-bit hash space with very good dispersion and is very fast.
+ */
+function hash32FNV1aUTF(str: string) {
```
TejasQ: Maybe we can `export const hash = ` this.

micha-f: Do we need this in other places? If so - wouldn't it make more sense to move it out of the **color** utils? If we don't have another use case, does it make sense to keep it private as an implementation detail for now?

TejasQ: Right right. My mistake. I thought this was in a different file than `color.ts`. This looks good.

packages/components/src/utils/__tests__/color.test.tsx
```diff
@@ -0,0 +1,21 @@
+import * as React from "react"
```
fabien0102: Why do you need react?

TejasQ:
Sweet. A rebase and let's merge.

packages/components/src/utils/color.ts
```diff
@@ -1,2 +1,84 @@
...
+  let v0 = 0x9dc5
+  let t1 = 0
+  let v1 = 0x811c
+
+  for (i = 0; i < l; i = i + 1) {
+    c = str.charCodeAt(i)
+    if (c < 128) {
+      v0 ^= c
+    } else if (c < 2048) {
+      v0 ^= (c >> 6) | 192
```
fabien0102: So nice 😍 I like binary operators in js 💯

packages/components/src/utils/__tests__/color.test.tsx
micha-f: Can be removed

fabien0102:
It look very nice but where is it used?

TejasQ:
LGTM

fabien0102:
@peterszerzo can you rebase to fix the netlify build?

packages/components/src/NameTag/NameTag.tsx
```diff
@@ -15,7 +17,10 @@ export interface Props {
    * Indicates that this component is right of other content, and adds an appropriate left margin.
    */
   right?: boolean
-  children?: React.ReactNode
+  /**
+   * Children to this component are expected to be a plain string
+   */
+  children?: string
```
fabien0102: Even if is a plain string, I prefer `React.ReactNode` to be able to use `react-intl` easily.

```jsx
import { FormattedMessage } from "react-intl"

export default = () =>
   <NameTag><FormattedMessage id="myTag" defaultMessage="my tag" /></NameTag>
```

versus

```jsx
import { injectIntl, InjectedIntlProps } from "react-intl"

const MyTag: React.SFC<InjectedIntlProps> = ({intl}) =>
  <NameTag>{intl.formatMessage({id: "myTag", defaultMessage: "my tag"})}</NameTag>

export default injectIntl(MyTag)
```

fabien0102: For a NameTag is not really important, but it's more as a general rule :wink: (I have play a bit with `react-intl` and `operational`, and it's a bit annoying to have to inject the `intl` everytime

peterszerzo: I think it is important to communicate that only strings are allowed here.

It is likely that several components will have to get refactored if we wind up supporting internationalization in this codebase, but luckily those will be pretty low-risk.

Issue #561, opened 2018-07-06
Project selector and user menu


ImogenF:
## Project Selector and User Menu
Project selector: https://trello.com/c/hMNW8dop
User menu: https://trello.com/c/VSs4XyMC

There are two possible APIs.

### Proposed API 1
The first option is to make 2 simple components, a `ProjectSelector` and a `UserMenu` (naming open for discussion). These are just  the clickable parts of the header bar, which trigger context menus. This has the advantage of allowing each option in the menus to have its own `onClick` method, since, unlike in a `Select` dropdown component, the actions might be completely different. For example, one might open a pop-up, another might navigate to a new screen, etc. Since this is only really necessary for the user menu, we could enable an `onClick` prop within the `ContextMenu` as a shortcut which passes the same method on to all `ContextMenuItem`s.

```
<ContextMenu>
  <ProjectSelector>Click here</ProjectSelector>
  <ContextMenuItem onClick={() => alert("clicked")}>Menu 1</ContextMenuItem>
  <ContextMenuItem onClick={() => alert("clicked")}>Menu 2</ContextMenuItem>
  <ContextMenuItem onClick={() => alert("clicked")}>Menu 3</ContextMenuItem>
</ContextMenu>
```

 OR:

```
<ContextMenu onClick={() => alert(“clicked”)}>
  <HeaderMenu>Click here</HeaderMenu>
  <ContextMenuItem>Menu 1</ContextMenuItem>
  <ContextMenuItem>Menu 2</ContextMenuItem>
  <ContextMenuItem>Menu 3</ContextMenuItem>
</ContextMenu>
```

### Proposed API 2
The other alternative is to create 2 more complex components which use `ContextMenu` internally. This would simplify things for the end user.
```
const projectOptions = [
  { key: “project1”, label: “Project 1” },
  { key: “project2”, label: “Project 2” },
  { key: “project3”, label: “Project 3” },
]
<ProjectSelector options={projectOptions} active={this.state.activeProject}></ProjectSelector>

const userOptions = [
  { label: “My account”, onClick: () => {…} },
  { label: “Logout”, onClick: () => {…} },
]
<UserMenu options={userOptions}></UserMenu>
```

peterszerzo:
[...]

PR #566, opened 2018-07-06
HeaderMenu

ImogenF:
This is a wrapper for rendering a context menu dropdown within a header bar or page title bar. 
Unlike the base `ContextMenu` component, dropdown options can be passed in as an array and are rendered as `ContextMenuItem`s within the `HeaderMenu` component. Each option can have its own `onClick` method, or a single `onClick` method can be passed as a prop to `HeaderMenu`.

This component adds the following additional features to the context menu:
- Element that is clicked to open the menu has set width and height and an enhanced appearance on hover and when the menu is open.
- The user can optionally add a downward carat to the clickable element, via props.
- The menu can be right-aligned.

**Issue:** #561 

**Trello tickets:**
<blockquote class="trello-card"><a href="https://trello.com/c/hMNW8dop/235-project-selector">Project selector</a></blockquote>

<blockquote class="trello-card"><a href="https://trello.com/c/VSs4XyMC/236-user-menu">User menu</a></blockquote>


packages/components/src/ContextMenu/ContextMenu.tsx
```diff
@@ -10,17 +10,24 @@ export interface Props {
...
-  /** Handles click events anywhere outside the context menu container, including menu items. */
 
+  /** Handles click events anywhere outside the context menu container, including menu items. */
   onOutsideClick?: () => void
-  /** Suppresses the default behavior of closing the context menu when one of its items is clicked. */
 
+  /** Suppresses the default behavior of closing the context menu when one of its items is clicked. */
   keepOpenOnItemClick?: boolean
+
+  noOffset?: boolean
```
peterszerzo: Will we ever have an offset/isn't the no offset mode a better default?

ImogenF: This is being refactored and for now I've taken out the offset, but the idea behind the original offset is so that the dropdown menu is not pressed up right against the item that is clicked, if it doesn't have any padding (if it's just a text string, for instance).

packages/components/src/HeaderMenu/HeaderMenu.tsx
```diff
@@ -0,0 +1,96 @@
...
+    align: "left" | "right"
+    isOpen: boolean
+    withCarat: boolean
+  }) => ({
+    width: 250,
+    lineHeight: "50px",
+    padding: `0 ${theme.space.content}px`,
+    color: isOpen ? theme.color.white : "#ffffffcc",
+    backgroundColor: isOpen ? "hsla(0, 0%, 100%, 0.1)" : "transparent",
+    boxShadow: isOpen ? "0 3px 6px rgba(0, 0%, 0%, 0.3)" : "none",
```
fabien0102: We can maybe add this to the constants (at least extract this value in a `const` to avoid the L51 duplication ;) )

packages/components/src/HeaderMenu/HeaderMenu.tsx
```diff
@@ -0,0 +1,99 @@
...
+  children: React.ReactNode[]
+
+  /** Action when item in dropdown is selected - if specified here, it is applied to all dropdown items */
+  onClick?: any
+
+  /** Items to display in dropdown */
+  items?: any[]
+
+  /** Display carat on opposite side to align prop */
+  carat?: boolean
```
peterszerzo: Most sources I'm looking at spell this sign `caret` - does my research hold? Also, I'd call this `withCaret` to have it read a bit more clearly when used in client code.

packages/components/src/HeaderMenu/HeaderMenu.tsx
```diff
@@ -0,0 +1,99 @@
+import * as React from "react"
+import styled from "react-emotion"
+import { OperationalStyleConstants } from "../utils/constants"
+import { ContextMenu } from "../"
+
+export interface Props {
+  /** Options to display in dropdown */
+  children: React.ReactNode[]
+
```
peterszerzo: Removing these newlines would help tell apart the moving parts (props -> styled components -> entry point) parts of this file when reading it at a glance. They were an artifact from a codemod transformation which we didn't remove across the components yet.

packages/components/src/ContextMenu/ContextMenu.tsx
```diff
@@ -3,53 +3,72 @@ import styled from "react-emotion"
...
-    ...(isExpanded
-      ? {
-          display: "block",
-          animation: `${fadeIn} ease-in-out forwards 0.2s`,
-        }
-      : {
-          display: "none",
-        }),
+    display: isExpanded ? "block" : "none",
+    animation: `${fadeIn} ease-in-out forwards 0.2s`,
```
peterszerzo: I find the vertical movement in this animation looking a bit strange, intersecting the child and then settling right underneath it. Can we make a version of `fadeIn` that doesn't have the vertical movement?

packages/components/src/HeaderMenu/HeaderMenu.tsx
ImogenF: Yup, good catch.

packages/components/src/ContextMenu/ContextMenu.tsx
```diff
@@ -70,11 +88,14 @@ class ContextMenu extends React.Component<Props, State> {
       this.props.onOutsideClick()
     }
 
-    if (isTargetInsideContainer && this.props.onClick) {
-      this.props.onClick()
-    }
+    const newIsActive = isTargetInsideMenu
```
fabien0102: It's quite hard to read ^^ and it should be refactorable (all else cases are `false` so you can use the boolean logically)

packages/components/src/Breadcrumbs/README.md
```diff
@@ -23,16 +23,13 @@ Using a `to` prop navigates automatically, and render proper anchor tags with hr
...
 
 ### Usage with context menu
-```jsx
-const Breadcrumb = require("../Breadcrumb/Breadcrumb").default;
-const ContextMenuItem =  require("../ContextMenuItem/ContextMenuItem").default;
 
-<Breadcrumbs>
-  <ContextMenu css={{ display: "inline-block", margin: 0 }}>
+```jsx
+const Breadcrumb = require("../Breadcrumb/Breadcrumb").default
```
TejasQ: This is unneeded.

packages/components/src/ContextMenu/ContextMenu.Item.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+const Container = styled("div")(
+  ({
+    theme,
+    clickable,
+    condensed,
+  }: {
+    theme?: OperationalStyleConstants
+    clickable: boolean
+    condensed: boolean
```
TejasQ: How about `Props["condensed"]`?

packages/components/src/ContextMenu/ContextMenu.tsx
TejasQ: Or maybe not _as much_ vertical movement?

packages/components/src/ContextMenu/ContextMenu.tsx
```diff
@@ -89,50 +97,35 @@ class ContextMenu extends React.Component<Props, State> {
...
-        {children}
+        {typeof this.props.children === "function" ? this.props.children(this.state.isOpen) : this.props.children}
         <MenuContainer
           innerRef={node => {
             this.menuContainerNode = node
           }}
           isExpanded={this.props.open || this.state.isOpen}
         >
-          {menuItems}
+          {(this.props.items || []).map((item: any, index: number) => {
```
TejasQ: So defensive 😂 

Can we check do

```jsx
{this.props.items && this.props.items.map
```

instead? That we we don't map over an empty array for nothing.

Perhaps we can also type `item` better.

packages/components/src/HeaderMenu/HeaderMenu.tsx
```diff
@@ -0,0 +1,95 @@
+import * as React from "react"
+import styled from "react-emotion"
+import { OperationalStyleConstants } from "../utils/constants"
+import { ContextMenu } from "../"
+
+export interface Props {
+  /** Options to display in dropdown */
```
TejasQ: The options go in `children`? 🤔

packages/components/src/HeaderMenu/HeaderMenu.tsx
```diff
@@ -0,0 +1,95 @@
...
+  children: React.ReactNode[]
+  /** Action when item in dropdown is selected - if specified here, it is applied to all dropdown items */
+  onClick?: any
+  /** Items to display in dropdown */
```
TejasQ: How is this different from line 7?

packages/components/src/HeaderMenu/HeaderMenu.tsx
```diff
@@ -0,0 +1,95 @@
...
+  items?: any[]
```
TejasQ: Should we allow `any` in here?

packages/components/src/HeaderMenu/HeaderMenu.tsx
```diff
@@ -0,0 +1,95 @@
...
+  /** Display caret on opposite side to align prop */
+  withCaret?: boolean
```
TejasQ: I think we can just call this `caret`.

packages/components/src/HeaderMenu/README.md
```diff
@@ -0,0 +1,49 @@
+Menus that drop down from the header bar, i.e. project selectors, and user menus.
+
+### Usage
+
+```jsx
+const { ContextMenuItem } = require("../")
```
TejasQ: Not sure this is needed :think

packages/components/src/HeaderMenu/HeaderMenu.tsx
peterszerzo: I just made @ImogenF change it :) my feeling was that `<HeaderManu caret />` reads confusing (what about a caret?), whereas in the verbose version it's clearer that it is an addition to the element that is rendered in addition to the existing UI.

packages/components/src/HeaderMenu/README.md
peterszerzo: Definitely not, this can just be deleted.

packages/components/src/ContextMenu/ContextMenu.tsx
peterszerzo: What bothers me visually is that the white menu box overlaps/intersects the headermenu (which is on a dark background), so the interface looks a bit broken until it falls back into its place. Any amount of vertical movement will still retain that effect.

packages/components/src/HeaderMenu/HeaderMenu.tsx
TejasQ: Sure.

packages/components/src/ContextMenu/ContextMenu.tsx
peterszerzo: Shouldn't there always be an `items` array, and therefore the user is forced to add an empty array if they don't have choices? I feel like not having choices here is not ideal from a UX perspective, and if that's the case, it should be obvious (it is easier to read that `<HeaderMenu options={[]} withCaret />` is misleading to the user).

packages/components/src/ContextMenu/ContextMenu.tsx
TejasQ: I agree. I actually think it looks better without any animation at all.

ImogenF: I've taken out the vertical movement so now the `fadeIn` animation is simply opacity. This does, however, also impact other components (`Select`, `Overlay`, `DatePicker`) although I think they look good still. 

I could completely remove the animation from the `ContextMenu` entirely but I think it's nice to have it consistent with the other dropdowns.

packages/components/src/ContextMenu/ContextMenu.tsx
ImogenF: Have refactored this.

packages/components/src/ContextMenu/ContextMenu.tsx
fabien0102: You can also define `items` as `[]` in `defaultProps` for this kind of case (it should works if items is not `null`)

ImogenF: I'm inclined to agree with @peterszerzo on this one, and make `props.items` non-optional for both `ContextMenu` and `HeaderMenu`

packages/components/src/ContextMenu/ContextMenu.tsx
peterszerzo: Can we have both, something like `fadeIn` and `floatIn`? The `Confirm` component does a vertical movement, and it'd be nice to keep it around in a few other places.

ImogenF: So:
- rename the old `fadeIn` as `floatIn`
- add new `fadeIn` and use only for `ContextMenu`, `Select` and `DatePicker`?

peterszerzo: `Select` and `DatePicker` can still float in I'd say.

packages/components/src/ContextMenu/ContextMenu.tsx
TejasQ: You'll still have a complete dismount of your component and an ugly error in a non-typescript environment with that approach.

I suggest an extra layer of feedback to our users in addition to making it non-optional: at the very top of `render()`, check if there's an `items` array, and if not, throw a descriptive error.

ImogenF:
@fabien0102 @peterszerzo @TejasQ I've addressed all of the comments as far as I can see. Please let me know if I've missed anything!

packages/components/src/ContextMenu/ContextMenu.tsx
```diff
@@ -89,50 +102,40 @@ class ContextMenu extends React.Component<Props, State> {
...
-                  onClick()
-                }),
-            }),
-          )
-        } else {
-          children.push(child)
-        }
-      },
-    )
+    {
```
TejasQ: This block is unnecessary here. 🤔

packages/components/src/ContextMenu/ContextMenu.tsx
```diff
@@ -89,50 +102,40 @@ class ContextMenu extends React.Component<Props, State> {
...
+        {typeof this.props.children === "function" ? this.props.children(this.state.isOpen) : this.props.children}
         <MenuContainer
           innerRef={node => {
             this.menuContainerNode = node
           }}
           isExpanded={this.props.open || this.state.isOpen}
         >
-          {menuItems}
+          {this.props.items.map((item: string | Item, index: number) => {
+            const clickHandler = (typeof item !== "string" && item.onClick) || this.props.onClick
```
TejasQ: Beautiful. 🔥

packages/components/src/HeaderMenu/HeaderMenu.tsx
```diff
@@ -0,0 +1,105 @@
...
+  /** Clickable component(s) from which menu appears  */
+  children: React.ReactNode[]
+  /** Action when item in dropdown is selected - if specified here, it is applied to all dropdown items */
+  onClick?: (item?: string | Item) => void
+  /** Items to display in dropdown */
+  items: (string | Item)[]
+  /** Display caret on opposite side to align prop */
+  withCaret?: boolean
+  /** Alignment */
+  align: "left" | "right"
```
TejasQ: Can this be optional and fallback to a default?

packages/components/src/HeaderMenu/HeaderMenu.tsx
```diff
@@ -0,0 +1,105 @@
...
+            },
+          },
+        }
+      : {}),
+  }),
+)
+
+const HeaderMenu: React.SFC<Props> = (props: Props) => {
+  {
+    if (!props.items) {
```
TejasQ: ? Doesn't `ContextMenu` already do this?

packages/components/src/ContextMenu/ContextMenu.Item.tsx
```diff
@@ -0,0 +1,62 @@
+import * as React from "react"
+import styled from "react-emotion"
+import { OperationalStyleConstants } from "../utils/constants"
+import { darken } from "@operational/utils"
+import { WithTheme, Css, CssStatic } from "../types"
+
+export interface Props {
+  id?: string
+  className?: string
+  children?: any
```
TejasQ: `id`, `className`, `children`, `onClick` are all part of `React.SFC<{}>` on line 50. They can be removed from here.

packages/components/src/ContextMenu/ContextMenu.Item.tsx
```diff
@@ -0,0 +1,62 @@
...
+    "&:not(:first-child)": {
+      borderTop: "1px solid",
+      borderColor: theme.color.separators.default,
+    },
+  }),
+)
+
+const ContextMenuItem: React.SFC<Props> = (props: Props) => (
+  <Container
+    id={props.id}
```
TejasQ: `id`, `className`, `children`, `onClick` are all part of `React.SFC<{}>` and can be spread in here:

```jsx
<Container
  {...props}
```

packages/components/src/HeaderMenu/HeaderMenu.tsx
```diff
@@ -0,0 +1,100 @@
...
+import { ContextMenu } from "../"
+
+export interface Item {
+  label: string
+  onClick?: (item: string | Item) => void
+}
+
+export interface Props {
+  /** Clickable component(s) from which menu appears  */
+  children: React.ReactNode[]
```
TejasQ: This is inferred from `React.SFC`

packages/components/src/HeaderMenu/HeaderMenu.tsx
```diff
@@ -0,0 +1,100 @@
+import * as React from "react"
+import styled from "react-emotion"
+import { OperationalStyleConstants } from "../utils/constants"
+import { ContextMenu } from "../"
+
+export interface Item {
```
TejasQ: Instead of defining this twice here, you could import the props from `ContextMenu` or `ContextMenuItem` and annotate `onClick` here with `ContextMenuItem["onClick"]` since it's the same behavior.

packages/components/src/ContextMenu/ContextMenu.Item.tsx
ImogenF: Removing them throws up TS errors: "Property 'className' does not exist on type 'Props'."

peterszerzo: I believe that error goes away after you spread, per comment below.

ImogenF: Even with that: ```TS2322: Type '(props: Props) => Element' is not assignable to type 'StatelessComponent<{}>'.
  Types of parameters 'props' and 'props' are incompatible.
    Type '{ children?: ReactNode; }' has no properties in common with type 'Props'.```

TejasQ: Hmmm, let's see... *inspects*

TejasQ: You need to remove the `Props` annotation from the argument.

ImogenF:
@TejasQ Have made the changes you requested and rebased - can you do a final check please?

packages/components/src/ContextMenu/ContextMenu.Item.tsx
```diff
@@ -0,0 +1,52 @@
...
+const Container = styled("div")(
+  ({
+    theme,
+    onClick,
+    condensed,
+  }: {
+      theme?: OperationalStyleConstants
+      onClick?: () => void
+      condensed: Props["condensed"]
+    }): any => ({
```
TejasQ: We can remove this type annotation.

packages/components/src/HeaderMenu/HeaderMenu.tsx
```diff
@@ -0,0 +1,96 @@
+import * as React from "react"
+import styled from "react-emotion"
+import { OperationalStyleConstants } from "../utils/constants"
+import { ContextMenu } from "../"
```
TejasQ: This is going to break the styleguide on netlify. Consider importing `ContextMenu` from its file instead on the next line.

ImogenF:
@TejasQ I was going for consistency with the screenshots in Trello, with the separator lines not going fully across. I can make the hover background go all the way to the edges, but then the separators would also have to extend fully.

micha-f:
In the screen the space between separator and edge was 20px (more than implemented currently). I forgot about the background on hover though which might look weird if it starts only at the first pixel of the text content.

@ImogenF Can you check how it looks with 20px spacing and no padding left/right for the text? If you don't like how it looks, let's go with separators that extend fully.

micha-f:
Can we make the menu appear faster? Or try no fade in animation at all? Fade out is fine. Right now opening it feels laggy. Lucas pointed this out and I agree.

ImogenF:
@TejasQ @micha-f sorry, I must have just missed these comments on Friday. 

20px spacing and no text padding looks strange, so I'll extend the separators.

image

And now that you mention it, it does feel laggy. Will sort that out.

PR #569, opened 2018-07-09
Fix transform origin. Add left/right props

peterszerzo:
* fixes a bug where the spinner's transform origin is messed up when used within the button.
* adds the missing left/right prop support for spinners.

spinnerfix

### DIFF
--- a/packages/components/src/Spinner/Spinner.tsx
+++ b/packages/components/src/Spinner/Spinner.tsx
@@ -71,8 +71,8 @@ const Container = styled("div")(
     width: size || defaultSize,
     height: size || defaultSize,
     color: expandColor(theme, color) || "currentColor",
-    ...(left ? { marginRight: theme.space.small } : {}),
-    ...(right ? { marginLeft: theme.space.small } : {}),
+    marginRight: left ? theme.space.small : 0,
+    marginLeft: right ? theme.space.small : 0,
     "& svg": {
       fill: "currentColor",
     },
@@ -111,8 +111,10 @@ const BouncingSpinnerContainer = styled("div")({
 })
 
 /**
- * The bouncing spinner is constructed out of 3 80x80 boxes spaced out horizontally on a 360x360 grid (these units don't refer to pixels, they simply mimic the grid of the icons.
- * The math used in here lays these boxes out so they're vertically centered and spaced equally on the horizontal axis without any gutter.
+ * The bouncing spinner is constructed out of 3 80x80 boxes spaced out horizontally on a 360x360 grid
+ * (these units don't refer to pixels, they simply mimic the grid of the icons.
+ * The math used in here lays these boxes out so they're vertically centered and spaced
+ * equally on the horizontal axis without any gutter.
  */
 const BouncingSpinnerBox = styled("div")(({ no }: { no: number }) => ({
   width: `${(80 / 360) * 100}%`,
--- a/packages/components/src/Spinner/Spinner.tsx
+++ b/packages/components/src/Spinner/Spinner.tsx
@@ -80,7 +80,8 @@ const Container = styled("div")(
 )
 
 /**
- * This additional container is introduced to make transforms set on the main container from the outside (e.g. `styled` helper) do not mess up the rotation origin.
+ * This additional container is introduced to make transforms set on the main container from the outside
+ * (e.g. `styled` helper) do not mess up the rotation origin.
  */
 const AnimationContainer = styled("div")(({ bounce, size }: { bounce?: boolean; size?: number }) => ({
   margin: 0,
@@ -88,7 +89,11 @@ const AnimationContainer = styled("div")(({ bounce, size }: { bounce?: boolean; 
   width: size || defaultSize,
   height: size || defaultSize,
   transformOrigin: "center center",
-  ...(bounce ? {} : { animation: `${spinKeyframes} 1.5s infinite linear` }),
+  /*
+   * When the bounce animation is used, animation properties are set on the individual bouncing squares,
+   * therefore no animation is set on the container.
+   */
+  animation: bounce ? "none" : `${spinKeyframes} 1.5s infinite linear`,
 }))
 
 const RegularSpinner = () => (
--- a/packages/components/src/Spinner/Spinner.tsx
+++ b/packages/components/src/Spinner/Spinner.tsx
@@ -15,14 +15,24 @@ export interface Props {
   size?: number
   /** Renders a bouncing animation as opposed to a regular spinning one. */
   bounce?: boolean
+  /**
+   * Indicates that this component is left of other content, and adds an appropriate right margin.
+   */
+  left?: boolean
+  /**
+   * Indicates that this component is right of other content, and adds an appropriate left margin.
+   */
+  right?: boolean
 }
 
 const spinKeyframes = keyframes({
   "0%": {
     transform: "rotate(0deg)",
+    transformOrigin: "center center",
   },
   "100%": {
     transform: "rotate(360deg)",
+    transformOrigin: "center center",
   },
 })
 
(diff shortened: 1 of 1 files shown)


packages/components/src/Spinner/Spinner.tsx
```diff
@@ -15,14 +15,24 @@ export interface Props {
...
+  /**
+   * Indicates that this component is right of other content, and adds an appropriate left margin.
+   */
+  right?: boolean
 }
 
 const spinKeyframes = keyframes({
   "0%": {
     transform: "rotate(0deg)",
+    transformOrigin: "center center",
```
TejasQ: We're animating to and from the same value... why not just put this in the style definition?

packages/components/src/Spinner/Spinner.tsx
```diff
@@ -49,24 +59,40 @@ const Container = styled("div")(
...
+    right,
   }: {
     size?: number
     color?: string
     bounce?: boolean
     theme?: OperationalStyleConstants
+    left?: boolean
+    right?: boolean
   }) => ({
     label: "spinner",
```
TejasQ: We don't need this anymore.

packages/components/src/Spinner/Spinner.tsx
```diff
@@ -49,24 +59,40 @@ const Container = styled("div")(
...
     display: "inline-block",
     width: size || defaultSize,
     height: size || defaultSize,
-    ...(bounce ? {} : { animation: `${spinKeyframes} 1.5s infinite linear` }),
     color: expandColor(theme, color) || "currentColor",
+    ...(left ? { marginRight: theme.space.small } : {}),
```
TejasQ: Wouldn't `marginRight: left ? theme.space.small : 0` be easier to read? @fabien0102

packages/components/src/Spinner/Spinner.tsx
```diff
@@ -49,24 +59,40 @@ const Container = styled("div")(
...
 )
 
+/**
+ * This additional container is introduced to make transforms set on the main container from the outside (e.g. `styled` helper) do not mess up the rotation origin.
+ */
+const AnimationContainer = styled("div")(({ bounce, size }: { bounce?: boolean; size?: number }) => ({
+  margin: 0,
+  lineHeight: 0,
+  width: size || defaultSize,
+  height: size || defaultSize,
```
TejasQ: How about we put the `transform-origin` here?

packages/components/src/Spinner/Spinner.tsx
peterszerzo: You're right, it is, though I think `undefined` is clearer to follow in a sense that you either set something explicitly or not change the existing styles at all.

packages/components/src/Spinner/Spinner.tsx
peterszerzo: Absolutely - it was a debugging trial-and-error that I never cleaned up. Thanks for the catch :)

packages/components/src/Spinner/Spinner.tsx
peterszerzo: Though even then, doesn't it feel a bit off that we're introducing defaults that may cause style regressions just to express the fact that we don't want to set the margin right after all? Or add the `undefined` and have to keep in mind `emotion`'s implementation on whether it will clear out other styles or not?

TejasQ: I don’t think a zero margin is a default but an _opinion_ in our opinionated library.

peterszerzo: Wouldn't you run into unexpected behavior if you just assumed margins were auto but found that they weren't?

packages/components/src/Spinner/Spinner.tsx
peterszerzo: This is addressed in `b86f0a9`

packages/components/src/Spinner/Spinner.tsx
```diff
@@ -49,24 +57,40 @@ const Container = styled("div")(
...
+    ...(left ? { marginRight: theme.space.small } : {}),
+    ...(right ? { marginLeft: theme.space.small } : {}),
     "& svg": {
       fill: "currentColor",
     },
   }),
 )
 
+/**
+ * This additional container is introduced to make transforms set on the main container from the outside (e.g. `styled` helper) do not mess up the rotation origin.
```
TejasQ: Can we split this comment into 2 lines? A horizontal scrollbar [here](https://github.com/contiamo/operational-ui/blob/63032537eb54b6cc52b05c735813de90f80a3843/packages/components/src/Spinner/Spinner.tsx) typically is a sign that a comment is too long.

Same comment for [these](https://github.com/contiamo/operational-ui/blob/63032537eb54b6cc52b05c735813de90f80a3843/packages/components/src/Spinner/Spinner.tsx#L113-L116).

packages/components/src/Spinner/Spinner.tsx
```diff
@@ -49,24 +57,40 @@ const Container = styled("div")(
...
+ */
+const AnimationContainer = styled("div")(({ bounce, size }: { bounce?: boolean; size?: number }) => ({
```
TejasQ: The type annotation can be `Pick<Props, "bounce" | "size">` for more safety. Same comment for [Container](https://github.com/contiamo/operational-ui/blob/63032537eb54b6cc52b05c735813de90f80a3843/packages/components/src/Spinner/Spinner.tsx#L63-L68).

packages/components/src/Spinner/Spinner.tsx
TejasQ: What? I'm not sure I understand.

peterszerzo: What I mean is that we are already used to the default behavior of `margin: auto` as it is applied to native DOM elements in general. Why set it to zero when it doesn't change the visual behavior here but may confuse people in other cases?

Or to take a step back: doesn't `margin: 0` mean that the hint is right next to its siblings? Is that semantics that describes the design effectively?

packages/components/src/Spinner/Spinner.tsx
peterszerzo: Can we do this in a different branch after we set a standard for comments? I find that multiline comments involve a lot of guesswork unless we all configure our editors to do some automated magic.

packages/components/src/Spinner/Spinner.tsx
TejasQ: I think as an _opinionated_ UI library, we ought to have solid values as opinions. I'd argue that the default behavior, even if we're used to it, is not an opinion and _"loosens"_ Operational's opinionation.

packages/components/src/Spinner/Spinner.tsx
TejasQ: ? set a standard for comments?

I'm not sure we need a standard... if it's too long, break it into multiple lines?

packages/components/src/Spinner/Spinner.tsx
```diff
@@ -49,24 +57,40 @@ const Container = styled("div")(
...
+  margin: 0,
+  lineHeight: 0,
+  width: size || defaultSize,
+  height: size || defaultSize,
+  transformOrigin: "center center",
+  ...(bounce ? {} : { animation: `${spinKeyframes} 1.5s infinite linear` }),
```
TejasQ: Can we use a ternary here?

packages/components/src/Spinner/Spinner.tsx
TejasQ: ? But they _are_ coupled because the props are passed to `AnimationContainer` from its parent, no?

peterszerzo: I wouldn't say they're coupled, they just happen to have similar fields in the interface. This may not be true in the future, nor should the reader of this file think that there is some kind of relation to the exported component's props and the arguments of a self-contained styled component.

PR #572, opened 2018-07-09
Suggestions for HeaderMenu

peterszerzo:
* context menu now accepts function as a child to which it can pass its open state. This makes it possible to style header menu without classnames (90% of implementation is in there, to be fine-tuned shortly).
* make `ContextMenuItem` a private component, and packed it inside as `ContextMenu/ContextMenu.Item.tsx`.
* move tests inside context menu folder (cosmetics..).


TejasQ:
I don't see why this is needed.

peterszerzo:
@TejasQ see the last code example in https://deploy-preview-572--operational-ui.netlify.com/#contextmenu. The point is that context menu exposes its open state to its child so that said child can be styled differently when the menu is open or it is not.

Can you elaborate on why you think this isn't needed, or what alternatives you have in mind for our open state-based styling needs?

peterszerzo:
To summarize an offline discussion on this: we decided that we merge this function-as-a-child functionality around, but it might make more sense to make the component controlled. This will be handled in a different PR.

Issue #575, opened 2018-07-10
Confirm component


fabien0102:
### Mockup
![image](https://user-images.githubusercontent.com/1761469/42502434-4cf38928-8436-11e8-91ed-91ca7567a09f.png)

### Challenges

* The first challenge is the position of the modal, to be on the center of the `PageArea`, the component should be a direct child of the `PageArea`

* Deal with multiple confirms. Indeed our first use case is a page with 2 differents `Confirm` so ideally I want to provide the best readibility for this and deal with the less state as possible

### API Proposition
 I tried some patterns on my side, the best one at the moment is this one:

```jsx
<Page title="Delete the internet">
  <Confirm>
    {confirm => (
      <Card title="The red button">
        <Button
          color="error"
          onClick={() =>
            confirm({
              title: "Are you sure?",
              body: (
                <React.Fragment>
                  <p>This action will delete the entire internet</p>
                  <p>You can continue safely if you are a bad guy!</p>
                </React.Fragment>
              ),
              cancelButton: <Button color="success">I'm feel guilty :-/</Button>,
              actionButton: <Button color="error">I'm a bad guy!</Button>,
              onConfirm: () => alert("boooooommm!"),
              onCancel: () => alert("you are weak!"),
            })
          }
        >
          Delete the internet
        </Button>
      </Card>
    )}
  </Confirm>
</Page>
```

With this pattern, we have:
* `Confirm` is the direct child of `Page`, so the styling should be "easy".
* we can deal with multiple confirm in the same `Page`, and the content of the confirm is really near the `Button`.
* All states are inside the `Confirm` component, so we don't need to reimplement the state management everytime

What do you think about this pattern? Better ideas?

PR #577, opened 2018-07-11
Feat/confirm

fabien0102:
# Confirm component

d10922e1-6abc-4f5c-bce4-f740161c684e

### Doc link

https://deploy-preview-577--operational-ui.netlify.com/#!/Confirm

### Related issue

#575 

### To test

- [ ] The modal have the correct position (middle top of the PageArea)
- [ ] `onConfirm` is called if I confirm
- [ ] `onCancel` is called if I click on the cancel button
- [ ] `onCancel` is called if I click on the overlay
- [ ] Nothing explode if I don't provide any callbacks
- [ ] I can have multiple confirm action in one page without conflicts
- [ ] It's works on any layout (please challenge the component with Tabs, etc…)
- [ ] This component is amazing :tada:


packages/components/src/Confirm/Confirm.tsx
```diff
@@ -0,0 +1,116 @@
...
+  }
+
+  100% {
+    opacity: 1;
+  }
+`
+
+const fromBottom = keyframes`
+  0% {
+    top: -150%
```
peterszerzo: A bit too much movement here, shall we try -20px to make it a bit more subtle?

packages/components/src/Confirm/Confirm.tsx
```diff
@@ -0,0 +1,116 @@
...
+  }
+
+  100% {
+    top: 0
+  }
+`
+
+const Overlay = styled("div")`
```
peterszerzo: Let's look into moving these into `utils/mixins.ts` and re-using them for the `Modal` component. Might as well revive it a bit.

packages/components/src/Confirm/Confirm.tsx
```diff
@@ -0,0 +1,116 @@
...
+export interface ConfirmOptions {
+  title: React.ReactNode
+  body: React.ReactNode
+  cancelButton?: React.ReactElement<ButtonProps>
+  actionButton?: React.ReactElement<ButtonProps>
+  onConfirm?: () => void
+  onCancel?: () => void
+}
+
+export type ConfirmState = Readonly<Partial<ConfirmOptions> & { isOpen: boolean }>
```
peterszerzo: It seems to me that `isOpen` is redundant here. Don't we know the current confirm state from either having or not having a confirm object in state?

This can be `export type ConfirmState = ConfirmOptions | undefined` or something for now, and we can introduce a boolean flag if we ever implement fancy things like a smooth animated close.

packages/components/src/Confirm/Confirm.tsx
fabien0102: Good idea, but I will just isolate this component in `Overlay/Overlay.tsx` (it's a real component, he deserve a real folder ^^)

packages/components/src/Confirm/Confirm.tsx
```diff
@@ -0,0 +1,101 @@
...
+  title: React.ReactNode
+  body: React.ReactNode
+  cancelButton?: React.ReactElement<ButtonProps>
+  actionButton?: React.ReactElement<ButtonProps>
+  onConfirm?: () => void
+  onCancel?: () => void
+}
+
+export interface State {
+  options?: ConfirmOptions
```
peterszerzo: Using the word `options` here makes me think more of config/settings as opposed to the actual bread-and-butter content of a confirm window. The names `data`, `content` or `children` come to mind (with appropriate renames above `ConfirmOptions -> ConfirmData / ConfirmContent / ConfirmChildren`.

All four are imperfect semantically, but I wonder if any of these is worth considering.

PR #579, opened 2018-07-11
Replace webpack with file-by-file tsc build for components

peterszerzo:
For packaging `@operational/components`, this PR uses the `tsc` binary to compile components files one by one without packaging `react` etc. Tested locally through `npm link` on a dependent project (Pantheon UI).

I am suspecting that this will solve the bug in which Input components rendered by a package that use operational UI and are themselves packaged by webpack lose focus after a character is typed in.\

Note that webpack configuration and dependencies as kept for styleguidist builds.


TejasQ:
What benefits does this approach provide? Why should we move away from webpack?

peterszerzo:
It solves the multiple bundling issue in the simplest possible way. Dependencies are packaged file by file and only point to their dependencies through simple `require` calls instead of packaging them together into a single bundle. Having a single bundle will result in multiple instances of react as well as operational included in our production projects, leading to the input bug I described above. It also makes builds more difficult to track and optimize.

What are the benefits of webpack? Is it worth engineering the webpack-based solution to the issues mentioned above?

TejasQ:
I see the following benefits of having webpack responsible for packaging Operational UI:

- We need it around _anyway_ because of styleguidist.
- It can be environment aware. If people are building projects that interally use Operational UI, we can really optimize and remove dead code at the time they build their projects for `production`, while producing larger dev-friendly files with source maps on other environments.
- We can cleanly exclude peer dependencies by name and only expose our compiled code.

Webpack also exports file-by-file builds as well per component. We can totally do that as well.

TejasQ:
We could even run a different command based on `NODE_ENV`.

TejasQ:
image

Just got this in the IDP UI with `tsc` compilation 🙄

fabien0102:
I will stress test a bit this PR before approval (but maybe tomorrow if you don't mind)

fabien0102:
@peterszerzo  You need to remove `  "module": "./src/index.ts",` into package.json to have this really working ;)

Otherwise, I play a bit with this branch, it seams to work well (but I can't try the idp-ui focus bug without a npm version, link -> link -> link doesn't work for me…)

peterszerzo:
@fabien0102 I removed the module field.

PR #580, opened 2018-07-11
Export Confirm

fabien0102:
No comment…

I will add an integration test for this kind of thing… I think that is the 5th times that I done this stupid error…


fabien0102:
image

For the next time ^^

peterszerzo:
@fabien0102 what really helps here is to have a snapshot test that requires the component from the package's entry point as opposed to the hidden file. This is how most tests are set up currently.

packages/components/src/__tests__/libExports.ts
```diff
@@ -0,0 +1,16 @@
+import { readdirSync, readFileSync } from "fs"
+import { join } from "path"
+
+const BLACK_LIST = ["Internals", "utils", "__tests__", "types.ts", "index.ts"]
+
+describe("lib exports", () => {
```
peterszerzo: Beautiful!

PR #582, opened 2018-07-12
Add a pull request template

fabien0102:
The idea is to enforce good practices, with a basic checklist to avoid to break netlify again (ref #581 )

I'm open to any suggestions here, this PR is to open the discussion ;)

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -8,6 +8,14 @@
 
 ### To be tested
 
+Tester 1
+
 - [ ] No error/warning in the console
 - [ ] The netlify build is working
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,13 @@
...
+
+<!-- Some context about this PR: screenshots and links to the docs are appreciate -->
+
+### Related issue
+
+<!-- Paste the github issue here -->
+
+### To be tested
+
+- [ ] No error/warning in the console
```
peterszerzo: How about we add multiple copies of this with names so individual testers can add their names and check the ones they've verified themselves, something like this:
```
Tester 1

Tester 2
```

PR #583, opened 2018-07-12
Fix spinner margin regression in card header

peterszerzo:
Fixes the pivoting bug inside CardHeader:

spinner-in-card

In the current setup, `!important` styles are necessary to make sure the cardheader's svg descendant selector rule doesn't override the spinner's styles. Ideally, we'll have fewer loose descendant selectors in the future, making these overrides unnecessary.


TejasQ:
Closing because of an amicable offline discussion with Peter about how CardHeaders should handle icons.

PR #588, opened 2018-07-16
Add visual tests for adding series and updating palettes.

ImogenF:
### Summary

<blockquote class="trello-card"><a href="https://trello.com/c/t6RbmhKd/261-add-visual-tests-with-more-series">Add visual tests with more series</a></blockquote>

<blockquote class="trello-card"><a href="https://trello.com/c/CUF8xcHB/251-add-visual-tests-for-visualisation-colour-palettes">Add visual tests for visualisation colour palettes.</a></blockquote>

Visual tests added to check:

- How the default color palette looks when adding series one-by-one (for lines and bars).
- How all of the available color palettes look for a range of renderer types (lines, bars, symbols).

The palest colours have also been removed from some palettes to ensure enough contrast between the rendered data and the background.

### To be tested

Tejas
- [x] https://deploy-preview-588--operational-ui.netlify.com/visual-tests/#charts/lines-1
- [x] https://deploy-preview-588--operational-ui.netlify.com/visual-tests/#charts/lines-2
- [x] https://deploy-preview-588--operational-ui.netlify.com/visual-tests/#charts/bars-5
- [x] https://deploy-preview-588--operational-ui.netlify.com/visual-tests/#charts/bars-6
- [x] https://deploy-preview-588--operational-ui.netlify.com/visual-tests/#charts/symbol-5

Tester 2

- [x] https://deploy-preview-588--operational-ui.netlify.com/visual-tests/#charts/lines-1
- [x] https://deploy-preview-588--operational-ui.netlify.com/visual-tests/#charts/lines-2
- [x] https://deploy-preview-588--operational-ui.netlify.com/visual-tests/#charts/bars-5
- [x] https://deploy-preview-588--operational-ui.netlify.com/visual-tests/#charts/bars-6
- [x] https://deploy-preview-588--operational-ui.netlify.com/visual-tests/#charts/symbol-5


TejasQ:
@micha-f can you please check the visuals?

micha-f:
These look great!
I discussed with @ImogenF that we'll switch the default color palette, then merge.

TejasQ:
[Add visual tests for visualisation colour palettes.](https://trello.com/c/CUF8xcHB/251-add-visual-tests-for-visualisation-colour-palettes)

TejasQ:
[Add visual tests with more series](https://trello.com/c/t6RbmhKd/261-add-visual-tests-with-more-series)

packages/visualizations/src/Chart/facade.ts
```diff
@@ -149,15 +149,19 @@ class ChartFacade implements Facade {
...
+        true,
+      )
     }
     return this.state.config(config)
   }
 
-  accessors(type: string, accessors: Accessors<any>): Accessors<any> {
-    if (type === "series" && has("legendColor")(accessors)) {
+  accessors(type: string, accessors: Accessors<any>, suppressFlagModify: boolean = false): Accessors<any> {
+    if (type === "series" && has("legendColor")(accessors) && !suppressFlagModify) {
```
micha-f: What is the idea behind `supressFlagModify`? Can we add a comment?

packages/visualizations/src/Chart/facade.ts
```diff
@@ -147,21 +147,17 @@ class ChartFacade implements Facade {
   }
 
   config(config?: Partial<ChartConfig>): ChartConfig {
+    // Changing the palette config only updates the legendColor accessor if the default is still be used.
+    // It will not overwrite a user-defined legendColor accessor.
```
TejasQ: Consider leaving [multiline comments](http://processingjs.org/reference/multiline%20comment/) instead for multiline comments.

packages/visualizations/src/Chart/series_manager.ts
```diff
@@ -70,7 +71,7 @@ class ChartSeriesManager implements SeriesManager {
       this.assignBarIndices.bind(this),
       this.handleGroupedSeries("stacked", this.computeStack.bind(this)),
       this.handleGroupedSeries("range", this.computeRange.bind(this)),
-    )(this.state.current.get("accessors").data.series(this.state.current.get("data")))
+    )(cloneDeep(this.state.current.get("accessors").data.series(this.state.current.get("data"))))
```
micha-f: @ImogenF Why is this required? Does it have to do with the color palettes?

ImogenF: It's linked - otherwise the data seems to be overwritten in place, so if the palette is changed without new data being entered, the stacking algorithm ends up calculating the totals wrong - it adds the running totals to each other, rather than adding the raw values.

ImogenF: Ok @micha-f, the raw data saved in state is definitely being modified. This means that the stateHandler `readOnly` option is failing to prevent this. I think this is probably an issue to be dealt with in a separate PR.

micha-f: Does this change mean that we always deepclone all data you pass to a visualization?

micha-f: @ImogenF @TejasQ I believe this could be quite costly

ImogenF: Only for the charts (not all visualizations), but yes, that is what's currently happening. I agree it's not ideal, but as I say, I think we need to look into making the readonly `state` actually readonly

micha-f: @ImogenF  Once we make it readonly, computed stacks should actually not work anymore and would need another way to store their computations - correct?

ImogenF: Hmm. I suppose. Basically, whatever is passed in at L74 needs to be a fresh copy of the raw data each time `chart.draw()` is called

micha-f: Ok, let's keep this for now but let's open a ticket in trello right away for fix readonly and to think about stacks.

PR #591, opened 2018-07-16
Update status indicator colors.

ImogenF:
### Summary

<blockquote class="trello-card"><a href="https://trello.com/c/ZDktty09/264-status-indicator-tweak-grey-add-yellow-in-progress">Status indicator: tweak grey, add yellow (in progress)</a></blockquote>

### To be tested

Tester 1 (Peter)

Tester 2

### DIFF
--- a/packages/components/src/Status/Status.tsx
+++ b/packages/components/src/Status/Status.tsx
@@ -4,12 +4,17 @@ import { OperationalStyleConstants } from "../utils/constants"
 import tinycolor from "tinycolor2"
 
 export interface Props {
+  running?: boolean
   success?: boolean
   error?: boolean
   theme?: OperationalStyleConstants
 }
 
-const getColorFromProps = ({ success, error, theme }: Props): string => {
+const getColorFromProps = ({ running, success, error, theme }: Props): string => {
+  if (running) {
+    return theme.color.running
+  }
+
   if (success) {
     return theme.color.success
   }
@@ -18,7 +23,7 @@ const getColorFromProps = ({ success, error, theme }: Props): string => {
     return theme.color.error
   }
 
-  return theme.color.background.dark
+  return "#989898"
 }
 
 export const Status = styled("div")((props: Props) => ({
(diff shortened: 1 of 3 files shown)


ImogenF:
[Status indicator: tweak grey, add yellow (in progress)](https://trello.com/c/ZDktty09/264-status-indicator-tweak-grey-add-yellow-in-progress)

packages/components/src/Status/Status.tsx
```diff
@@ -18,7 +23,7 @@ const getColorFromProps = ({ success, error, theme }: Props): string => {
     return theme.color.error
   }
 
-  return theme.color.background.dark
+  return "#989898"
```
TejasQ: Why can't we use a constant here?

packages/components/src/utils/constants.ts
```diff
@@ -74,6 +74,7 @@ const borderColors = {
 const color = {
   primary: primaryColor,
   disabled: "#f5f5f5",
+  running: "#ffbd41",
```
TejasQ: Can we call this `warn` instead? I foresee it being used for that case more than to indicate a _running_ state.

packages/components/src/Status/Status.tsx
ImogenF: We could. My thinking was that there are so many greys in the constants used for so many things that it doesn't really make sense to keep adding to them every time we want a very particular grey. The closest constant we currently have is `theme.color.text.lightest`, as far as I can see, which doesn't really have anything to do with a neutral status. 

Should we maybe refactor the constants and have a list of greys, which `theme.color.text` etc can then make use of?

TejasQ: @micha-f would it make sense to use one of the many _already existing_ greys for this case?

micha-f: I think it'll be difficult to use a common set of greys for text, backgrounds, separators and shades. I'd say it's best to just use the specified grey here. I think design should go first. And this value looks good for a neutral status indicator.

packages/components/src/utils/constants.ts
```diff
@@ -74,6 +74,7 @@ const borderColors = {
 const color = {
   primary: primaryColor,
   disabled: "#f5f5f5",
+  warn: "#ffbd41",
```
peterszerzo: Can we call this `warning`?

PR #592, opened 2018-07-16
Update Button padding.

ImogenF:
### Summary

<blockquote class="trello-card"><a href="https://trello.com/c/BVsaoPw9/282-button-horizontal-padding">Button horizontal padding</a></blockquote>

### To be tested

Tester 1 (Tejas)

Tester 2 (Peter)

### DIFF
--- a/packages/components/src/utils/constants.ts
+++ b/packages/components/src/utils/constants.ts
@@ -146,6 +146,9 @@ const space = {
   /** Small space is `8px` */
   small: 8,
 
+  /** Medium space is `12px` */
+  medium: 12,
+
   /** Content space is `16px` */
   content: 16,
 
--- a/packages/components/src/Button/Button.tsx
+++ b/packages/components/src/Button/Button.tsx
@@ -72,7 +72,7 @@ const containerStyles: Interpolation<Props> = ({
     display: "inline-flex",
     alignItems: "center",
     justifyContent: "center",
-    padding: `0 ${condensed ? theme.space.small : theme.space.content}px`,
+    padding: `0 ${condensed ? theme.space.medium : theme.space.element}px`,
     borderRadius: theme.borderRadius,
     border: 0,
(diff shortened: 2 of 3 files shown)


ImogenF:
[Button horizontal padding](https://trello.com/c/BVsaoPw9/282-button-horizontal-padding)

packages/components/src/utils/constants.ts
```diff
@@ -145,6 +145,9 @@ const space = {
   /** Small space is `8px` */
   small: 8,
 
+  /** Medium space is `12px` */
+  medium: 12,
```
TejasQ: 👍

TejasQ:
I think it's an _"either or"_ case. 

I would love to see all the buttons as disabled and as condensed. If we think it's bloated, I'd call into question why we even need the second row of disabled styles.

I'm fine with:
a) having _one_ disabled button for show, and _one_ condensed, or
b) having all buttons with all props and a lot of duplication displayed as examples.

I think it strange to have all normal, and disabled states displayed as they are now, but only one condensed example.

TejasQ:
What do you prefer? @ImogenF @peterszerzo @fabien0102?

ImogenF:
My thought process was that disabled looks different depending on the colour of the button. Condensed buttons do not change the color in any way, just the size, so showing condensed buttons for every possible color seems redundant.

peterszerzo:
Nothing wrong with a little redundancy with a huge matrix of all kinds of shapes and sizes I'd say - if anything, it can help with design decisions layout out all options one can choose from when working out how a new UI element should look like.

ImogenF:
How about I get rid of the condensed button in the colours row, and add to the next row down?

image

TejasQ:
That makes a lot of sense to me. 👍

peterszerzo:
Tells the whole story I'd say :).

TejasQ:
@micha-f how does it look?

micha-f:
Looks great 👍

Issue #388, opened 2018-04-13
Ability to set hard constraints on date range (date picker).


nagautier:
Right now, a user can specify any date range on the calendar. Often some data is only available for a certain date range. Clicking outside the range results in no data, which can be frustrating to the user as they are not always aware of the available data date range.
It would be useful to be able to specify hard limits on the date range above which the user cannot click (greyed out for example).

peterszerzo:
Awesome - great feature suggestion, @nagautier, we'll see if we can squeeze it in this week!

TejasQ:
@peterszerzo what's the status here? @nagautier do you still need this?

TejasQ:
Thanks!

peterszerzo:
I can take care of it by Monday close of business.

TejasQ:
Merged in #594, closing.

PR #594, opened 2018-07-17
Date picker min-max ranges and visual updates

peterszerzo:
### Summary

This PR adds min/max props that allow upper and lower date-bounds be set on this component - in `<DatePicker min="2018-01-01" ... />`, dates before the beginning of this year may not be selected. The restriction only applies to dates that may come out of the `onChange` method - any initial date range can be passed in, only ones within the allowed range can come out.

I took this opportunity to also improve the styling and add some hover background- and cursor feedback. 

screen shot 2018-07-17 at 9 20 33 am

### Related issue

https://github.com/contiamo/operational-ui/issues/388

### To be tested

Tester 1 (Tejas)

- [x] Date picker layout works on Chrome
- [x] Date picker layout works on Firefox
- [x] Date picker layout works on Safari (if mac user)
- [ ] Date picker layout works on IE11 and Edge (if windows user)
- [x] `min` / `max` props are working

Tester 2 (Imo)

- [x] Date picker layout works on Chrome
- [x] Date picker layout works on Firefox
- [ ] Date picker layout works on Safari (if mac user)
- [ ] Date picker layout works on IE11 and Edge (if windows user)
- [x] `min` / `max` props are working

### DIFF
--- a/packages/components/src/DatePicker/DatePicker.Month.tsx
+++ b/packages/components/src/DatePicker/DatePicker.Month.tsx
@@ -35,7 +35,6 @@ const isSelected = (date: string, current: IDatePair): boolean => {
 }
 
 const Month = ({ year, month, start, end, onChange, min, max }: Props) => {
-  console.log(min, max)
   const prevPlaceholderDays = monthStartDay(year, month)
   const nextMonth = month === 11 ? 0 : month + 1
   const nextYear = month === 11 ? year + 1 : year
--- a/packages/components/src/DatePicker/DatePicker.Month.tsx
+++ b/packages/components/src/DatePicker/DatePicker.Month.tsx
@@ -13,6 +13,8 @@ export interface Props {
   month: number
   start?: string
   end?: string
+  min?: string
+  max?: string
   onChange?: (date: IDatePair) => void
 }
 
@@ -32,7 +34,8 @@ const isSelected = (date: string, current: IDatePair): boolean => {
   return date === start || date === end || (!!start && !!end && date >= start && date <= end)
 }
 
-const Month = ({ year, month, start, end, onChange }: Props) => {
+const Month = ({ year, month, start, end, onChange, min, max }: Props) => {
+  console.log(min, max)
   const prevPlaceholderDays = monthStartDay(year, month)
   const nextMonth = month === 11 ? 0 : month + 1
   const nextYear = month === 11 ? year + 1 : year
@@ -47,6 +50,7 @@ const Month = ({ year, month, start, end, onChange }: Props) => {
       {range(prevPlaceholderDays).map((number, index) => {
         const day = daysInPreviousMonth + index - prevPlaceholderDays
         const date = toDate(prevYear, prevMonth, day)
+        const isDisabled = (min && date < min) || (max && date > max)
         return (
           <Day
             selected={isSelected(date, {
@@ -55,8 +59,12 @@ const Month = ({ year, month, start, end, onChange }: Props) => {
             })}
             key={index}
             isPlaceholder
+            isDisabled={isDisabled}
             onClick={(ev: any) => {
               ev.preventDefault()
+              if (isDisabled) {
+                return
+              }
               onChange &&
                 onChange(
                   setNewDate(date, {
@@ -72,6 +80,7 @@ const Month = ({ year, month, start, end, onChange }: Props) => {
       })}
       {range(daysInCurrentMonth).map((number, index) => {
         const date = toDate(year, month, index)
+        const isDisabled = (min && date < min) || (max && date > max)
         return (
           <Day
             selected={isSelected(date, {
@@ -79,8 +88,12 @@ const Month = ({ year, month, start, end, onChange }: Props) => {
               end,
             })}
             key={index}
+            isDisabled={isDisabled}
             onClick={(ev: any) => {
               ev.preventDefault()
+              if (isDisabled) {
+                return
+              }
               onChange &&
                 onChange(
                   setNewDate(date, {
@@ -96,6 +109,7 @@ const Month = ({ year, month, start, end, onChange }: Props) => {
       })}
       {range(nextPlaceholderDays).map((number, index) => {
         const date = toDate(nextYear, nextMonth, number)
+        const isDisabled = (min && date < min) || (max && date > max)
         return (
           <Day
             key={index}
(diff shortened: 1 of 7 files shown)


TejasQ:
What are your thoughts about the `minDate` and `maxDate` inputs being `Date` objects instead of strings in a particular format?

ImogenF:
Would be be possible to restrict it so that the user can't keep clicking backwards / forwards through months that they can't select? I feel like that would make it more obvious that the selection is restricted.

peterszerzo:
The reason I decided against date objects at the beginning is that they contain more information than is strictly relevant for this UI (exact time of day as opposed to just the day itself). The `2017-01-02` format seems simpler and more expressive in comparison (removes doubts like 'if I select until midnight right before this day, should it count or not?'), and it still allows date objects to be constructed from it, or date objects converted into it with very little effort.

peterszerzo:
In any case, I think we should be consistent with the formats of `start`, `end`, `min` and `max`.

TejasQ:
I agree with your reasoning. From the other side, I would find passing in a Date more comfortable, having Operational handle the formatting and processing internally.

```jsx
const myDate = new Date()

const Thing = () => <DatePicker maxDate={myDate} />
```

as opposed to 

```jsx
import { format } from "date-fns"

const Thing = () => <DatePicker minDate={format('YYYY-mm-DD', new Date())} />
```

I'd love more thoughts from @contiamo/frontend here.

peterszerzo:
@TejasQ regardless of what we decide here, I think this will be a substantial change that belongs in a different PR. Would you mind opening an issue on it so we can get this feature out and take our time with the public API? The changes will be able to build on top of each other without wasted effort I would say.

TejasQ:
Let's open an issue _if_ the current implementation is actually an issue.

TejasQ:
Since dates outside of the current month are greyed out, they look identical to disabled dates now and it's a little hard to tell the difference. See below.

hello

packages/components/src/DatePicker/DatePicker.Month.tsx
```diff
@@ -13,6 +13,8 @@ export interface Props {
   month: number
```
TejasQ: There seems to be a lot of duplication in this file. Shall we open an issue and revisit it later?

packages/components/src/DatePicker/DatePicker.styles.ts
```diff
@@ -65,10 +67,10 @@ export const Toggle = styled("div")(({ theme }: { theme?: OperationalStyleConsta
 }))
 
 export const MonthNav = styled("div")(({ theme }: { theme?: OperationalStyleConstants }) => ({
-  marginBottom: theme.deprecated.spacing / 2,
+  marginBottom: theme.space.element,
   textAlign: "center",
   "& > *": {
-    margin: `0 6px`,
+    margin: `0 2px`,
```
TejasQ: Why the backticks?

TejasQ:
It works great now! Can the chevrons be removed when the user can't navigate outside of the range?

packages/components/src/DatePicker/DatePicker.Month.tsx
peterszerzo: Do you mean this kind of snippet appearing three times:
```
onClick={(ev: any) => {
              ev.preventDefault()
              if (isDisabled) {
                return
              }
              onChange &&
                onChange(
                  setNewDate(date, {
                    start,
                    end,
                  }),
                )
            }}
```

peterszerzo:
I think it's better to keep the chevrons with disabled feedback - you are right that the change in opacity was too little. I bumped it a little in the newest commit.

packages/components/src/DatePicker/DatePicker.Month.tsx
TejasQ: I was talking about [this](https://github.com/contiamo/operational-ui/blob/a478fe4ddc21679ea364cf8d824abe474a79b5ef/packages/components/src/DatePicker/DatePicker.Month.tsx#L49-L79).

peterszerzo:
@TejasQ you're right about the snippet. I've opened https://github.com/contiamo/operational-ui/issues/603, as it will take a bit longer (I tried a quick abstraction but it turned out clunky). Fixing the last day issue.

packages/components/src/DatePicker/DatePicker.tsx
```diff
@@ -137,10 +136,13 @@ class DatePicker extends React.Component<Props, State> {
   }
 
   render() {
-    const { onChange, placeholder, start, end, label, id, className } = this.props
+    const { onChange, placeholder, start, end, label, min, max, id, className } = this.props
     const { isExpanded, month, year } = this.state
     const domId = id || (label && label.toLowerCase ? label.toLowerCase().replace(/\s/g, "-") : null)
 
+    const canGoToPreviousMonth = !min || min < toDate(this.state.year, this.state.month, 0)
+    const canGoToNextMonth = !max || max > toDate(this.state.year, this.state.month, 30)
```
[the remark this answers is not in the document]
peterszerzo: The 30th day (numbering starting from 0) is the last possible day, to which the max date can be compared to determine whether it goes over.

peterszerzo:
@TejasQ I refactored the code so it doesn't rely on the magical number 30 anymore. Thanks for the catch, I think it's a lot more reliable now.

packages/components/src/DatePicker/DatePicker.styles.ts
```diff
@@ -105,24 +138,34 @@ export const Day = styled("div")(
...
+    isDisabled?: boolean
   }) => ({
     ...theme.deprecated.typography.body,
     backgroundColor: selected ? theme.deprecated.colors.info : "transparent",
-    color: selected
-      ? theme.deprecated.colors.white
-      : isPlaceholder
-        ? theme.deprecated.colors.gray
-        : theme.deprecated.colors.black,
+    color: makeDayTextColor({ isPlaceholder, isDisabled, selected, theme }),
```
TejasQ: wow.

peterszerzo:
@TejasQ I added unit tests for `changeMonth`, it being more mission-critical.

@ImogenF having trouble reproducing safari and firefox bugs, will check on your machine when you have time.

TejasQ:
@fabien0102 would you help test on IE11 here?

PR #595, opened 2018-07-17
Play, pause and menu icons

peterszerzo:
### Summary

Adding four new icons: play, pause, resume and menu:

screen shot 2018-07-17 at 9 45 08 am
screen shot 2018-07-17 at 9 45 11 am
screen shot 2018-07-17 at 9 47 09 am
screen shot 2018-07-17 at 10 36 54 am

### Related issue

https://trello.com/c/DlYjCvJR/273-more-icons

### To be tested

Tester 1 (Imogen)

- [x] The new icons look good
- [x] They work well in their intended contexts (feel free to add your concerns pre- or post-merge)

Tester 2

- [ ] The new icons look good
- [ ] They work well in their intended contexts (feel free to add your concerns pre- or post-merge)


TejasQ:
@micha-f how do they look?

micha-f:
They look good! Can we add a "resume" icon, too? Any ideas for how that could look?

peterszerzo:
@micha-f what I could do is to add a small addition to the play icon and have it be similar but with a subtle cue (a little missing link in the path or something) that the play has already been started. I tend to see play icons be identical to resume, and I wonder if we'll confuse people if we went with anything drastically different.

peterszerzo:
@micha-f added the resume icon in the PR description.

micha-f:
@peterszerzo beautiful!

TejasQ:
The resume icon looks to me like a "seek to end" or "next", but I'm no designer so.

peterszerzo:
It does look a bit like the fast-forward icon, or as you say, navigate to the end of some list, but since you rarely see this icon on its own (only after you already paused it), it shouldn't cause too much confusion.

Issue #517, opened 2018-06-21
Add `className` prop to all components


peterszerzo:
~~Relying on class names is brittle, risking typos (or making string constants necessary) and opening up an escape hatch to meddle with the insides of components. It is better to remove `className` props entirely and rely on `emotion` component-based styling instead.~~

After running into some tricky-to-debug styling errors using the components, the scope of this PR has changed. Read along :)

peterszerzo:
I ran into the following classname-related issue: not passing `className` prop to a component makes it impossible to use `emotion`'s `styled` helper, an escape hatch that is necessary in internal implementation (e.g. `Icon.tsx`), as well as in hotfixes on client code. Worse yet, the error goes silent, and proved to be very difficult to debug. In light of this, I recommend we always pass class names to all our components.

Quick context here: in the `css` prop removal discussion, it came up that using `styled` is a more verbose than writing `<Button css={{ marginRight: 0 }} />`, and therefore should discourage itself as a CSS override escape hatch by virtue of being annoying to write.

TejasQ:
I'm totally in favor of allowing a `className` prop everywhere. I think it makes sense for components to receive `className`s all the time, but _not_ ad-hoc `css` styles.

Good catch, Peter!

fabien0102:
I think that we can try this pattern for our module -> https://github.com/pedronauck/docz/blob/master/examples/typescript/src/components/Alert.tsx

So we retrieve all native react props (including classname), and it's pretty 😎

TejasQ:
How about we just spread props in `Container`s?

peterszerzo:
That's what @fabien0102's link suggested as well I believe

PR #596, opened 2018-07-17
Feat/textarea fullwidth

fabien0102:
### Summary

image

To be able to do this screen, I added `fullWidth` property to `TextArea` and `flexColumn` to `CardColumn`

### Other fixes
* Remove the following warning: 
image

* Fix some type issue with `withLabel` that I have when I'm linking operational-ui into labs-ui

### Related issue

https://trello.com/c/FW1EFFIu

### To be tested

**Fabien**

- [x] I can reproduce the screenshot in labs-ui easily 😎 (with npm-link)

**Tejas**

- [ ] I can set TextArea in fullWidth mode
- [ ] ContextMenu is working as before
- [ ] Checkbox is working as before

**Peter**

- [x] I can set TextArea in fullWidth mode
- [x] ContextMenu is working as before
- [x] Checkbox is working as before

### DIFF
--- a/packages/components/src/CardColumn/CardColumn.tsx
+++ b/packages/components/src/CardColumn/CardColumn.tsx
@@ -7,10 +7,13 @@ export interface Props {
   title?: string
   /** Align the content to the right */
   contentRight?: boolean
+
+  /** Set the component as a flex-box column */
+  flexColumn?: boolean
 }
 
 const Container = styled("div")(
-  ({ theme, contentRight }: { theme?: OperationalStyleConstants; contentRight?: Props["contentRight"] }) => ({
+  ({ theme, contentRight, flexColumn }: { theme?: OperationalStyleConstants } & Props) => ({
     label: "card-column",
     minWidth: 280 / 2,
     padding: theme.space.element / 2,
@@ -19,6 +22,13 @@ const Container = styled("div")(
       maxWidth: "100%",
     },
     textAlign: contentRight ? "right" : null,
+    ...(flexColumn
+      ? {
+          display: "flex",
+          flexDirection: "column",
+          justifyContent: contentRight ? "flex-end" : "flex-start",
+        }
+      : {}),
   }),
 )
 
@@ -32,7 +42,7 @@ const Title = styled("div")(({ theme }: { theme?: OperationalStyleConstants }) =
   marginBottom: theme.space.content,
 }))
 
-const CardColumn: React.SFC<Props & React.HTMLAttributes<HTMLDivElement>> = ({ title, children, ...props }) => (
+const CardColumn: React.SFC<Props> = ({ title, children, ...props }) => (
   <Container {...props}>
     {title && <Title>{title}</Title>}
     {children}
--- a/packages/components/src/CardColumn/CardColumn.tsx
+++ b/packages/components/src/CardColumn/CardColumn.tsx
@@ -32,7 +32,7 @@ const Title = styled("div")(({ theme }: { theme?: OperationalStyleConstants }) =
   marginBottom: theme.space.content,
 }))
 
-const CardColumn: React.SFC<Props> = ({ title, children, ...props }) => (
+const CardColumn: React.SFC<Props & React.HTMLAttributes<HTMLDivElement>> = ({ title, children, ...props }) => (
   <Container {...props}>
     {title && <Title>{title}</Title>}
     {children}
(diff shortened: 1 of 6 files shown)


packages/components/src/Textarea/Textarea.tsx
```diff
@@ -30,6 +28,9 @@ export interface Props {
   error?: string
   /** Is it disabled? */
```
TejasQ: Do you see this blank line?

packages/components/src/Textarea/Textarea.tsx
```diff
@@ -30,6 +28,9 @@ export interface Props {
...
 
+  /** Should the input fill its container? */
+  fullWidth?: boolean
+
   disabled?: boolean
```
TejasQ: This should be on line 30. 🤔

packages/components/src/Textarea/Textarea.tsx
```diff
@@ -2,10 +2,8 @@ import * as React from "react"
 import styled from "react-emotion"
 import { OperationalStyleConstants } from "../utils/constants"
 
-import { Label, LabelText, FormFieldControls, FormFieldControl, FormFieldError, inputFocus } from "../utils/mixins"
-import { Icon, Hint } from "../"
-import Tooltip from "../Tooltip/Tooltip" // Styled components appears to have an internal bug that breaks when this is imported from index.ts
```
TejasQ: Great job removing all the trash.

packages/components/src/CardColumn/CardColumn.tsx
```diff
@@ -32,7 +32,7 @@ const Title = styled("div")(({ theme }: { theme?: OperationalStyleConstants }) =
   marginBottom: theme.space.content,
 }))
 
-const CardColumn: React.SFC<Props> = ({ title, children, ...props }) => (
+const CardColumn: React.SFC<Props & React.HTMLAttributes<HTMLDivElement>> = ({ title, children, ...props }) => (
```
TejasQ: ? I think #517 should solve this. I am not sure this is necessary.

fabien0102: I was wrong, `React.SFC` don't add so much props:

```jsx
type SFC<P = {}> = StatelessComponent<P>;
    interface StatelessComponent<P = {}> {
        (props: P & { children?: ReactNode }, context?: any): ReactElement<any> | null;
        propTypes?: ValidationMap<P>;
        contextTypes?: ValidationMap<any>;
        defaultProps?: Partial<P>;
        displayName?: string;
    }
```

fabien0102: I can also use `JSX.IntrincisElements<div>` here I think (I'm trying to follow the type definition of emotion as example ;) )

packages/components/src/utils/with-label.tsx
```diff
@@ -15,16 +14,14 @@ const Label = styled("label")(({ theme }: { theme?: OperationalStyleConstants })
   marginBottom: theme.deprecated.spacing / 4,
 }))
 
-type IComponent = Pick<React.HTMLProps<HTMLDivElement>, "className" | "label" | "id"> & {
-  css?: Css
-}
+type IComponent = Pick<React.HTMLProps<HTMLDivElement>, "className" | "label" | "id">
```
TejasQ: Is this type used? If yes, can we rename `IComponent` to something more descriptive? I see this and I immediately think "why the `I` prefix"? 

I know it's old, but since we touched it, let's leave it better than it was.

fabien0102: I was thinking the same, but the respond is to avoid the conflict with the line 27 ;) I can also call this `ComponentType` if you prefer

TejasQ: Let's refactor the entire implementation per our offline conversation.

packages/components/src/Textarea/Textarea.tsx
```diff
@@ -2,34 +2,35 @@ import * as React from "react"
...
-import { Css, CssStatic } from "../types"
+import { Label, LabelText, FormFieldControls, FormFieldError, inputFocus } from "../utils/mixins"
+import Hint from "../Hint/Hint"
 
 export interface Props {
   id?: string
   className?: string
```
peterszerzo: Can we remove the extra lines here?

packages/components/src/CardColumn/CardColumn.tsx
```diff
@@ -7,10 +7,13 @@ export interface Props {
   title?: string
   /** Align the content to the right */
   contentRight?: boolean
+
```
peterszerzo: Remove extra line

PR #598, opened 2018-07-17
Add action context menu.

### LINKED ISSUE #608: Action menu styling
### Problem

Currently, since `ActionMenu` makes use of the `ContextMenu` component, the two are styled independently. This results in the boxShadow along the top border of the `ContextMenu` overlapping the "Actions" button:

image

Ideally, we would want the box shadow to go fully around the button *and* the menu together.

### Proposed solutions

1) Redesign `ContextMenu`:

Enclosing (a copy of) the item from which the context menu is triggered and the menu itself inside a container, so that the boxShadow can be applied to this container instead of just the menu. This would also be useful for the `HeaderMenu` component as well.

However, this would make `ContextMenu` unsuitable for use with anything except buttons or rectangular divs of the same width as the menu. We don't currently have any uses where this is not the case, but could be problematic. 

2) Do not use `ContextMenu`:

This would require completely redesigning the `ActionMenu` and implementing the menu and its behaviour from scratch.

ImogenF:
### Summary

<blockquote class="trello-card"><a href="https://trello.com/c/cJm742k2/281-action-context-menu">Action context menu</a></blockquote>

Fixes #608 

### To be tested

Tejas

- [x] Look of the component - difficult to make it exactly the same as in Zeplin.

Fabien

- [x] Look of the component - difficult to make it exactly the same as in Zeplin.

### DIFF
--- a/packages/components/src/ActionMenu/ActionMenu.tsx
+++ b/packages/components/src/ActionMenu/ActionMenu.tsx
@@ -28,7 +28,7 @@ const Container = styled("div")(({ theme }: { theme?: OperationalStyleConstants 
   alignItems: "center",
   justifyContent: "space-between",
   userSelect: "none",
-  outline: `1px solid ${theme.color.separators.light}`,
+  boxShadow: `0 0 0 1px ${theme.color.separators.light}`,
   "&:hover": {
     backgroundColor: darken(theme.color.white, 5),
   },
--- a/packages/components/src/ActionMenu/ActionMenu.tsx
+++ b/packages/components/src/ActionMenu/ActionMenu.tsx
@@ -1,0 +1,56 @@
+import * as React from "react"
+import styled from "react-emotion"
+import { OperationalStyleConstants } from "../utils/constants"
+import ContextMenu, { Props as CMProps } from "../ContextMenu/ContextMenu"
+import Icon from "../Icon/Icon"
+
+export interface Props {
+  /** Action when item in dropdown is selected - if specified here, it is applied to all dropdown items */
+  onClick?: CMProps["onClick"]
+  /** Title */
+  title?: string
+  /** Actions to display in dropdown */
+  items: CMProps["items"]
+}
+
+const Container = styled("div")(({ theme, isOpen }: { theme?: OperationalStyleConstants; isOpen: boolean }) => ({
+  width: 144,
+  height: 35,
+  padding: `0 ${theme.space.content}px`,
+  backgroundColor: theme.color.white,
+  ...(isOpen
+    ? {
+        boxShadow: `0 3px 12px ${theme.color.border.disabled}`,
+        borderColor: `${theme.color.separators.light} ${theme.color.white}`,
+        borderWidth: "1px",
+        borderStyle: "solid",
+      }
+    : {
+        border: `1px solid ${theme.color.separators.light}`,
+      }),
+  color: theme.color.primary,
+  fontWeight: theme.font.weight.medium,
+  borderRadius: theme.borderRadius,
+  display: "flex",
+  alignItems: "center",
+  justifyContent: "space-between",
+}))
+
+const ActionMenu: React.SFC<Props> = (props: Props) => {
+  return (
+    <ContextMenu {...props} items={props.items} width={144} condensed>
+      {isOpen => (
+        <Container isOpen={isOpen}>
+          {props.title}
+          <Icon name={"Menu"} />
+        </Container>
+      )}
+    </ContextMenu>
+  )
+}
+
+ActionMenu.defaultProps = {
+  title: "Actions",
+}
+
+export default ActionMenu
--- a/packages/components/src/ActionMenu/ActionMenu.tsx
+++ b/packages/components/src/ActionMenu/ActionMenu.tsx
@@ -43,18 +43,16 @@ const TitleContainer = styled("p")({
   textOverflow: "ellipsis",
 })
 
-const ActionMenu: React.SFC<Props> = (props: Props) => {
-  return (
-    <ContextMenu {...props} items={props.items} width={144} condensed>
-      {isOpen => (
-        <Container isOpen={isOpen}>
-          <TitleContainer>{props.title}</TitleContainer>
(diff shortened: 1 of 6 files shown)


ImogenF:
[Action context menu](https://trello.com/c/cJm742k2/281-action-context-menu)

packages/components/src/ActionMenu/ActionMenu.tsx
```diff
@@ -0,0 +1,56 @@
+import * as React from "react"
+import styled from "react-emotion"
+import { OperationalStyleConstants } from "../utils/constants"
+import ContextMenu, { Props as CMProps } from "../ContextMenu/ContextMenu"
```
peterszerzo: Can we write out `ContextMenuProps` here? It makes the file easier to scan.

packages/components/src/ContextMenu/ContextMenu.Item.tsx
```diff
@@ -13,13 +14,15 @@ const Container = styled("div")(
     theme,
```
peterszerzo: Let's add a `userSelect: "none"` to this to make sure it doesn't get selected when the context menu expands.

packages/components/src/ActionMenu/ActionMenu.tsx
```diff
@@ -0,0 +1,56 @@
...
+  justifyContent: "space-between",
+}))
+
+const ActionMenu: React.SFC<Props> = (props: Props) => {
+  return (
+    <ContextMenu {...props} items={props.items} width={144} condensed>
+      {isOpen => (
+        <Container isOpen={isOpen}>
+          {props.title}
+          <Icon name={"Menu"} />
```
peterszerzo: Change the icon to `ChevronUp` when the menu is open.

fabien0102:
image
The component can break with a long `title` ;)

packages/components/src/ActionMenu/ActionMenu.tsx
```diff
@@ -0,0 +1,56 @@
...
+      }),
+  color: theme.color.primary,
+  fontWeight: theme.font.weight.medium,
+  borderRadius: theme.borderRadius,
+  display: "flex",
+  alignItems: "center",
+  justifyContent: "space-between",
+}))
+
+const ActionMenu: React.SFC<Props> = (props: Props) => {
```
fabien0102: Here you can return directly the jsx like this:

```jsx
conrst ActionMenu: React.SFC<Props> = (props: Props) => (
  <MyBeautifulComponent />
)
```

packages/components/src/ContextMenu/ContextMenu.tsx
```diff
@@ -70,8 +72,8 @@ class ContextMenu extends React.Component<Props, State> {
     isOpen: false,
   }
 
-  defaultProps = {
-    align: "left",
+  static defaultProps = {
```
fabien0102:
```jsx
static defaultProps: Partial<Props> = {}
```

packages/components/src/ActionMenu/ActionMenu.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+const TitleContainer = styled("p")({
+  width: 90,
+  whiteSpace: "nowrap",
+  overflow: "hidden",
+  textOverflow: "ellipsis",
+})
+
+const ActionMenu: React.SFC<Props> = (props: Props) => (
+  <ContextMenu {...props} items={props.items} width={144} condensed>
```
TejasQ: Width `144` here and on line 17 seems a bit of a smell.

packages/components/src/ActionMenu/README.md
```diff
@@ -0,0 +1,10 @@
+### Simple usage
+
+```jsx
+const items = [
+  { label: "Open editor", onClick: () => {} },
```
TejasQ: Why is `onClick` mandatory?

ImogenF: It's not. It's just in there so that the hover effect over the items can be seen. If there is no `onClick` method for an item, there's no hover effect and just the regular arrow cursor.

ImogenF:
@peterszerzo @TejasQ @fabien0102 I've implemented all of the changes you suggested. Would appreciate another review!

packages/components/src/ActionMenu/README.md
TejasQ: Best to probably leave a comment there since this is the docs website.

TejasQ:
image

It passes the test. 👍

TejasQ:
@micha-f [what do you think](https://deploy-preview-598--operational-ui.netlify.com/#!/ActionMenu)?

micha-f:
@TejasQ Discussed styling with @ImogenF. She'll try to get rid of the shadow below "Actions".

ImogenF:
There is no obvious solution to the styling problem. Have detailed the problem and suggested a few potential solutions in: https://github.com/contiamo/operational-ui/issues/608. Would appreciate any feedback or alternative suggestions.

ImogenF:
@micha-f Can you check you're happy with this visually please? Should be closer to what you had in mind now.

micha-f:
@ImogenF Looks amazing. One last thing - doesn't look like the 2px border-radius is applied. Can you check?

ImogenF:
@TejasQ Duuuuh, that was stupid of me. Coming right up.

packages/components/src/ActionMenu/ActionMenu.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+const TitleContainer = styled("p")({
+  width: 90,
+  whiteSpace: "nowrap",
+  overflow: "hidden",
+  textOverflow: "ellipsis",
+})
+
+const ActionMenu: React.SFC<Props> = (props: Props) => (
+  <ContextMenu {...props} items={props.items} width={width} condensed embedChildrenInMenu>
```
TejasQ: I'm not so sure the naming of `embedChildrenInMenu` is ideal. Are there other alternatives we cank think of?

PR #600, opened 2018-07-17
Remove unused components

peterszerzo:
### Summary

Removing `Tile`, `Record`, `Timeline` and `Header` components. Moving snapshot tests in with the individual components. Fix blacklist bug on checking imports. 

### To be tested

Tester 1 (Imo)

- [x] Remaining components look and work as expected

Tester 2 (Tej)

- [x] Remaining components look and work as expected
- [x] Netlify looks good


ImogenF:
Is component/docs/getting-started.md actually being used anywhere? If so, it uses the `Header` component so might be a good idea to take that out. Otherwise looks good!

peterszerzo:
@ImogenF I'll deal with the imported `Header` component when it updates. Took out the docs, though.

packages/components/src/__tests__/libExports.ts
```diff
@@ -1,7 +1,7 @@
 import { readdirSync, readFileSync } from "fs"
 import { join } from "path"
 
-const BLACK_LIST = ["Internals", "utils", "__tests__", "types.ts", "index.ts"]
+const BLACK_LIST = ["Internals", "utils", "__tests__", "types.ts", "index.ts", ".DS_Store"]
```
fabien0102: I'm surprise that nobody add this before ^^ 🍎 🍏

peterszerzo: You were supposed to remember? Then you started ranting about how Macs shouldn't generate this file :)

PR #610, opened 2018-07-19
Fix page routing sync

fabien0102:
### Summary
The page routing can be out of sync if the router state change and the component is already mount.

Before:
ef415011-6e60-4cc4-93e6-0c1c1257e0a0

After:
a75bb28e-c789-4802-a552-9db0a7a5ee37

### To be tested
Fabien 

Tejas
- [x] The Page tab is sync with the router state (cf Page/With activeTabName trigger from outside (router) example)

Tester 2

- [ ] The Page tab is sync with the router state (cf Page/With activeTabName trigger from outside (router) example)

### DIFF
--- a/packages/components/src/Page/Page.tsx
+++ b/packages/components/src/Page/Page.tsx
@@ -103,6 +103,15 @@ class Page extends React.Component<Props, Readonly<typeof initialState>> {
     }
   }
 
+  componentDidUpdate(props: Props) {
+    if (this.props.activeTabName && props.activeTabName !== this.props.activeTabName) {
+      const index = this.props.tabs.findIndex(
+        ({ name }) => name.toLowerCase() === this.props.activeTabName.toLowerCase(),
+      )
+      this.setState({ activeTab: index === -1 ? 0 : index })
+    }
+  }
+
   onTabClick(index: number) {
     this.setState({ activeTab: index })
     this.props.onTabChange && this.props.onTabChange(this.props.tabs[index].name.toLowerCase())
--- a/packages/components/src/Page/Page.tsx
+++ b/packages/components/src/Page/Page.tsx
@@ -24,7 +24,7 @@ export interface Props {
   /**
    * Active tab name
    *
-   * Useful for easy router mapping
+   * If not specified, active tab is controlled by internal state.
    */
   activeTabName?: string
   /**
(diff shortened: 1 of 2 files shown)


packages/components/src/Page/Page.tsx
```diff
@@ -103,6 +103,15 @@ class Page extends React.Component<Props, Readonly<typeof initialState>> {
     }
   }
 
+  componentDidUpdate(props: Props) {
+    if (this.props.activeTabName && props.activeTabName !== this.props.activeTabName) {
+      const index = this.props.tabs.findIndex(
+        ({ name }) => name.toLowerCase() === this.props.activeTabName.toLowerCase(),
+      )
+      this.setState({ activeTab: index === -1 ? 0 : index })
```
peterszerzo: Can we avoid duplicating state here? It looks like the `render` method can decide deterministically whether `this.props.activeTagName` or `this.state.activeTab` should win, making it unnecessary to cause another render loop, and making the separation of responsibilities between internal state and outside props easier to follow.

peterszerzo: Isn't it the case that we know for sure whether the component is controlled or not at the time we're using it? If so, we can just say that rendering should be controlled by whatever is in props, and when that is absent, it goes by whatever is in state. I think we should avoid mixing/syncing between them if possible.

TejasQ: Please summarize the fruits of your offline conversation here.

peterszerzo: Sure thing - we sketched out the solution without copying over updated props values into state and we wound up liking it in the end. You should see it in the next commit.

We're also adding some documentation so it's clear when the component is controlled and when it's uncontrolled.

packages/components/src/Page/Page.tsx
```diff
@@ -90,27 +90,30 @@ class Page extends React.Component<Props, Readonly<typeof initialState>> {
...
 
+  getActiveTab(): number {
+    let activeTab: number
+    if (this.props.activeTabName) {
+      const index = this.props.tabs.findIndex(
+        ({ name }) => name.toLowerCase() === this.props.activeTabName.toLowerCase(),
+      )
+      activeTab = index === -1 ? 0 : index
+    } else {
+      activeTab = this.state.activeTab
```
TejasQ: If we early return, we can remove the `activeTab` flag and tidy up the implementation a little bit (read: stroke my functional programming muscle).

PR #611, opened 2018-07-19
Add tree component

### LINKED ISSUE #590: Tree component
This is an API proposal for the tree component per the following sketch:

image

### Design goals

* component renders elements in a folder tree-like structure, rendering a label and an optional `NameTag` at each node. The component should be agnostic about the level of nesting to avoid verbose type definitions.
* supports any HTML attributes and event listeners at each level - both click events as well as drag-and-drop handlers are important for individual nodes.
* the outside world need not know about open-closed state - this should be handled in internal state.
* opinionated on node styling (font-weight, color etc.) based on the nesting level / whether there is a tag or not.

### API

The basis for the props of this component is a recursive data structure that models an arbitrarily nested tree:

```
export interface Tree {
  label: string
  tag?: string
  childNodes: Tree[]
}
```

Create a component like so:

```
(
  <Tree
    tree={{
      label: "Dimensions",
      childNodes: [
        {
          label: "Dim 1",
          tag: "DM",
          childNodes: [],
        },
        {
          label: "Dim 2",
          tag: "DM",
          onClick: () => {
            console.log("clicked")
          },
          childNodes: [
            {
              label: "Subdim 1",
              tag: "SM",
              childNodes: [],
            },
          ],
        },
      ],
    }}
  />
)
```

The component's implementation handles active state by keeping track of active node paths in an array, sorting out conflicts when new props arrive (although this should be infrequent).

peterszerzo:
### Summary

Collapsible tree component per following gif:

tree3

### Progress

* [x] basic state management, public API.
* [x] handle initially open state.
* [x] handle custom tag colors.
* [x] handle disabled state.
* [x] visual fine-tuning.

### Related issue

Fixes #590.

https://trello.com/c/hAECaWyR/178-simple-expandable-tree-with-3-levels

### To be tested

Tester 1 - Tekas

- [x] Looks and works as expected with tagged nodes at different levels.

Tester 2

- [ ] Looks and works as expected with tagged nodes at different levels.

### DIFF
--- a/packages/components/src/Tree/Tree.tsx
+++ b/packages/components/src/Tree/Tree.tsx
@@ -2,10 +2,10 @@ import * as React from "react"
 import styled from "react-emotion"
 
 import constants, { OperationalStyleConstants, expandColor } from "../utils/constants"
-import NameTag from "../NameTag/NameTag"
 import Icon from "../Icon/Icon"
 import { togglePath, containsPath, getInitialOpenPaths } from "./Tree.utils"
 import { Tree as ITree } from "./Tree.types"
+import SmallNameTag from "../Internals/SmallNameTag"
 
 export interface Props {
   /** An array of tree structures */
@@ -94,9 +94,9 @@ const TreeRecursive: React.SFC<{
       >
         <Icon left name={isOpen ? "ChevronDown" : "Add"} size={12} />
         {tree.tag && (
-          <NameTag color={tagColor} left>
+          <SmallNameTag color={tagColor} left>
             {tree.tag}
-          </NameTag>
+          </SmallNameTag>
         )}
         {tree.label}
       </TreeItem>
--- a/packages/components/src/Tree/Tree.tsx
+++ b/packages/components/src/Tree/Tree.tsx
@@ -35,6 +35,7 @@ const TreeItem = styled("div")`
   padding: 4px 8px;
   width: fit-content;
   min-width: 160px;
+  user-select: none;
   ${({
     theme,
     hasChildren,
@@ -118,7 +119,7 @@ const TreeRecursive: React.SFC<{
 }
 
 class Tree extends React.Component<Props, State> {
-  state: State = {
+  readonly state: State = {
     openPaths: this.props.trees
       .map((tree, index) => getInitialOpenPaths([index])(tree))
       .reduce((current, accumulator) => [...current, ...accumulator], []),
--- a/packages/components/src/Tree/Tree.tsx
+++ b/packages/components/src/Tree/Tree.tsx
@@ -1,22 +1,14 @@
 import * as React from "react"
 import styled from "react-emotion"
 
-import { OperationalStyleConstants } from "../utils/constants"
+import constants, { OperationalStyleConstants, expandColor } from "../utils/constants"
 import NameTag from "../NameTag/NameTag"
 import Icon from "../Icon/Icon"
-import { togglePath, containsPath } from "./Tree.utils"
-
-/**
- * ITree naming convention is used because it would otherwise clash with the name of the component.
- * (in this case, component cannot be renamed TreeComponent because of styleguidist constraints)
- */
-export interface ITree {
-  label: string
-  tag?: string
-  childNodes: ITree[]
-}
+import { togglePath, containsPath, getInitialOpenPaths } from "./Tree.utils"
+import { Tree as ITree } from "./Tree.types"
 
 export interface Props {
+  /** An array of tree structures */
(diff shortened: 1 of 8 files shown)


TejasQ:
The code looks incredible so far!

TejasQ:
[Simple expandable tree with 3 levels](https://trello.com/c/hAECaWyR/178-simple-expandable-tree-with-3-levels)

TejasQ:
This fixes #590

micha-f:
@peterszerzo Looks great! Would you be ok with making the colored tags just a little bit smaller?

peterszerzo:
@micha-f I agree - a bit of a shrink on top and bottom padding would do it a lot of good visually. Since we are using the `NameTag` component internally, can we make that change across all UI's?

micha-f:
I am not sure whether this needs to use the name tag. Originally it was supposed to look quite different. More like a square with rounded corners - and only a single letter inside.

image

packages/components/src/Tree/Tree.tsx
```diff
@@ -0,0 +1,150 @@
...
+              />
+            ))}
+          </TreeChildren>
+        )}
+    </TreeContainer>
+  )
+}
+
+class Tree extends React.Component<Props, State> {
+  state: State = {
```
TejasQ:
```diff
- state: State = {
+ readonly state: State = {
```

since state is readonly and it's more explicit.

packages/components/src/Tree/Tree.tsx
```diff
@@ -0,0 +1,150 @@
...
+                openPaths={openPaths}
+              />
+            ))}
+          </TreeChildren>
+        )}
+    </TreeContainer>
+  )
+}
+
+class Tree extends React.Component<Props, State> {
```
TejasQ: You could also

```diff
- class Tree extends React.Component<Props, State> {
+ class Tree extends React.Component<Props, Readonly<State>> {
```

TejasQ:
It may be helpful to add `user-select: none` because I accidentally highlight stuff while clicking on nodes.

peterszerzo:
@micha-f makes sense - I'll take care of adding the square!

TejasQ:
@peterszerzo does this support drag and drop?

peterszerzo:
@TejasQ - yepp, have a look at the code snippet in netlify - one of the nodes has the appropriate props so that it's draggable.

peterszerzo:
@micha-f I updated the design - see updated gif above and the netlify link.

micha-f:
Looks great! The single letter ("D") feels just a little bit too big for me. Maybe we can increase the font-size by one?

packages/components/src/Tree/Tree.utils.ts
```diff
@@ -0,0 +1,37 @@
...
+ * Adds or removes a path from a list of paths.
+ * togglePath([1, 2])([[1, 2, 3], [0]]) -> [[1, 2, 3], [0], [1, 2]]
+ * togglePath([1, 2])([[1, 2], [0]]) -> [[0]]
+ */
+export const togglePath = (path: number[]) => (paths: number[][]): number[][] => {
+  if (paths.length === 0) {
+    return [path]
+  }
+  const [head, ...tail] = paths
+  if (head.join("") === path.join("")) {
```
TejasQ: What is this supposed to do?

packages/components/src/Tree/Tree.utils.ts
```diff
@@ -0,0 +1,37 @@
...
+  }
+  return [head, ...togglePath(path)(tail)]
+}
+
+export const containsPath = (path: number[]) => (paths: number[][]): boolean => {
+  if (paths.length === 0) {
+    return false
+  }
+  const [head, ...tail] = paths
+  if (head.join("") === path.join("")) {
```
TejasQ: You seem to have a bit of duplication here (see line 22).

packages/components/src/Tree/Tree.utils.ts
fabien0102: I bet on an Array compare :-p

peterszerzo: It checks whether a path is the same as another path like [ 0, 1 ] with another [ 0, 1 ]

packages/components/src/Tree/Tree.types.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * ITree naming convention is used because it would otherwise clash with the name of the component.
```
TejasQ: This isn't really needed is it?

PR #615, opened 2018-07-19
Adjust default color of the spinner

TejasQ:
### Summary
The current spinner takes on the text-color of its parent as its fill color, so it's usually black. This PR makes it grey by default according to our design.

### Related issue
https://trello.com/c/OURJ6v33

### To be tested

Fabien

- [x] The spinner looks good
- [x] The car is sexy 🚗 
  

Imogen

- [x] The spinner looks good
- [ ] The car is sexy :car:


TejasQ:
[Default color for spinners should be #747474](https://trello.com/c/OURJ6v33/292-default-color-for-spinners-should-be-747474)

PR #616, opened 2018-07-20
Enhance/new contiamo logo

peterszerzo:
### Summary

Adding the new contiamo logo (well, not exactly new, more like redrawn a bit thicker), and adding opinionated padding and a `to` prop to make it easier to link to the homepage inside a product.

Also removing paddings from `HeaderBar` inner containers so that the click areas on logo links extend to the edges.

screen shot 2018-07-20 at 11 45 40 am

Progress:
* [x] new regular logo with padding
* [x] `to` prop
* [x] `HeaderBar` padding removal
* [x] stacked logo

### To be tested

Tester 1 (Imo)

- [x] Regular logo looks good
- [x] Stacked logo looks good

Tester 2

- [ ] Regular logo looks good
- [ ] Stacked logo looks good


packages/components/src/ContiamoLogo/ContiamoLogo.tsx
```diff
@@ -1,73 +1,87 @@
...
 import styled from "react-emotion"
 import { OperationalStyleConstants, expandColor } from "../utils/constants"
+import { isModifiedEvent } from "../utils"
 import { Icon } from ".."
+import { Consumer, Context } from "../OperationalUI/OperationalUI"
 
 type WithTheme = { theme?: OperationalStyleConstants }
 
-export interface LogoProps {
+export interface Props {
```
TejasQ: I actually prefer the namespaced `Props` interfaces because when we import them in cases where we need them, we `import { LogoProps }` instead of `import { Props as LogoProps }`. The former I find more concise.

packages/components/src/ContiamoLogo/ContiamoLogo.tsx
```diff
@@ -1,73 +1,87 @@
...
-
   /** A color from the constants, or an arbitrary hex value */
   color?: keyof OperationalStyleConstants["color"] | string
-
   /** Do we want it stacked vertically? */
   stack?: boolean
+  /** Link url prop a'la react-router */
+  to?: string
+  /** Click handler */
+  onClick?: (ev: any) => void
```
TejasQ: `MouseEvent`? :D

packages/components/src/ContiamoLogo/ContiamoLogo.tsx
```diff
@@ -1,73 +1,87 @@
...
-  },
-  ({ size, stack }: LogoProps & WithTheme) => ({
-    width: size,
-    display: "flex",
-    flexDirection: stack ? "column" : "row",
-  }),
-)
+const containerStyles = ({
+  stack,
+  color_,
```
TejasQ: What's happening with the `_`?

packages/components/src/HeaderBar/HeaderBar.tsx
```diff
@@ -39,9 +40,11 @@ const Bar = styled("div")(
   },
 )
 
-const StartContainer = styled("div")()
-const CenterContainer = styled("div")()
-const EndContainer = styled("div")({ justifyContent: "flex-end" })
+const StartContainer = styled("div")``
```
TejasQ: You could replace the backticks with `()`. I think it reads more idiomatically for glancers.

packages/components/src/ContiamoLogo/ContiamoLogo.tsx
TejasQ: I agree with your second statement, while I'd argue about the first one: a search for `Props` _should_ also bring up `LogoProps` since the substring matches...

peterszerzo: I mean more like scanning the file visually, or searching for the full word props as `<Props>`.

packages/components/src/ContiamoLogo/ContiamoLogo.tsx
TejasQ: But how do you not end up with `color_` on the div in this case?

TejasQ:
[Fine-tune Pantheon and Labs icons and logos, add to operational and use in standalone UIs.](https://trello.com/c/WwfeOVCc/242-fine-tune-pantheon-and-labs-icons-and-logos-add-to-operational-and-use-in-standalone-uis)

packages/components/src/ContiamoLogo/ContiamoLogo.tsx
peterszerzo: Because that is not a valid html attribute on a div. It would only show up on custom elements.

PR #618, opened 2018-07-20
Add aliases to ProgressPanel to match more usages

fabien0102:
### Summary

Just add some aliases to ProgressPanel to match our current backend convention

### To be tested

Fabien 

Tester 1 - Peter
- [x] The component works with all kind of status

Tester 2 - Tejas

- [x] The component works with all kind of status

### DIFF
--- a/packages/components/src/ProgressPanel/ProgressPanel.tsx
+++ b/packages/components/src/ProgressPanel/ProgressPanel.tsx
@@ -26,27 +26,27 @@ const getVariation = (status: Status, theme: OperationalStyleConstants = default
       return {
         iconColor: theme.color.success,
         textColor: theme.color.text.default,
-        Icon: <Icon left name="Yes" />,
+        icon: <Icon left name="Yes" />,
       }
     case "failure":
     case "failed":
       return {
         iconColor: theme.color.error,
         textColor: theme.color.text.default,
-        Icon: <Icon left name="No" />,
+        icon: <Icon left name="No" />,
       }
     case "waiting":
     case "todo":
       return {
         iconColor: theme.color.text.lightest,
         textColor: theme.color.text.lighter,
-        Icon: <Icon left name="EmptyCircle" />,
+        icon: <Icon left name="EmptyCircle" />,
       }
     case "running":
       return {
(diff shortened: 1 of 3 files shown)


packages/components/src/ProgressPanel/ProgressPanel.tsx
```diff
@@ -19,29 +19,35 @@ export interface Props {
   }[]
 }
 
-const makeIconColor = (status: Status, theme: OperationalStyleConstants) => {
+const getVariation = (status: Status, theme: OperationalStyleConstants = { color: { text: {} } } as any) => {
```
TejasQ: What's this typing? I find it a little hard to understand.

packages/components/src/ProgressPanel/ProgressPanel.tsx
```diff
@@ -75,25 +81,12 @@ const Error = styled("p")`
...
-  }
-}
-
 const ProgressPanel: React.SFC<Props> = props => (
   <Container>
     {props.items.map(({ status, title, error }, index) => (
       <Item key={index}>
         <Body status={status}>
-          <ProgressPanelIcon status={status} />
+          {getVariation(status).Icon}
```
TejasQ: Is there a reason why `Icon` is capitalized?

packages/components/src/ProgressPanel/ProgressPanel.tsx
TejasQ: Yes. 

`import defaultTheme, { OperationalStyleConstants } from '../utils/constants'` at the top and on this line, `theme = defaultTheme`.

packages/components/src/ProgressPanel/ProgressPanel.tsx
TejasQ: I think `<getVariation.Icon` would work but a function invocation in JSX pragmas? That'd be fun to try.

TejasQ:
[Update ProgressPanel API to match Labs](https://trello.com/c/FgGsuzZT/297-update-progresspanel-api-to-match-labs)

PR #620, opened 2018-07-23
Modal redesign.

ImogenF:
### Summary

Redesign of the `Modal` component.

- `title` and `action` props added: passed directly to styled `Card` component.
- `Confirm` component restructured to make use of this

Specifically intended to facilitate: 

<blockquote class="trello-card"><a href="https://trello.com/c/JdshXW0a/274-display-ssh-key-in-bundle-settings">Display ssh key in bundle settings</a></blockquote>

### To be tested

Tester 1

Tester 2

### DIFF
--- a/packages/components/src/Card/Card.tsx
+++ b/packages/components/src/Card/Card.tsx
@@ -35,8 +35,6 @@ const Container = styled("div")(({ theme }: { theme?: OperationalStyleConstants 
 }))
 
 class Card<T = {}> extends React.PureComponent<Props<T>> {
-  containerNode: HTMLElement
-
   static defaultProps = {
     keyFormatter: (title: string) => title,
   }
@@ -51,12 +49,7 @@ class Card<T = {}> extends React.PureComponent<Props<T>> {
     )
 
     return (
-      <Container
-        {...props}
-        innerRef={containerNode => {
-          this.containerNode = containerNode
-        }}
-      >
+      <Container {...props}>
         {(title || action) && <CardHeader title={title} action={action && <this.props.action />} />}
         {data && titles.map((title, i) => <CardItem key={i} value={values[i]} title={title} />)}
         {children}
--- a/packages/components/src/Card/Card.tsx
+++ b/packages/components/src/Card/Card.tsx
@@ -35,6 +35,8 @@ const Container = styled("div")(({ theme }: { theme?: OperationalStyleConstants 
 }))
 
 class Card<T = {}> extends React.PureComponent<Props<T>> {
+  containerNode: HTMLElement
+
   static defaultProps = {
     keyFormatter: (title: string) => title,
   }
@@ -49,7 +51,12 @@ class Card<T = {}> extends React.PureComponent<Props<T>> {
     )
 
     return (
-      <Container {...props}>
+      <Container
+        {...props}
+        innerRef={containerNode => {
+          this.containerNode = containerNode
+        }}
+      >
         {(title || action) && <CardHeader title={title} action={action && <this.props.action />} />}
         {data && titles.map((title, i) => <CardItem key={i} value={values[i]} title={title} />)}
         {children}
--- a/packages/components/src/Confirm/Confirm.tsx
+++ b/packages/components/src/Confirm/Confirm.tsx
@@ -1,37 +1,12 @@
 import * as React from "react"
-import styled, { keyframes } from "react-emotion"
-import Card from "../Card/Card"
+import styled from "react-emotion"
 import { Props as ButtonProps } from "../Button/Button"
-import { OperationalStyleConstants } from "../utils/constants"
-import Overlay from "../Internals/Overlay"
+import Modal from "../Modal/Modal"
 
-const fromBottom = keyframes`
-  0% {
-    top: -10px
-  }
-
-  100% {
-    top: 0
-  }
-`
-
-const Modal = styled(Card)`
-  position: absolute;
-  top: 0;
-  left: 50%;
-  transform: translate(-50%, 0);
-  animation: ${fromBottom} 0.2s;
-  min-width: 600px;
-  min-height: 200px;
-  z-index: ${({ theme }: { theme?: OperationalStyleConstants }) => theme.zIndex.confirm + 1};
-`
-
-const Actions = styled("div")`
-  text-align: right;
-  position: absolute;
-  bottom: 20px;
-  right: 20px;
-`
+const Actions = styled("div")({
+  paddingTop: 20,
+  float: "right",
+})
 
 export interface ConfirmOptions {
   title: React.ReactNode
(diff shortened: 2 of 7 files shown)


packages/components/src/Card/Card.tsx
```diff
@@ -35,6 +35,8 @@ const Container = styled("div")(({ theme }: { theme?: OperationalStyleConstants
 }))
 
 class Card<T = {}> extends React.PureComponent<Props<T>> {
+  containerNode: HTMLElement
```
TejasQ: why? 🤔

packages/components/src/Card/Card.tsx
```diff
@@ -49,7 +51,12 @@ class Card<T = {}> extends React.PureComponent<Props<T>> {
     )
 
     return (
-      <Container {...props}>
+      <Container
+        {...props}
+        innerRef={containerNode => {
```
TejasQ: why?

ImogenF: Needed for https://github.com/contiamo/operational-ui/pull/620/files#diff-cb1950615caaee10e34e8eff41e77b79R42 and https://github.com/contiamo/operational-ui/pull/620/files#diff-cb1950615caaee10e34e8eff41e77b79R59

fabien0102: You can have a look here -> https://reactjs.org/docs/forwarding-refs.html I think that is the correct way to implement this usecase (never used before ^^)

fabien0102: But I know really understand your usecase here

packages/components/src/Modal/README.md
```diff
@@ -11,18 +11,24 @@ class ContentWithModal extends React.Component {
...
         {this.state.isModalOpen ? (
-          <Modal
-            onClose={() => {
-              this.setState(prevState => ({
-                isModalOpen: false,
-              }))
-            }}
-          >
-            <div style={{ width: 300, height: 240 }}>Hello</div>
+          <Modal title="Modal header" onClose={this.onClose.bind(this)}>
```
fabien0102: Can we mimic the `Confirm` API here? so we don't need any state to deal with the `isModalOpen` when we use this component.

packages/components/src/Card/Card.tsx
ImogenF: If this is removed, then the `innerRef` call in `Modal.tsx` returns an instance of the `Card` component, rather than an element.

ImogenF: Oh, just seen your comment @fabien0102. Will take a look

packages/components/src/Modal/README.md
peterszerzo: @fabien0102 what advantages does that bring? You don't need state to deal with `isModalOpen`, the modal simply renders when it should render, and signals to the outside when someone closed it. There is no opinionated multistep process involved, and the current basic API is honest about that.

ImogenF: Also, @fabien0102 , it's possible that a modal may have multiple action buttons, right? So simply having these options: 
```  
cancelButton?: React.ReactElement<ButtonProps>
actionButton?: React.ReactElement<ButtonProps>
onConfirm?: () => void
onCancel?: () => void
``` 

wouldn't be generic enough.

peterszerzo: The modal component should be able to have anything inside it, otherwise it wouldn't really be truthful to its name. It is also a solid building block for other opinionated components, for which `Confirm` is a great example.

TejasQ: I agree – actions on a Modal seems a bit overkill for a simple Modal dialog.

ImogenF:
[Redesign modal component](https://trello.com/c/MwOBVb5p/312-redesign-modal-component)

packages/components/src/Modal/Modal.tsx
```diff
@@ -1,65 +1,41 @@
...
+
 import Overlay from "../Internals/Overlay"
+import Card, { Props as CardProps } from "../Card/Card"
 
 export interface Props {
   id?: string
   className?: string
   /** Content class name */
-
   contentClassName?: string
```
TejasQ: This could be renamed to `className` as one would expect any given `className` to be applied to the Modal and not necessarily to an overlay.

packages/components/src/Confirm/Confirm.tsx
```diff
@@ -74,27 +49,22 @@ export class Confirm extends React.Component<Props, Readonly<State>> {
...
     const isOpen = Boolean(this.state.options)
 
-    if (isOpen) {
-      const { title, body, cancelButton, actionButton } = this.state.options
-      return (
-        <>
-          <Modal title={title}>
-            {body}
+    return (
+      <>
```
TejasQ: If we didn't remove the destructuring, we wouldn't need to repeat `this.state.` and `this.props.` throughout this file.

Was there a good reason to remove it?

packages/components/src/Confirm/Confirm.tsx
```diff
@@ -1,37 +1,12 @@
...
-`
-
-const Actions = styled("div")`
-  text-align: right;
-  position: absolute;
-  bottom: 20px;
-  right: 20px;
-`
+const Actions = styled("div")({
+  paddingTop: 20,
```
TejasQ: `theme.space.element` I believe. Further, is the intent here to space the actions from the content? I think a `marginTop` would be more applicable here.

ImogenF:
@TejasQ agreed that the props section should be removed. The rest of it is fine though, right? The code itself I've been updating as I updated the component

packages/components/src/Modal/README.md
```diff
@@ -11,18 +11,24 @@ class ContentWithModal extends React.Component {
...
+            <p>Modal content</p>
+            Any <Icon name="OperationalUI" size={16} /> components or HTML elements can be rendered here.
+            <div style={{ position: "absolute", bottom: 20, right: 20 }}>
```
TejasQ: Even if this is an example, I think this styling could be better. :)

ImogenF: Can you be more specific please?

TejasQ: As far as I can tell, the intent of this `div` is to place a button _under_ the content – not to remove the button from the layout or place it _above or below_ a certain z-index in float-order.

If we want to simply place a `div` under another `div`, I'm saying `position: absolute` is not the correct approach.

Consider instead

```jsx
<div style={{ marginTop: 16, display: "flex", justifyContent: "flex-end" }}>
```

or a similar variant.

ImogenF: The intent is to place the button in the bottom right corner of the modal, regardless of whether the rest of the content fills up the full height of the modal or not.

packages/components/src/Confirm/Confirm.tsx
ImogenF: If the destructuring is before the return statement, outside of the `{isOpen && ...}` block, then if the modal is not open, there is no `this.state.options` to destructure and an error is thrown.

packages/components/src/Modal/Modal.tsx
peterszerzo: There is already a `className` right above - this was supposed to work on the container of the modal content, not the overall modal content, and not the overlay. I think there's still a use-case for this, even with a `dangerousClassName` thing, at least until we have a strategy to style modal contents.

packages/components/src/Modal/README.md
TejasQ: I don't understand. 🤔 

What is the full height of the Modal? Could you please clarify?

TejasQ:
[Isolate Operational Components into Single Repo](https://trello.com/c/cM6xlQjB/212-isolate-operational-components-into-single-repo)

packages/components/src/Modal/README.md
ImogenF: The `Modal` component has a `minHeight` property. If there is very little content inside the modal, then the styling you suggested would result in the button being halfway up the modal, and not at the bottom

TejasQ: Wow! I missed the `minHeight` property. Is there a reason we have that restriction?

ImogenF: The idea was that if a modal is created with no or very little content, it wouldn't be a ridiculously small box. But I can take it out (or greatly reduce the numbers).

packages/components/src/Modal/Modal.tsx
```diff
@@ -1,65 +1,39 @@
...
-    backgroundColor: theme.deprecated.colors.white,
-    padding: theme.deprecated.spacing,
-    boxShadow: theme.deprecated.shadows.popup,
-  }),
+const Content = styled(Card)(({ theme }: { theme?: OperationalStyleConstants }) => ({
+  position: "fixed",
+  top: "50%",
+  left: "50%",
+  transform: "translate(-50%, -50%)",
+  zIndex: theme.zIndex.confirm + 1,
```
TejasQ: We should probably rename `theme.zIndex.modal` -> `theme.zIndex.modal`.

packages/components/src/Modal/README.md
TejasQ: Yeah, take it out. Whoever wants a Modal should be responsible for its contents.

ImogenF: Already gone!

PR #623, opened 2018-07-24
Fix layout main overflow

peterszerzo:
### Summary

Removes double scrollbars by setting the `Layout` component's main section's overflow to hidden. Overflow setting is the responsibility of the child component, which `Page` takes care of already.

### To be tested

Tester 1

- [x] Layout overflow as expected

Tester 2

- [ ] Layout overflow as expected

### DIFF
--- a/packages/components/src/Layout/Layout.tsx
+++ b/packages/components/src/Layout/Layout.tsx
@@ -43,7 +43,7 @@ const GridContainer = styled("div")(
 )
 
 const Main = styled("div")(({ theme }: { theme?: OperationalStyleConstants }) => ({
-  overflow: "auto",
+  overflow: "hidden",
   backgroundColor: theme.color.white,
   gridColumnStart: "2",
   gridColumnEnd: "span 1",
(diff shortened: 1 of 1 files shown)


packages/components/src/Layout/Layout.tsx
```diff
@@ -43,7 +43,7 @@ const GridContainer = styled("div")(
 )
 
 const Main = styled("div")(({ theme }: { theme?: OperationalStyleConstants }) => ({
-  overflow: "auto",
+  overflow: "hidden",
```
TejasQ: 🤔

TejasQ:
Thanks, Peter! It looks good, although confusing. Why does `overflow: hidden` fix things?

While you're at it, there's a double scrollbar situation over [on Labs](https://labs.fe.ctmo.io/4c495436-4973-5138-bb06-7a304df61932/bundles/630d8b33-1802-45f6-baaf-261310e55c6a/overview/). You should see two scrollbars on the right of your browser.

Can we fix this too?

PR #628, opened 2018-07-26
Modal style

fabien0102:
### Summary

image

```tsx
// Simple page with `main` layout
<Page>{({modal, confirm}) => <div/>}</Page>
```

```tsx
// With tabs
const Tab = n => () => (
  <PageContent>
    <PageArea>
      {({ confirm, modal }) => <div />}
    </PageArea>
  </PageContent>
)
;<Page
  title="Bundle detail"
  tabs={[
    { name: "overview", component: Tab("overview") },
    { name: "jobs", component: Tab("jobs") },
    { name: "functions", component: Tab("functions") },
  ]}
/>
```

### Related issue

#626 

### To be tested

Fabien 

Tester 1

- [ ] The modal is on top center of the page as the design specs
- [ ] The confirm is on top center of the page as the design specs
- [ ] All the callbacks are working as expected
- [ ] No regression on Page
- [ ] No regression on PageArea
- [ ] I can create/play with `confirm` and `modal` into `Page` and `PageArea`

Tester 2

- [ ] The modal is on top center of the page as the design specs
- [ ] The confirm is on top center of the page as the design specs
- [ ] All the callbacks are working as expected
- [ ] No regression on Page
- [ ] No regression on PageArea
- [ ] I can create/play with `confirm` and `modal` into `Page` and `PageArea`


src/Confirm/README.md
```diff
@@ -1,43 +1,3 @@
-This component is a confirm modal that should use on any risky user actions (delete something for example).
+This component is deprecated in his standalone mode.
```
TejasQ: You can just remove this.

src/Internals/ControlledModal.tsx
```diff
@@ -15,11 +15,23 @@ export interface Props {
   action?: CardProps["action"]
 }
 
+const fromBottom = keyframes`
+  0% {
+    top: -10px
```
TejasQ: Can you animate `transform: translateY(-10px)` instead? It is more performant.

src/Modal/README.md
```diff
@@ -1,30 +1,3 @@
-Modals are customizable full-screen alert boxes. They should be used sparingly, but they come in handy when there is a legitimate reason to block the rest of the screen.
+This component is deprecated in his standalone mode.
```
TejasQ: Well, remove it then.

src/Page/Page.tsx
```diff
@@ -1,35 +1,49 @@
...
+  /** Areas template for `PageArea` disposition */
+  areas?: "main"
+  /** Fill the entire width */
+  fill?: boolean
+  tabs?: never
+  activeTabName?: never
+  onTabChange?: never
+}
+
+export interface PropsWithComplexePage extends BaseProps {
```
TejasQ: `Complexe` -> `Complex`

src/Confirm/README.md
fabien0102: It was to avoid a breaking change for now, but I can remove this if you prefer

fabien0102:
@TejasQ I'm not sure about documenting the modal position, it was documented before due to usage of the `Confirm` component (that was really a restriction to be the first child of `PageArea` to have the correct design) but now it's just an implementation detail since we can't use anymore `Confirm` and `Modal` outside of `PageArea` with our new API design

fabien0102:
@TejasQ mmmmm, good catch, this mean that if we have a modal on a bottom of a page, we can't see the modal… 

@micha-f Some design recommandations for this behaviour? 😃

fabien0102:
From an offline conversation with @micha-f :

We decide to put the Modal center to content of `Page` and always `20px` from the top.

Means that I need to reimplement everything, but it happens. I will probably use a `React.Context` to be able to have the `confirm` and `modal` in every layers possible, to have the best flexibility as possible 😈

TejasQ:
Always `20px` from the top of the page? What happens if your scroll position is at the bottom?

TejasQ:
[Fix style/API for Modal/Confirm](https://trello.com/c/dsgvISzg/330-fix-style-api-for-modal-confirm)

src/Page/Page.tsx
```diff
@@ -120,59 +158,62 @@ class Page extends React.Component<Props, Readonly<typeof initialState>> {
     return activeTab
   }
 
+  private renderPageWithTabs() {
+    const { tabs } = this.props
+    if (tabs) {
```
TejasQ: This this make sense in light of [line 213](https://github.com/contiamo/operational-ui/pull/628/files#diff-88ff81fac534225deda44f519755137cR213) since there's already a check there?

src/Page/README.md
```diff
@@ -258,36 +258,158 @@ const actions = (
...
-    )}
-  </Page>
-</OperationalUI>
+```jsx
+<Page title="With a modal">
+  {({ confirm, modal }) => (
+    <Card
+      action={() => (
+        <div>
+          A nice text generated by <a href="https://baconipsum.com/">https://baconipsum.com/</a>
```
TejasQ: Needs bacon emoji. 🥓

src/Page/Page.tsx
```diff
@@ -1,34 +1,57 @@
...
-import Confirm, { ConfirmOptions } from "../Confirm/Confirm"
-import Modal, { ModalOptions } from "../Modal/Modal"
+import Confirm, { ConfirmOptions } from "../Internals/Confirm"
+import Modal, { ModalOptions } from "../Internals/Modal"
 import PageArea from "../PageArea/PageArea"
 import PageContent from "../PageContent/PageContent"
 import styled from "../utils/styled"
 
-export type Tabs = Array<{ name: string; component: React.ComponentType; hidden?: boolean }>
+export type Tabs<T extends { __modalConfirmContext?: ModalConfirmContext } = any> = Array<{
```
peterszerzo: Add a comment here that makes it clear that tabs children can only be `PageContent` (the existence of the `__modalConfirmContext` prop makes this quite implicit). We can also name this `__pageContentModalConfirmContext` or something to not rely only on comments.

TejasQ:
[Display ssh key in bundle settings](https://trello.com/c/JdshXW0a/274-display-ssh-key-in-bundle-settings)

src/Page/Page.tsx
fabien0102: It's for typescript discrimination…

peterszerzo:
@ImogenF I like the general direction of separating out page types into their own components - makes the intent clearer simpler and implementations easier to follow (same goes for `PageAreas`). Since we're not running into any significant issues at the moment on the page-with-tabs/page-without-tabs business (do correct me if I'm wrong), I say we keep the public API like this for now and keep our eyes peeled for a good opportunity to separate?

PR #633, opened 2018-07-30
BREAKING: Allow only react elements in page tabs

peterszerzo:
### Summary

This PR ~allows for~ mandates that only simple react elements to be passed in a `Page` component's tab object in order to prevent node rebuild behavior that loses focus- and other internal state in tab children. In both of the following code snippets, focus state is not maintained on the `Input` component (see added temporary styleguidist snippet to experiment, in line with the gif below):

```jsx
<Page
  title="Bundle detail"
  activeTabName={this.state.tabName}
  onTabChange={this.goTo.bind(this)}
  tabs={[
    {
      name: "overview",
      component: (props) => (
        <PageContent {...props}>
          <Input value={this.state.value} onChange={newValue => {
            this.setState(() => ({ value: newValue }))
          }} />
        </PageContent>
      )
    },
  ]}
/>
```

```jsx
<Page
  title="Bundle detail"
  activeTabName={this.state.tabName}
  onTabChange={this.goTo.bind(this)}
  tabs={[
    {
      name: "overview",
      component: (props) => (
        <Input value={this.state.value} onChange={newValue => {
          this.setState(() => ({ value: newValue }))
        }} />
      )
    },
  ]}
/>
```

tab-focus-loss

Having the client create full-fledged React components to have this simple example working forces needless atomization and parent-child communication boilerplate, which this feature is looking to avoid.

### To be tested

Tester 1

- [x] Tabs interactions work as before

Tester 2

- [ ] Tabs interactions work as before


TejasQ:
I think this could fix https://github.com/contiamo/labs-ui/issues/81 actually too possibly. 🤔

peterszerzo:
@TejasQ @fabien0102 it is possible that it contributes to the Labs UI issue, and I finally understand why: in the code snippets above, a new `() => <TabContents />` function is generated on every render that is not referentially equal to the previous function. So even though the diffing algorithm should look at the internal virtual dom nodes and diff on the tab contents level, it stops right at the tabs child container because the node types aren't equal (same as if I turned a `div` tag into an `a` tag).

To avoid this kind of behavior, and given that arrow function usage make it into the Labs UI codebase just because the API allows it, we should only allow react nodes to be passed here and guarantee that dom diffing will always work as expected. It's a really intricate bug to pin down (for my brain at least) and I'd rather save my colleagues the effort in the future. It should allow for the same flexibility, even if some preferences for sugar might be gone. But safety first..

src/Page/Page.tsx
```diff
@@ -6,7 +6,7 @@ import PageArea from "../PageArea/PageArea"
 import PageContent from "../PageContent/PageContent"
 import styled from "../utils/styled"
 
-export type Tabs = Array<{ name: string; component: React.ComponentType; hidden?: boolean }>
+export type Tabs = Array<{ name: string; content: React.ReactNode; hidden?: boolean }>
```
TejasQ: Can we name this `children` to be more semantic?

ImogenF:
I was just about to comment on this to say that the styleguidist example for pages with hidden tabs seems to not be working correctly. Don't think it's a problem with the component though, just the snippet.

PR #631, opened 2018-07-27
BREAKING: Simplify Loading Progress & Add to Context

peterszerzo:
### Summary

Adding window size- and loading state management to operational context:

```
<OperationalUI>
  <OperationalContext>
    {ctx => (
      <div style={{ width: ctx.windowSize.width / 2 }}>
        <Button onClick={() => ctx.setLoading(true)} />
      </div>
    )}
  </OperationalContext>
</OperationalUI>
```

loader

This removes the need for window size tracking for different apps (some client projects + Pantheon), helps the internal tooltip implementation, and allows the top progress bar to be controlled downstream. Granted, the top bar solution is on the imperative side, and we can wind up with multiple components competing in setting the progress bar state, although alternative solutions with portals et. al. can present similar issues. Let me know in the discussion below if you have ideas for improvement.

### To be tested

Tester 1

- [ ] No error/warning in the console
- [ ] Window size is tracked
- [ ] Loading state manipulation works
  

Tester 2

- [ ] No error/warning in the console
- [ ] Window size is tracked
- [ ] Loading state manipulation works

PR #635, opened 2018-07-30
Add OperationalContext as its own component

peterszerzo:
This is a prototype in relation to #631, looking at making `OperationalContext` into its own component folder so examples are cleaner to separate. Styleguidist is giving me a hard time putting it on the site, though, mind taking a peek at it @fabien0102?


src/OperationalContext/README.md
```diff
@@ -0,0 +1,80 @@
+The `OperationalContext` component provides utility methods and data that can be used inside `OperationalUI`:
+
+### Window size
+
+Child components can get access to window dimensions as follows:
+
+```jsx
+<OperationalUI>
+  <OperationalContext>
```
TejasQ: I'm not sure the preceding line is necessary because everything here _should_ be already wrapped with `OperationalUI` per the styleguidist config.

src/OperationalContext/README.md
```diff
@@ -0,0 +1,80 @@
...
+  </OperationalContext>
+</OperationalUI>
+```
+
+### Messages and Loaders
+
+You can use `OperationalUI`'s flash- and progress bar features to automatically render and manage these universal UI elements using the `pushMessage` and `setLoadingState` methods provided in context, as shown in the code snippet below:
+
+```jsx
+class MessageHandler extends React.Component {
```
TejasQ: Is there a good reason for this to be a class component?

src/OperationalContext/README.md
peterszerzo: Good point..

src/OperationalContext/README.md
peterszerzo: Nope :)

PR #644, opened 2018-07-31
Enable custom starting textarea height.

ImogenF:
### Summary

Previously: Textarea width can be controlled by using the `fullWidth` flag, but the height always started at the `minHeight` value of 120. 

Now: Textarea has a `height` prop which sets the starting height of the component. `minHeight` remains the same, and resizing is not affected.

Needed for <blockquote class="trello-card"><a href="https://trello.com/c/JdshXW0a/274-display-ssh-key-in-bundle-settings">Display ssh key in bundle settings</a></blockquote>

### To be tested

Tester 1

- [x] `height` prop works and doesn't break resizing or affect minimum height
  

Tester 2

- [x] `height` prop works and doesn't break resizing or affect minimum height

### DIFF
--- a/src/Textarea/README.md
+++ b/src/Textarea/README.md
@@ -12,6 +12,7 @@ class MyForm extends React.Component {
       v5: "",
       v6: "",
       v7: "",
+      v8: "",
     }
     this.handleChange = key => value => {
       this.setState({ [key]: value })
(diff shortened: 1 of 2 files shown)


ImogenF:
[Display ssh key in bundle settings](https://trello.com/c/JdshXW0a/274-display-ssh-key-in-bundle-settings)

PR #647, opened 2018-07-31
Feature: Splash Screen

peterszerzo:
# Summary

This PR generalizes and exposes the `Splash` component so that `operational-visualizations` can get their own similar website, like so:

```jsx
// This is not in the regular exports, only available under a subfolder
import Splash from "@operational/components/website"

const VizWebsite = props => (
  <Splash
    title="Operational Visualizations"
    color="#88AB12"
    actions={<><Button>Button1</Button></>}
  >
    <p>Visualizations body</p>
  </Splash>
)
```
splash

## Error Boundary
This PR also adds a catch-all error boundary above all contained Operational components – at the point where this error boundary is invoked, the app is completely unmounted. Because of this, we have a choice: either we unmount the whole app, or tell users to implement more localized error boundaries with a handy docs link. 

### Screens
Production:
Production

Development:
Development

This PR aims to make Operational a little more developer friendly.

In a production environment, the error is a little more obscured.

# Related Issue
https://trello.com/c/mil1pcAI

# To be tested

Peter

Tester 1

- [ ] Can configure `Splash` in `styleguide/StyleGuideRenderer.tsx` to create a prototype for the visualizations website
- [ ] **Error Boundary Looks good in production**: in any playground on the Netlify-deployed styleguide, enter `window.haha.hoho` to trigger an error.
- [ ] **Error Boundary Looks good in development**: run locally and in any playground on the styleguide, enter `window.haha.hoho` to trigger an error.

Tester 2

- [ ] Can configure `Splash` in `styleguide/StyleGuideRenderer.tsx` to create a prototype for the visualizations website
- [ ] **Error Boundary Looks good in production**: in any playground on the Netlify-deployed styleguide, enter `window.haha.hoho` to trigger an error.
- [ ] **Error Boundary Looks good in development**: run locally and in any playground on the styleguide, enter `window.haha.hoho` to trigger an error.


TejasQ:
Can we have `Splash` just be a component?

src/Splash/README.md
```diff
@@ -0,0 +1,16 @@
+The splash component can be used to render a simple animated splash screen for a project. It is used internally by this package and by https://github.com/contiamo/operational-visualizations.
+
+### Usage
+
+```jsx static
```
TejasQ: It would be cooler to actually have a button toggler and use it in the styleguidist page. Also meta.

peterszerzo:
@TejasQ I added the interactive example. Logo customization will be part of a future improvement as per #650.

src/Splash/README.md
```diff
@@ -0,0 +1,67 @@
+The splash component can be used to render a simple animated splash screen for a project. It is used internally by this package and by https://github.com/contiamo/operational-visualizations.
+
+### Usage
+
+```jsx
+const Dismiss = styled.default("div")({
```
TejasQ: For the example, could this be more concise?

```jsx
const initialState = { open: false };

const DismissButton = styled.default(Button)({
  position: "fixed",
  zIndex: "1000000",
  top: 20,
  right: 20,
});

<>
  {state.isOpen && (
    <>
      <DismissButton
          icon="No"
          onClick={() => setState({ isOpen: false })}
        >
          Dismiss
      </DismissButton>
      <Splash
        title="My Interesting Project"
        actions={<Button>Action</Button>}
      >
        <p>I made a software project. It is interesting for the following reasons:</p>
      </Splash>
    </>
  )}
  <Button
    color="primary"
    onClick={() => setState({ isOpen: true })}
  >
    Open splash
   </Button>
</>;
```

src/Splash/Splash.tsx
```diff
@@ -1,32 +1,43 @@
...
   left: 0px;
   right: 0px;
   bottom: 0px;
   padding: 10px;
-  background-color: #005f96;
   display: flex;
   align-items: center;
   justify-content: center;
   transition: opacity 0.2s;
+  ${({ color_, theme }) => {
```
TejasQ: Since there's just a return here, maybe we strip the `{}` from the lambda.

fabien0102:
Why the `Dismiss` button is on the top-right position? I think that is make sense to be the close as possible of our real use case ;)

src/Splash/README.md
fabien0102: I think that we can remove this `DismissButton` totally and put a button into `actions` instead

peterszerzo:
@fabien0102 it is in the corner to not interfere with the looks of the page.

src/Splash/README.md
peterszerzo: I think that mixes things up a bit - the center of the page should demo what this component does, and not how its popup container should be controlled.

peterszerzo:
@TejasQ I don't think that's a good idea (see explanation in previous responses to @fabien0102's comments)

src/Splash/README.md
TejasQ: I disagree. Since it's an _example_ in the style guide, it ought to fully demonstrate the use case: the use case does _not_ include having some floating action button somewhere to dismiss it (see the _actual_ Splash screen), but instead has actions where actions are.

TejasQ:
@peterszerzo sorry, missed that. [Responded](https://github.com/contiamo/operational-ui/pull/647#discussion_r207129829).

src/Splash/Splash.tsx
```diff
@@ -1,32 +1,41 @@
...
+import styled from "../utils/styled"
+import Animation from "./Splash.Animation"
+import OperationalLogo from "./Splash.Logo"
 
 export interface Props {
-  hide: () => void
+  /** The title of the project */
+  title: string
+  /** Actions displayed below the title, typically a fragment of `<Button/>` elements */
+  actions: React.ReactNode
```
TejasQ: This should be optional.

src/Splash/README.md
peterszerzo: Ok, let's do it that way.

peterszerzo:
@TejasQ see #650 for logo.

src/Splash/Splash.tsx
```diff
@@ -1,32 +1,41 @@
...
   right: 0px;
   bottom: 0px;
   padding: 10px;
-  background-color: #005f96;
   display: flex;
   align-items: center;
   justify-content: center;
   transition: opacity 0.2s;
+  ${({ color_, theme }) => `
+      background-color: ${expandColor(theme, color_) || theme.color.primary};
```
TejasQ: We do this, but we don't set a `readableTextColor`. What happens if it's white?

src/Splash/README.md
```diff
@@ -2,15 +2,42 @@ The splash component can be used to render a simple animated splash screen for a
...
+  isOpen: false,
+}
+;<>
+  {state.isOpen && (
     <>
-      <Button>Action</Button>
+      <Splash
+        title="My Interesting Project"
+        actions={
+          <>
```
TejasQ: This fragment is not needed.

src/Splash/README.md
```diff
@@ -2,15 +2,42 @@ The splash component can be used to render a simple animated splash screen for a
...
+            <Button
+              onClick={() => {
+                setState(() => ({
```
TejasQ: The function callback for setState bloats the example.

src/Splash/Splash.tsx
peterszerzo: This is intentional - there should be a call-to-action element on a splash screen, otherwise it is both visually unappealing and would take a UX hit.

src/Splash/README.md
peterszerzo: Do you refer to the `setState(() => ({}))` instead of `setState({})`? I wouldn't ever want to write it in the latter form, due to possible bugs in the render loop.

src/Splash/Splash.tsx
peterszerzo: I haven't really thought about how things will look on a too bright background - I'll add it to the docs that a dark enough color is expected for the `color` prop, and add it as an optimization to #650.

src/Splash/README.md
TejasQ: Could you explain a little bit more about possible bugs in the render loop?

peterszerzo: When two `setState` calls happen very close to each other, it is possible that both changes are applied based off the state value from before either of the changes happened, flat-out discarding one of the changes.

Of course, this only happens when the call uses the previous state value, which is not the case here, but I prefer to always use a function.

TejasQ: Since we're not relying on previous state updates here, and _especially_ since state is so sandboxed in this playground, I think it's a bit overkill. But leave it in if you feel strongly about it. 🤷‍♂️

peterszerzo: Ok, I'll do that then.

src/Splash/README.md
```diff
@@ -0,0 +1,39 @@
+The splash component can be used to render a simple animated splash screen for a project. It is used internally by this package and by https://github.com/contiamo/operational-visualizations.
+
+### Usage
+
+```jsx
+const initialState = {
+  isOpen: false,
+}
+;<>
```
TejasQ: We can get rid of the fragment by turning the next line into a ternary.

PR #654, opened 2018-08-01
Improve confirm api

fabien0102:
### Summary

image

To be able to deal with this kind of use case, we need a way to add some internal states to the `body`. This is the topic of this PR!

The proposed API:
```tsx
export interface ConfirmOptions<T = {}> {
  title: React.ReactNode
  body: React.ReactNode | React.ComponentType<{ setConfirmState: (confirmState?: Pick<T, keyof T>) => void; state?: T }>
  cancelButton?: React.ReactElement<ButtonProps>
  actionButton?: React.ReactElement<ButtonProps>
  onConfirm?: (confirmState?: T) => void
  onCancel?: (confirmState?: T) => void
  state?: T
}
```

So I can give a `React.Component` as body and interact with `this.state.options.state` inside the `body`.

This `this.state.options.state` is propagated to `onConfirm` and `onCancel` to be able to do any kind of custom logics.

I can also give an initial state into `state`, to finalize the loop 😃 

Note: This API can also be useful for `enter the project name to confirm` validation pattern (as github do for very risky operations)

### Demo :tada:

https://deploy-preview-654--operational-ui.netlify.com/#!/Page/27

### Related issue

https://trello.com/c/EKedZwMG

### To be tested

Me
- [x] The typescript signature works perfectly

Tester 1

- [ ] Every modal and confirm are working as before
- [ ] I can inject/modify and retrieve a custom state with the new API (#!/Page/27)

Tester 2

- [ ] Every modal and confirm are working as before
- [ ] I can inject/modify and retrieve a custom state with the new API (#!/Page/27)


src/Internals/Confirm.tsx
```diff
@@ -4,20 +4,18 @@ import styled from "../utils/styled"
...
   title: React.ReactNode
-  body: React.ReactNode
+  body: React.ReactNode | React.ComponentType<{ setConfirmState: (state?: Pick<T, keyof T>) => void; state?: T }>
   cancelButton?: React.ReactElement<ButtonProps>
   actionButton?: React.ReactElement<ButtonProps>
-  onConfirm?: () => void
-  onCancel?: () => void
+  onConfirm?: (confirmState?: T) => void
+  onCancel?: (confirmState?: T) => void
+  state?: T
```
TejasQ: This should probably be called `initialState` unless I've missed something with brain compilation.

src/Internals/Confirm.tsx
```diff
@@ -33,28 +31,40 @@ export class Confirm extends React.Component<Props, Readonly<State>> {
...
+      this.state.options.onConfirm(this.state.options.state)
     }
     this.closeConfirm()
   }
 
+  private setConfirmState = (state: any) => {
+    this.setState(prevState => ({
+      options: {
+        ...prevState.options,
+        state: {
```
TejasQ: `state`, `options.state`, `this.state`, `state` is a function argument, how about we name things a _little_ more specifically?

src/Typography/Body.tsx
```diff
@@ -8,7 +8,7 @@ export const Body = styled("p")<{
   fontFamily: theme.font.family.main,
   fontWeight: theme.font.weight.regular,
   lineHeight: theme.font.lineHeight,
-  margin: `${theme.space.base}px 0`,
+  margin: `0 0 ${theme.space.element}px 0`,
```
TejasQ: Or, consider `marginBottom: theme.space.element`.

src/Internals/Confirm.tsx
```diff
@@ -4,20 +4,18 @@ import styled from "../utils/styled"
...
-  textAlign: "right",
-  bottom: theme.space.element,
-  right: theme.space.element,
-}))
+const Actions = styled("div")({
+  alignSelf: "flex-end",
+})
 
-export interface ConfirmOptions {
+export interface ConfirmOptions<T = {}> {
```
peterszerzo: Propagate the `T` generic into the component's props so `setConfirmState` can have accurate typings.

src/Internals/Confirm.tsx
```diff
@@ -33,28 +31,40 @@ export class Confirm extends React.Component<Props, Readonly<State>> {
...
-  public onActionClick = () => {
+  private onActionClick = () => {
     if (this.state.options.onConfirm) {
-      this.state.options.onConfirm()
+      this.state.options.onConfirm(this.state.options.state)
     }
     this.closeConfirm()
   }
 
+  private setConfirmState = (state: any) => {
```
peterszerzo: Support async `setState` so the API is the same as React's `setState` (this is probably very inaccurate, read with a grain of salt :) ):

```ts
setConfirmState = (state: Partial<T> | ((prevState: T) => Partial<T>))
```

```js
this.setState(prevState => ({
      options: {
        ...prevState.options,
        state: {
          ...prevState.options.state,
          ...(typeof state === "function" ? state(prevState.options.state) : state),
        },
      },
    }))
```

src/Internals/Confirm.tsx
peterszerzo: Consider adding this API extension to the `Modal` component instead. This way it is kind of losing the semantic value of `Confirm`, and `Modal` is intended to be the more generic/flexible cousin of this component anyway.

src/Typography/Body.tsx
fabien0102: Here I want to be sure that I don't have any other margin ;)

src/Internals/Confirm.tsx
fabien0102: This will be done later #657 :wink:

styleguide/StyleGuideRenderer.tsx
```diff
@@ -69,7 +69,7 @@ class StyleGuideRenderer extends React.Component<StyleGuideRendererProps, Readon
 
   private updateActiveComponent = (activeComponent: string) => {
     if (this.props.hasSidebar) {
-      window.history.pushState(null, undefined, `#${activeComponent}`)
+      window.history.pushState(null, "", `#${activeComponent}`)
```
TejasQ: Why?

Issue #629, opened 2018-07-27
Card typings for values


TejasQ:
Currently the Card component has a really nice functionality that will allow you to give it `data`, and then access/filter/format the `data` in your card as needed.

## Demo

![08b6fafb-e2c1-427d-9f1d-f66d8c3d86f0](https://user-images.githubusercontent.com/1761469/43310914-38388c44-9189-11e8-8e50-275fbd1ef707.gif)

## [Example](http://operational-ui.netlify.com/#!/Card/5)

[The typing for this](https://github.com/contiamo/operational-ui/pull/627/files#diff-a35de8bc4180536d7c718152b396c178R48) is a little bit sketchy and currently does not satisfy the typescript compiler.

This issue aims to adjust this typing.

stereobooster:
```ts
type A<T extends { [key: string]: any }> =
    { [P in keyof T]?: (value: T[P]) => null };
let x: A<{ a: string, b: number, c: symbol }>;
x.a("a")
x.b(1)
x.c(Symbol(1))
```

TejasQ:
Thanks, @stereobooster! Want to submit a PR?

TejasQ:
Fixed in #655

PR #655, opened 2018-08-01
[#629] Better types for Card component

stereobooster:
### Summary

PR improves types for Card component. Not TS can deduce params of `valueFormatter` based on provided data. As you can see in screenshot for `a` it asks `number | void`, for `b` it asks `string | void`

screen shot 2018-08-01 at 21 23 08
screen shot 2018-08-01 at 21 23 28

Caveats: 
- I didn't find how to do this trick with React Component type, because a lot of other function do not expect parametric React Component Class and start to complain about incompatibility. That is why I switched to function component
- `data` is marked as optional param and there is separate `keys` param, that is why type of `valueFormatter` contains `void`

### Related issue

https://github.com/contiamo/operational-ui/issues/629

### To be tested

Me

Tester 1

Tester 2


src/Card/Card.tsx
```diff
@@ -34,33 +34,34 @@ const Container = styled("div")(({ theme }) => ({
   },
 }))
 
-class Card<T = {}> extends React.PureComponent<Props<T>> {
-  public static defaultProps = {
-    keyFormatter: (title: string) => title,
-  }
+const objectKeys = <T extends {}>(x: T): keyof T => {
+  // @ts-ignore
```
TejasQ: @stereobooster why?

src/Card/Card.tsx
```diff
@@ -34,33 +34,34 @@ const Container = styled("div")(({ theme }) => ({
...
 
-    return (
-      <Container {...props}>
-        {(title || Action) && <CardHeader title={title} action={Action && <Action />} />}
-        {data && titles.map((cardItemTitle, i) => <CardItem key={i} value={values[i]} title={cardItemTitle} />)}
-        {children}
-      </Container>
-    )
-  }
+// @ts-ignore
```
TejasQ: @stereobooster why?

src/Card/Card.tsx
fabien0102: The correct version:

```js
const getKeys = <T extends {}>(o: T) => Object.keys(o) as Array<keyof T>
```

The function should return `Array<keyof T>` instead of `keyof T`, this is why typescript is not happy :wink:

stereobooster: Right

src/Card/Card.tsx
```diff
@@ -34,33 +34,34 @@ const Container = styled("div")(({ theme }) => ({
...
+Card.defaultProps = {
```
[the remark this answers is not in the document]
stereobooster: I removed default props because it makes no difference in this case, but also possible to do it other way

stereobooster:
@fabien0102 I tested it

```
const data = {
  a: "toto",
  b: "titi",
  c: "riri"
}

const myCard = <Card
  data={data}
  keyFormatter={val => val} 
  valueFormatters={{a: (val: string) => val}} // no error I checked it
/>
```

src/Card/Card.tsx
```diff
@@ -34,33 +36,25 @@ const Container = styled("div")(({ theme }) => ({
...
 }))
 
-class Card<T = {}> extends React.PureComponent<CardProps<T>> {
-  public static defaultProps = {
-    keyFormatter: (title: string) => title,
-  }
-
-  public render() {
-    const { title, keyFormatter, valueFormatters = {}, data, keys, children, action: Action, ...props } = this.props
+const objectKeys = <T extends {}>(x: T) => Object.keys(x) as Array<keyof T>
```
fabien0102: Nice 👍 I think that we can extract this function into `utils`, it's a commun case.

src/Card/Card.tsx
```diff
@@ -3,13 +3,13 @@ import { CardHeader, CardItem } from "../"
...
-export interface CardProps<T = {}> extends DefaultProps {
+export interface CardProps<T extends {} = {}> extends DefaultProps {
   /** Any object to show. The key is the title of the data. */
   data?: T
   /**  A function to format keys of `data` */
-  keyFormatter?: (key: string) => string
+  keyFormatter?: (key: Extract<keyof T, string>) => string
   /** A key-value object to format values of `data`. */
-  valueFormatters?: { [P in Extract<keyof T, string>]?: (value: string) => React.ReactNode }
+  valueFormatters?: { [P in Extract<keyof T, string>]?: (value: T[P] | void) => React.ReactNode }
```
fabien0102: Why this `void`? Logically we don't want and don't have `void` here, it's all the point of this `valueFormatter`.

BTW, I have the same error as before locally:

```tsx
const testdata = {
  a: "toto",
  b: "titi",
  c: "riri",
}

const myCard = <Card
  data={testdata}
  keyFormatter={val => val}
  valueFormatters={{ a: (val: string) => val }} // error here
/>
```

That can be resolve if I remove this `| void` and also add a `!` on line 48 to said to typescript that this value can't be undefined in this case.

```diff
-    return valueFormatter ? valueFormatter(value) : value
+    return valueFormatter ? valueFormatter(value!) : value
```

After I think that the logic can be improved near this line 48 to avoid this impossible `undefined` case, but it's better than before 👍

src/Card/Card.tsx
```diff
@@ -22,6 +22,8 @@ export interface CardProps<T = {}> extends DefaultProps {
   children?: React.ReactNode
 }
 
+export type CardComponent = <T extends {}>(props: CardProps<T>) => React.ReactElement<any> | null
```
fabien0102: A function that take `props` and return a `React.ReactElement` 🤔  I think that you have just reimplemented `React.SFC` 😉

src/Card/Card.tsx
```diff
@@ -3,13 +3,13 @@ import { CardHeader, CardItem } from "../"
 import { DefaultProps } from "../types"
 import styled from "../utils/styled"
 
-export interface CardProps<T = {}> extends DefaultProps {
+export interface CardProps<T extends {} = {}> extends DefaultProps {
```
fabien0102: To try (I'm not sure):
```tsx
extends { [key: string]: unknown }
```

So you need to `Extract<keyof T, string>` everywhere after, we get rid of `number` and `Symbol` as key, but in this case, I think that is make sense.

fabien0102: The good think with the `Extract <keyof T, string>` is that typescript infer that `keyof T` is actually a `string`, so your solution is fine :wink:

stereobooster: I tried `extends { [key: string]: unknown }`, but it seems TypeScript doesn't care.

src/Card/Card.tsx
stereobooster: well except that `React.SFC` have fixed parametric type (`React.SFC<Props>`), but this function is able to deduce type from environment e.g. `data` param.  `<? extends {}>(props: CardProps<?>)` -> `props = {data: {a: 123}}` ->  `CardProps<{a: number}>` -> `<{a: number}>(props: CardProps<{a: number}>)` etc

fabien0102: This is the original reason of my move to `React.PureComponent`, so you can do this:

```tsx
class Card<T> extends React.PureComponent<Props<T>>{}
```

So you can inject your generic into your props without create an extra type.

stereobooster: I tried it. It didn't work. It breaks on `wrapDefaultTheme<T>(Comp: React.ComponentType<T>): React.SFC<T>`. This was my initial attempt. It matches `<T>` from `Card<T>` and thinks you provided `React.ComponentType` without parameter. I can dig dipper.

stereobooster:
[I was working on types for redux-loop](https://github.com/redux-loop/redux-loop/pulls), now I'm back in the business. 

To make sure we are on the same page [`Object.keys` always returns strings](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/keys)

```js
Object.keys({ [Symbol(1)]: 1, [1]: 1, a: 1 })
//(2) ["1", "a"]
```

I'm using `Extract<keyof T, string>` to convince TS keys are strings, where is it should assume it by default. It is not just strings but also other types, because they [generalised index types](https://github.com/Microsoft/TypeScript/wiki/Breaking-Changes#keyof-now-includes-string-number-and-symbol-keys). I also tried `T extends {[key:string]: any}`, but this has no effect. Probably possible to do but  `Extract<keyof T, string>` works for now.

---

About parametric SFC component e.g. `class Card<T> extends React.PureComponent<Props<T>>{}`. I remember I tried something similar before and it breaks with some subtle errors. I believe that main issue here is that class expects parametric type to be provided before class will be initiated. It is more "static" in sense of types than functions. This is not precise, I can be wrong. I need to read how they model classes in TypeScript to explain the issue.

stereobooster:
## String keys

I can not replace 

```
export interface CardProps<T extends {} = {}> extends DefaultProps {
  keys?: Array<Extract<keyof T, string>> // string
```

with

```ts
CardProps<T extends { [key: string]: any } = {}> extends DefaultProps {
  keys?: Array<keyof T> // keyof T is string | number | symbol
}
```

but I can do

```ts
CardProps<T extends { [key: string]: any } = {}> extends DefaultProps {
  keys?: Array<Extract<keyof T, string>> // string
}
```

---

## Generic Classes vs Generic Function

We can create generic class like this (1):

```ts
interface CardComponent<T extends {}> extends React.PureComponent<CardProps<T>> {};
```

and we can create generic function (2):

```ts
type CardComponent<T extends {}> = (a: CardProps<T>) => React.ReactElement<any> | null; 
```

or this way (3):

```ts
type CardComponent = <T extends {}>(props: CardProps<T>) => React.ReactElement<any> | null
```

All types (and interface) are generic here. 1 and 2 sre similar, but 3 is different, it has free (unbound) parametric type. Free parametric type says type system - "I don't know what is this, use constraint solver and figure out what the type should be depending on the context where function used".
This is the difference, there is no way (at least to my knowledge) to create class with free parametric type.

With free parametric type it is possible to write:

```ts
<Card data={{ a: 1 }} valueFormatters={{ a: /* type is checked here */}}></Card>
```

but without we will need to write something like:

```ts
type Data = { a: number }
<Card<Data> data={{ a: 1 }} valueFormatters={{ a: /* type is checked here */}}></Card>
```

Generics without free parametric type requier parameter as soon as the used:

```
src/Card/__tests__/Card.test.tsx:9:49 - error TS2314: Generic type 'CardComponent<T>' requires 1 type argument(s).

9 const Card = wrapDefaultTheme(ThemelessCard) as CardComponent
                                                  ~~~~~~~~~~~~~
```

---

## Final thoughts

Current implementation works as expected to my opinion. Are there more changes required? Do you have any worries about using function type signature instead of TypeScript's provided `SFC` or `StatelessComponent` or `PureComponent`? I mean from React's point of view usage of function is legitimate and I don't see any potential violations of types. The problem would be if we need to use `defaultProps` property, but we don't need.

### defaultProps issue

The root issue here is that free parametric type intentionally not provided in implementation, but `defaultProps` require type at the moment we assign it

```ts
export type CardComponent = <T extends {}>(props: CardProps<T>) => React.ReactElement<any> | null
const Card: CardComponent = props => ...
const defaultProps: CardProps<{[key: string]: any}> = ...
// @ts-ignore
CardComponent.defaultProps = defaultProps
```

we can not do

```ts
const defaultProps: CardProps<T> = ...
```

We already using it, we already need to provide it.

Final solution:

```ts
export type CardComponent = (<T extends {}>(props: CardProps<T>) => React.ReactElement<any> | null) & {
  defaultProps?: Partial<CardProps<{ [key: string]: any }>>
}
```

As you can see this is pretty similar to `React.SFC`, but not the same because of free parametric type.

src/Card/Card.tsx
```diff
@@ -34,33 +36,25 @@ const Container = styled("div")(({ theme }) => ({
...
-    const titles = keyFormatter ? _keys.map(keyFormatter) : _keys
-    const values = _keys.map(
-      /**
-       * @todo Improve typings: the typing of this does not satisfy the typescript compiler
-       * https://github.com/contiamo/operational-ui/issues/629
-       */
-      // @ts-ignore
-      (i: Extract<keyof T, string>) => (valueFormatters[i] ? valueFormatters[i](data[i] as any) : data[i]),
-    )
+const Card: CardComponent = props => {
```
fabien0102: With a bit of luck, if you export this `Card` it can fix styleguidist. This little boy is a bit sensible 😕 

image

(I know, it can be a real fight to have TS and styleguidist happy together sometime, but it's a price to have a nice doc and DX)

fabien0102: https://deploy-preview-655--operational-ui.netlify.com/#CardColumns or `yarn start` locally ;)

stereobooster: Yes known pain point. I have this issue too, when styleguidist saliently goes broken. I looked for options to do typechecking inside `.md` files. Other options would be typechecking `.mdx` files and use something like `docz`. I will fix it

fabien0102: I've tried docz, but since it don't provide any playground and livereload, I don't really like it… (but maybe I missed something). I really like the approch of styleguidist, parse a simple markdown file, so you also have docs in github and it's really simple to maintain :) 

I would love to do some typecheking inside `.md` 😍 ! For now it's the biggest limitation of styleguidist (many open issues about this https://github.com/styleguidist/react-styleguidist/issues?utf8=%E2%9C%93&q=is%3Aissue+is%3Aopen+typescript) but I don't find the time to go deeper about this problematic.

(I also want to improve the styleguidist performance, I will probably try something with gatsby as doc generator when I will have some free time 😄 )

stereobooster: The issue is inside `react-docgen-typescript`. I believe it is possible to fix it inside `react-docgen-typescript`, I found workaround.

It doesn't like

```ts
const Card:X = () => ...
export default Card;
```

but it is ok with

```ts
export default function Card() {...}
```

One way or another I fixed it.

fabien0102:
@stereobooster I will do a final check this morning, I was a little busy this friday afternoon :wink:

stereobooster:
@fabien0102 I discovered that TS 3+ have some kind of type algebra (e.g. you can write function which accept type and return type), for example:

```ts
type UnwrapPromise<T> = T extends Promise<infer U> ? U : T;
type UnwrappedReturn<F> = F extends (...args: any[]) => infer R ? UnwrapPromise<R> : never; 
type FunctionArgs<F> = F extends (...args: infer T) => any ? T : never;
```

I guess this task can have another solution based on that. This is just FYI

fabien0102:
@stereobooster Yes it's true, I've already used this notation to have the return type of a function, but I can't remember in which project ^^ So many feature in TS, it's quite hard to remember everything 😃 

If you like this kind of advanced TS patterns, you should have a look to this repo https://github.com/tycho01/typical, I often find great solutions from there💯

src/Card/__tests__/Card.test.tsx
```diff
@@ -1,9 +1,12 @@
...
 import * as React from "react"
-import { Card as ThemelessCard } from "../../index"
+import { Card as ThemelessCard, CardComponent } from "../../index"
 import wrapDefaultTheme from "../../utils/wrap-default-theme"
 
-const Card = wrapDefaultTheme(ThemelessCard)
+// we need to reinforce type
+// because after wrapDefaultTheme it is general React.SFC
+// but we want parametric function
+const Card = wrapDefaultTheme(ThemelessCard) as CardComponent
```
fabien0102: If `as typeof CardComponent` works here, you can remove the export of `CardComponent`

stereobooster: done

TejasQ:
@stereobooster if you rebase this, we can merge! 🚀 🎉

stereobooster:
@TejasQ did it

PR #658, opened 2018-08-03
Workaround compiled `.d.ts` files

peterszerzo:
### Summary

This PR fixes `.d.ts` files so dependent projects build with typings successfully. The errors fixed are the following:

#### 1. Paths

Fixes an error that incorrectly references internal `.d.ts` files when compiled. A project trying to compile depending on the package might get something like this:

```
node_modules/@operational/components/lib/src/utils/constants.d.ts:184:17 - error TS2307: Cannot find module 'src/utils/constants/deprecatedTheme'.

184     deprecated: import("src/utils/constants/deprecatedTheme").Theme;
```

#### 2. `import("eac")`

When `styled` is used in a file but `React` isn't, the compiler gets confused and truncates `import("react")` to `import("eac")`. This is fixed by adding react to scope. 

### To be tested

Me
- [x] Components look and work as before
- [x] When compiled, a full search of `lib/**/*.d.ts` for `import("eac")` only yields occurrences that are commented out

Tester 1

- [ ] Components look and work as before
- [ ] When compiled, a full search of `lib/**/*.d.ts` for `import("eac")` only yields occurrences that are commented out
  

Tester 2

- [ ] Components look and work as before
- [ ] When compiled, a full search of `lib/**/*.d.ts` for `import("eac")` only yields occurrences that are commented out


TejasQ:
The PR that undoes this is going to make me so happy.

PR #661, opened 2018-08-06
Feat theme

fabien0102:
### Summary

This is required for the Use Case Library app since we need the power of Operational UI with some extensions, namely to the primary color.

image

### To be tested

Me

Tester 1

Tester 2

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,7 +66,6 @@
     ]
   },
   "dependencies": {
-    "deepmerge": "^2.1.1",
     "emotion": "^9.2.5",
     "emotion-theming": "^9.2.6",
     "lodash": "^4.17.10",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -66,6 +66,7 @@
     ]
   },
   "dependencies": {
+    "deepmerge": "^2.1.1",
```
TejasQ: Does lodash allow this functionality? Why are we using another library for this?

fabien0102: I find merge, but not deep merge. but maybe I missed something 🤔

TejasQ: AFAIK lodash merge _is_ deep. Why don’t we just `Object.assign` actually?

PR #665, opened 2018-08-09
Splash component improvements

### LINKED ISSUE #650: `Splash` component improvements
This issue describes changes planned for the [Splash](https://github.com/Contiamo/operational-ui/blob/master/src/Splash/Splash.tsx) component in order to make it more customizable.

### Customizable logos

Currently, the `Splash` component's logo is hard-coded `Operational` one, with the spinning animation relying on a state flag set every 8 seconds. This should be customizable for outside projects and other use-cases. The following steps are required to make this happen:
* the `rotation` handling and state management should move from `Splash.tsx` to [`Splash.Logo.tsx`](https://github.com/Contiamo/operational-ui/blob/master/src/Splash/Splash.Logo.tsx). This will allow the spinning logo to be self-contained and swappable to a different one with its own animations.
* `Splash.tsx` should take a `logo?: React.ReactNode` prop to allow for customizable logos. It should default to the current hard-coded one in case one isn't provided.

### Readable colors

Text colors in the `Splash` component are currently hard-coded to white, which won't be readable if the background is set to a very bright color. As in the [`Button` implementation](https://github.com/Contiamo/operational-ui/blob/master/src/Button/Button.tsx#L47) in , this color should toggle between white and dark using the `readableTextColor` utility. This would involve tweaking animation colors, logo colors and backdrop colors.

### Dev workflow

Simply clone the repo and run `yarn install` + `yarn start` to get the dev environment going. See [React Styleguidist](https://github.com/styleguidist/react-styleguidist) for help developing with styleguidist if you haven't had the chance to work with it yet.

Taym95:
Try to fix #650


TejasQ:
@Taym95 that's because your `textColor_` prop is optional on line 1 on your snippet. Removing the `?` should fix it.

Taym95:
Yeah, I tried to remove `?`, still the same issue, I removed from SplashProps `textColor: string`, I got: `Property 'textColor' is missing in type '{ children: Element[]; color: string; title: string; actions: Element; }'.`

TejasQ:
Right. You'll find the answer in the `readableTextColor`. I recommend having a look. Basically, `readableTextColor` expects only a string, so passing it an undefined value is a complete no-go. Excited to see your solution!

Taym95:
Got it.

TejasQ:
A _really good practice_ is to test things locally, and then test them on the automated deployment system: go take a look at [this](https://deploy-preview-665--operational-ui.netlify.com/).

src/Splash/Splash.tsx
```diff
@@ -12,12 +13,15 @@ export interface SplashProps {
   actions: React.ReactNode
   /** Main content */
   children: React.ReactNode
+  /** Customizable logo */
+  logo?: React.ReactNode
   /** Backdrop color, dark enough to support white text as in the example. */
   color?: string
+  /** Text color in the Splash component */
+  textColor?: string
```
peterszerzo: I don't think we need this prop - the `color` prop already determines whether the text color should be white or dark, and I don't see a design need to have text colors be more flexible than this for the time being.

src/Splash/Splash.tsx
```diff
@@ -38,9 +42,11 @@ const Container = styled("div")<{ color_?: string }>`
   `};
 `
 
-const Content = styled("div")`
+const Content = styled("div")<{ textColor_?: string }>`
   position: relative;
-  color: #ffffff;
+  ${({ textColor_, theme }) => `
+      color: ${textColor_ ? readableTextColor(textColor_, [theme.color.white, theme.color.black]) : "#ffffff"};
```
peterszerzo: I think we should pass in the `color` prop here, as the first argument to `readableTextColor` is the background color. Maybe we should even make the API of the helper something like `readableTextColor({ backgroundColor: "#1499CE", colorChoices: [ "#FFF", "#000" ] })` to make it clearer (that's just a note for later).

TejasQ:
@Taym95 we'll close this tomorrow since it's been two weeks with radio silence. Please respond if you'd like to continue on this by tomorrow.

PR #674, opened 2018-08-15
**Fix:** Clipped tooltip

peterszerzo:
## Summary

Attempting to fix the left-positioned copy-to-clipboard tooltip that is currently clipped by the sidenav. As a last resort, we can bottom-position like in the image below, but we should investigate a proper solutions first.

screen shot 2018-08-15 at 11 24 43 am

Things I'm trying:
* higher z-index hierarchies for `Main` than for `Side` in `Layout.tsx`.
* a number of components (`Main` in `Layout.tsx`, `ViewContainer` in `Page.tsx` have `overflow: hidden` to prevent double scrollbars, which unfortunately also clips sticking out horizontally. Changing those to `overflow-y: hidden; overflow-x: auto;` should help.

Some reading: https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context.

## To be tested

Me
- [ ] copy to clipboard note isn't clipped by sidenav

Tester 1

- [ ] copy to clipboard note isn't clipped by sidenav

Tester 2

- [ ] copy to clipboard note isn't clipped by sidenav

### DIFF
--- a/src/Input/Input.tsx
+++ b/src/Input/Input.tsx
@@ -151,7 +151,7 @@ class Input extends React.Component<InputProps, State> {
     return this.props.copy ? (
       <CopyToClipboard text={this.props.value || ""} onCopy={this.showTooltip}>
         <InputButton>
-          {this.state.showTooltip && <Tooltip left>Copied!</Tooltip>}
+          {this.state.showTooltip && <Tooltip bottom>Copied to clipboard</Tooltip>}
           <Icon name="Copy" size={16} />
         </InputButton>
       </CopyToClipboard>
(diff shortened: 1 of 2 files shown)


peterszerzo:
[Clipped tooltip between page content and sidenav](https://trello.com/c/Uo3tIFnX/374-clipped-tooltip-between-page-content-and-sidenav)

TejasQ:
I did some digging and found:

- `Layout` has `overflow: hidden`
  - `Page` has `overflow: hidden`
    - `PageContent` has `overflow: auto`

Each of these components and their styles clip the tooltip. When the overflow rules are relaxed, the tooltip is free.

TejasQ:
Closed because it's been inactive over a month. We'll come back to it if/when we work on it again.

peterszerzo:
Sorry I missed the question - yes, fixing this probably requires a fair bit of engineering, or cleanest of all, portals. Let's get back to this later.

Issue #660, opened 2018-08-06
Page height is broken. Again.


TejasQ:
On a `Page` component, the height is incorrect: the bottom of any page is clipped. This can be seen in all (❗️) of our internal UIs.

![image](https://user-images.githubusercontent.com/9947422/43705496-eb4eaf08-9962-11e8-8efc-2bf493174158.png)

peterszerzo:
I would say this behavior is preferable - if `Page` just gives you a content area that is precisely the available height, then child components can either make themselves scrollable or implement a non-scrolling single-page card layout where the cards scroll and not the page itself. This is best decided down the tree in my opinion.

TejasQ:
I think we're mixing concerns a bit here. I'm not talking about scrollability, I'm talking about padding.

From my point of view, one would expect a `Page` to correctly contain its children, including displaying them with correct padding.

[Take a look at the table list here](https://labs.fe.ctmo.io/21b899ae-3eba-56a7-97da-fa26c050d1c5/4c495436-4973-5138-bb06-7a304df61932/bundles/) to see how, irrespective of scroll nature or height of children, the `Page` just behaves unexpectedly: I would expect it to pad its children correctly and leave a bit of space at the bottom, instead of clipping its children.

peterszerzo:
Got it, that makes sense.

PR #675, opened 2018-08-15
**Breaking: ** Add `CardSection` component

peterszerzo:
## Summary

This PR adds `CardSection`, a component that subdivides cards in various stacking configurations, display states and actions.

To support this component, there are some additional changes to `Card`:
* removing the fight between padding and negative margins in `CardHeader`.
* the `Card` component's `action` prop is now a simple react node as opposed to a component, consistent with other react child props.
* bottom margin is cancelled on the last card in its container. This has caused unwanted scrollbars in layouts, and will make layout fixing in #660 more predictable.

screen shot 2018-08-15 at 4 09 04 pm

drag-and-drop

## To be tested

Me
- [ ] CardSection elements look good.
- [ ] CardSection elements support actions.
- [ ] CardSection elements support visual drag and drop feedback.
- [ ] Usages of `Card` do not break.

Tester 1

- [ ] CardSection elements look good.
- [ ] CardSection elements support actions.
- [ ] CardSection elements support visual drag and drop feedback.
- [ ] Usages of `Card` do not break.

Tester 2

- [ ] CardSection elements look good.
- [ ] CardSection elements support actions.
- [ ] CardSection elements support visual drag and drop feedback.
- [ ] Usages of `Card` do not break.


src/CardSection/CardSection.tsx
```diff
@@ -0,0 +1,83 @@
...
+  position: relative;
+  ${({ theme }) => `
+    border-right: 1px solid ${theme.color.separators.default};
+  `};
+`
+
+const Overlay = styled("div")`
+  position: absolute;
+  top: 0px;
+  bottom: 0px;
```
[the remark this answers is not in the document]
peterszerzo: It means stretch to cover the entire surface of the parent.

src/ActionMenu/ActionMenu.tsx
```diff
@@ -5,7 +5,7 @@ import { DefaultProps } from "../types"
 import { darken } from "../utils"
 import styled from "../utils/styled"
 
-const width = 144
+const width = 120
```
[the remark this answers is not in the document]
TejasQ: But why did that even come up if this is fixing `Card`? 🤔

src/Card/Card.tsx
```diff
@@ -15,29 +15,58 @@ export interface CardProps<T extends {} = {}> extends DefaultProps {
...
+   * Since there is no further element to separate this card from, the bottom margin is disabled.
+   * This avoids unwanted overflowing behavior especially in cases where there is a single card.
+   */
+  "&:last-child": {
+    marginBottom: 0,
+  },
 }))
 
+const Content = styled("div")`
+  padding: 20px;
```
TejasQ: `theme.space.element`?

src/CardSection/CardSection.tsx
```diff
@@ -0,0 +1,83 @@
...
+  right: 0px;
+  background-color: rgba(0, 0, 0, 0.1);
+  ${({ theme }) => `
+    z-index: ${theme.zIndex.tooltip};
+  `};
+`
+
+const Content = styled("div")`
+  display: block;
+  padding: 20px;
```
TejasQ: `theme.space.element`?

src/ActionMenu/ActionMenu.tsx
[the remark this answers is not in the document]
peterszerzo: It was more apparent that it wasn't looking good.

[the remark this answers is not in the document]
micha-f: If it looks better this way then let's change it!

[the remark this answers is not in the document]
peterszerzo: @micha-f can you think of the longest piece of text inside an action when the dropdown is open? I should make sure there won't be an overflow before we change it across the board, or pick an in-between width.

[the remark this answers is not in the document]
TejasQ: _"Backpfeifengesicht"_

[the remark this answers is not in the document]
micha-f: Inside the actions dropdown?
-> "Create saved query"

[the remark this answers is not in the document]
peterszerzo: I see, in that case we'll go back to 144px then and see how it fares when used with realistic card sized.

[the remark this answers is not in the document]
micha-f: 👍

src/CardSection/CardSection.tsx
```diff
@@ -0,0 +1,137 @@
...
+  position: relative;
+  ${({ theme }) => `
+    border-right: 1px solid ${theme.color.separators.default};
+  `};
+`
+
+const Overlay = styled("div")<{ overlayType: OverlayType }>`
+  position: absolute;
+  top: 0px;
+  bottom: 0px;
```
TejasQ: I still don't understand the need for both top _and_ bottom to have `0`.

src/Tree/Tree.tsx
```diff
@@ -35,43 +35,62 @@ const TreeItem = styled("div")<{
...
 
+const TreeLabel = styled("span")`
+  display: block;
+  flex: 1;
+`
+
+const CloseButton = styled("div")`
+  cursor: pointer;
+  width: fit-content;
+  padding: 4px;
```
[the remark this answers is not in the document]
peterszerzo: I wouldn't use a theme variable here yet - it is a unique element (basically a close button that is even smaller than the `NameTag` component) whose proportions should not change as theme variables are fine-tuned to make sure the X icon inside stays visible. I'll add a `@todo` note to look into formalizing and re-using it in other places.

src/Tree/Tree.tsx
```diff
@@ -35,43 +35,62 @@ const TreeItem = styled("div")<{
...
   width: fit-content;
   min-width: 160px;
-  user-select: none;
-  ${({ theme, hasChildren, hasTag, isTopLevel, isDisabled }) => `
+  ${({ theme, hasChildren, hasTag, isTopLevel, isDisabled, isRemovable }) => `
     font-size: ${hasTag ? theme.font.size.fineprint : theme.font.size.small}px;
     font-weight: ${hasTag || isTopLevel ? theme.font.weight.bold : theme.font.weight.regular};
     font-family: ${hasTag ? theme.font.family.code : theme.font.family.main};
     color: ${theme.color.text.dark};
+    ${isDisabled ? "opacity: 0.4;" : ""}
```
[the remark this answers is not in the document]
peterszerzo: Good point!

src/Card/Card.tsx
```diff
@@ -48,9 +79,15 @@ export default function Card<T extends {}>(props: CardProps<T>) {
...
     <Container {...rest}>
-      {(title || Action) && <CardHeader title={title} action={Action && <Action />} />}
-      {data && titles.map((cardItemTitle, i) => <CardItem key={i} value={values[i]} title={cardItemTitle} />)}
-      {children}
+      {(title || action) && <CardHeader title={title} action={action} />}
+      {sections ? (
+        <SectionsContainer stackHorizontal={stackSections === "horizontal"}>{sections}</SectionsContainer>
+      ) : (
+        <Content>
+          {data && titles.map((cardItemTitle, i) => <CardItem key={i} value={values[i]} title={cardItemTitle} />)}
```
TejasQ: Can titles be reordered?

src/CardSection/CardSection.tsx
```diff
@@ -0,0 +1,137 @@
...
+          background-color: rgba(239, 223, 50, 0.4);
+        `
+    }
+  }};
+`
+
+const Content = styled("div")`
+  display: block;
+  ${({ theme }) => `
+    padding: ${theme.space.element}px;
```
TejasQ: I'm curious as to why you prefer the function interpolation occurring _before_ the value as opposed to _after_:

```js
styled("div")`
  ${({ theme }) => `
    padding: ${theme.space.element}px;
  `}
`
```

as opposed to 

```js
styled("div")`
  padding: ${({ theme }) => theme.space.element}px;
`
```

src/CardSection/README.md
```diff
@@ -0,0 +1,162 @@
+Card sections are flexibly stacked sub-containers designed to fit within cards. If specified, cards ignore other content-setting properties (`data`, `children`).
+
+### Simple usage
+
+```jsx
+<div style={{ width: 300 }}>
+  <Card
+    sections={
+      <>
+        <CardSection title="Section 1">Content 1</CardSection>
```
TejasQ: Can I give it an array and have it map automatically to CardSections as syntax sugar? I'd find that shorthand nice and convenient.

src/CardSection/README.md
```diff
@@ -0,0 +1,162 @@
...
+    super(props)
+    this.state = {
+      labelsInSource: ["123", "456", "789"],
+      labelsInTarget: [],
+    }
+  }
+
+  render() {
+    return (
+      <OperationalUI>
```
TejasQ: Pretty sure you already have this from styleguidist. It may be a good idea to use the `state` global that styleguidist allows in order to keep this concise.

src/CardSection/README.md
```diff
@@ -0,0 +1,162 @@
...
+        <div style={{ width: 600, height: 300, margin: 20 }}>
+          <Card>
+            <Tree
+              trees={this.state.labelsInSource.map(label => ({
+                label,
+                draggable: true,
+                onDragStart: ev => {
+                  this.setState({ dragSource: label })
+                },
+                onDragEnd: ev => {
```
TejasQ: Isn't JS awesome because you don't have to deal with event handler typings?

src/CardSection/README.md
```diff
@@ -104,52 +112,51 @@ class DragAndDropExample extends React.Component {
             stackSections="horizontal"
             sections={
               <>
+                {/* Drop target identified by "container1" */}
```
TejasQ: How `container1` is declared and flows through this component _still seems_ a little ambiguous/not as idiomatic to a newcomer (like me). Can we somehow abstract a bit more? I'm unclear to me what the `dragAndDropFeedback` even does at this point.

peterszerzo: `dragAndDropFeedback` is the prop that can take a few string values making either a yellow background to signal that you can drag there, a slightly more intense one to signal that you're about to drop, and a red border to signal that the current drag item cannot be dropped in there. It is purely presentational. Any ideas for a better name?

src/CardSection/CardSection.tsx
peterszerzo: This allows me to have a single function in the implementation, and I find the rule itself - `padding: ${theme.space.element}px;`, easier to read than `padding: ${({ theme }) => theme.space.element}px;`.

src/CardSection/README.md
TejasQ: Nope, it works.

src/Card/Card.tsx
peterszerzo: I'm not that familiar with this feature, but reading through it it it doesn't look like reordering is possible. It also strikes me as a feature where there's no guarantee that the titles will be ordered the same way in different browsers, given that `Object.keys` returns an unpredictable order.

I think we should sort the keys alphabetically in https://github.com/Contiamo/operational-ui/blob/0f0cb668f90691744961f369017722b66d625617/src/Card/Card.tsx#L72, then accept a `sortKeys: (title1: string, title2: string) => number` prop that we can sort with if a custom ordering is required.

src/CardSection/CardSection.tsx
TejasQ: Right. Looks like it also scales better as in the case of line 88.

src/CardSection/README.md
```diff
@@ -0,0 +1,175 @@
...
+          </Card>
+          <Card
+            stackSections="horizontal"
+            sections={
+              <>
+                {/* Drop target identified by "container1" */}
+                <CardSection
+                  title="Container 1"
+                  dragAndDropFeedback={
+                    this.state.dragSource
```
TejasQ: Is `this.state.dragSource` only truthy when some dragging is in progress?

fabien0102:
```<!-- IMPORTANT: If this is a breaking change or a backwards compatible feature, please prefix the title of this PR with **Breaking: ** or **Feature: ** -->```


@peterszerzo This is the first line of the PR template 😛 Can you follow the convention please 😅

File: src/CardSection/README.md (hunk 14 above)
Comment (peterszerzo) [reply src/CardSection/README.md]: That's right - in this example, it simply contains the string value inside the draggable item when dragging is going on, otherwise it's `undefined`.

File: src/Card/Card.tsx [hunk 15]
```diff
@@ -15,29 +15,60 @@ export interface CardProps<T extends {} = {}> extends DefaultProps {
   /** Title of the card */
   title?: React.ReactNode
   /** Component containing buttons/links/actions assigned to the card */
-  action?: React.ComponentType
-  /** DOM id attribute, useful for hash linking */
-  id?: string
-  className?: string
+  action?: React.ReactNode
+  /** Card sections */
+  sections?: React.ReactNode
```
fabien0102: We can improve the types here to have 

```ts
interface CardWithSections {
sections?: React.ReactNode
children?: never
data?: never
}
```

peterszerzo: @fabien0102 thanks for the tip! Have a look in the new commit :)

src/Card/Card.tsx
```diff
@@ -3,7 +3,7 @@ import { CardHeader, CardItem } from "../"
 import { DefaultProps } from "../types"
 import styled from "../utils/styled"
 
-export interface CardProps<T extends {} = {}> extends DefaultProps {
+export interface CardPropsRegular<T extends {} = {}> extends DefaultProps {
```
TejasQ: Why `Regular`? We use `CardPropsBase` or `BaseCardProps` conventionally in other components with discriminant typings.

src/CardSection/README.md
```diff
@@ -0,0 +1,180 @@
...
+
+### A drag-and-drop example
+
+The following example shows how card sections can support a full-fledged drag-and-drop example. This example is quite verbose to implement the HTML5 drag-and-drop API, but beyond the logic, the different settings to toggle drag-and-drop feedback states are all represented.
+
+```jsx
+class DragAndDropExample extends React.Component {
+  constructor(props) {
+    super(props)
+    this.state = {
```
TejasQ: I'm wondering if we can use styleguidist's built in support for state here in order to make the example simpler (and possibly more readable?)

src/CardSection/README.md
```diff
@@ -0,0 +1,180 @@
...
+      /** The item being dragged out of the above two arrays, if any */
+      dragSource: undefined,
+      /** The identifier of the current drop target, if any */
+      dropTarget: undefined,
+    }
+  }
+
+  render() {
+    return (
+      <OperationalUI>
```
TejasQ: Using styleguidist's built in state support would even remove the need for this `OperationalUI` wrapper.

src/CardSection/README.md
```diff
@@ -0,0 +1,180 @@
...
+                      },
+                    }))}
+                  />
+                </CardSection>
+                {/* Drop target identified by "container2" */}
+                <CardSection
+                  title="Container 2"
+                  onDragOver={() => {
+                    ev.preventDefault()
+                    // This check is necessary to prevent blocking the render loop with queued up re-renders.
```
TejasQ: `// This check is necessary to prevent unnecessarily calling setState if this component is already the dropTarget`

src/Card/Card.tsx
peterszerzo: In https://github.com/Contiamo/operational-ui/blob/48ad60e27612e9fb4dbfe3a26aeb6f005bbbf515/src/Input/Input.tsx, for example, `BaseProps` means an interface that the two discriminant interfaces inherit from, which is not the case here. Renaming it to e.g. `CardWithDataOrChildrenProps` and `CardWithSectionsProps` to be more descriptive - what do you think?

src/CardSection/README.md
peterszerzo: This wasn't working for me in `styleguidist`.

src/CardSection/README.md
peterszerzo: I'd still mention the render loop blocking as the outcome I'd like to prevent - how about this:
```
/**
 * This check is necessary to prevent blocking the render loop by piling up unnecessary `setState` calls
 * when this element is already a drop target.
 */
```

TejasQ:
Adding one more comment so there are 50 comments.

TejasQ:
[Card header tabs](https://trello.com/c/1By5iUWw/375-card-header-tabs)

Issue #676, opened 2018-08-16
Page actions position


fabien0102:
![image](https://user-images.githubusercontent.com/1761469/44212881-d496e980-a16c-11e8-8ddc-0e663c05ae82.png)

### Proposed API

```jsx
<Page title="" actions={<Button />} actionsPosition="start|end" />
```

We also need to an `iconPosition` on `Button`, for now I can only have icon on right.

PR #679, opened 2018-08-20
Allow Page to show actions on the start or end of the title bar

mjopson:
## Summary

Hi! Thought I'd jump right in with this PR...

Allows setting the position of actions as described in https://github.com/contiamo/operational-ui/issues/676.  This doesn't address the idea of having buttons with configurable icon positions; figured that was a separate change.

You do get a bit of extra spacing between elements when putting actions on the left; not sure if this is desired or not.

image

## Related issue

https://github.com/contiamo/operational-ui/issues/676

## To be tested

Me

Tester 1

Tester 2

### DIFF
--- a/src/Page/Page.tsx
+++ b/src/Page/Page.tsx
@@ -66,13 +66,19 @@ const Container = styled("div")(({ theme }) => ({
   backgroundColor: theme.color.background.lighter,
 }))
 
-const TitleBar = styled("div")(({ theme }) => ({
+const TitleBar = styled("div")<{ actionsPosition: "start" | "end" }>(({ theme, actionsPosition }) => ({
   backgroundColor: theme.color.primary,
   display: "flex",
   alignItems: "center",
   padding: theme.space.element,
   height: theme.titleHeight,
   fontWeight: theme.font.weight.medium,
+  "& > :nth-child(n)": {
+    marginRight: theme.space.element,
+  },
+  "& > :nth-child(2)": {
+    order: actionsPosition === "start" ? -1 : undefined,
+  },
 }))
 
 const tabsBarHeight = 43
@@ -105,9 +111,7 @@ const ViewContainer = styled("div")<{ isInTab?: boolean }>(({ theme, isInTab }) 
   position: "relative",
 }))
 
-const EndContainer = styled("div")(({ theme }) => ({
-  marginLeft: theme.space.element,
-}))
+const ActionsContainer = styled("div")(() => ({}))
 
 const initialState = {
   activeTab: 0,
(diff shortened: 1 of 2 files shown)


TejasQ:
Thanks so much for your contribution, @mjopson! Great stuff! Let's quickly check with our core designer about the extra spacing on the left. @micha-f, how's it looking?

src/Page/Page.tsx
```diff
@@ -104,7 +105,7 @@ const ViewContainer = styled("div")<{ isInTab?: boolean }>(({ theme, isInTab })
   position: "relative",
 }))
 
-const ActionsContainer = styled("div")(({ theme }) => ({
+const EndContainer = styled("div")(({ theme }) => ({
```
TejasQ: Out of curiosity, why are we renaming this?

mjopson: It used to always contain actions, but now it contains whatever is on "the end", which might not be actions anymore.

fabien0102: For me, this container should be semantically `ActionContainer`, after we "just" need to move this `ActionContainer` to `start` or `end` regarding the options.

I think that we can clean up a bit this implementation :wink:

micha-f:
I think it is nice to have a bit more spacing when actions are on the left as that communicates that the page title does not belong to the actions. Looks good.

mjopson:
An issue is that ActionsContainer has a left margin on it due to it being the second item.  When the order is reversed, the margin creates a problem.

Perhaps this instead: we have TitleBar reorder and add margins to its children.  We use right margin rather than left for simplicity, because we can safely add it to all of the elements rather than having to exclude the first (this would continue to work if more elements were added to the title bar).

In TitleBar:

  "& > :nth-child(n)": {
    marginRight: theme.space.element,
  },
  "& > :nth-child(2)": {
    order: actionsFirst ? -1 : undefined,
  }

"nth-child(n)" is a specificity hack - Title has defines its own margins that we are overriding.  The nth-child(2) here feels fragile; what I really want to say is "the ActionsContainer element".  Supposedly we can do this with syntax like the following:

[ActionsContainer]: {
  ....
}

But it seems this is only possible with babel-plugin-emotion.

TejasQ:
I understand the concern, @mjopson. We actually need this feature rather quickly and so @fabien0102 from our team [was able to solve it already](https://github.com/contiamo/operational-ui/pull/683/files#diff-88ff81fac534225deda44f519755137c). I have to say – he has done a really good job as it does not really involve a lot of convolutions: it simply reverses flex direction and deals with the margin based on the passed in prop. Ideally, we should not have to override _our own_ CSS with specificity as we develop our own components. I'm not saying that we don't do this: I'm sure such hacks exist in this codebase – but rather, I'm suggesting we keep it to a minimum.

For this reason, I'm going to have to close this PR in favor of #683. I hope this is okay.

You're more than welcome to find another issue and submit a pull request. In fact, I'd encourage it. 😄  It may be more efficient to ask to take over an issue beforehand so we can evaluate its urgency and factor in the questions of _"do we need this sooner? Is this something the internal team should handle because it requires more rapid offline conversation and iteration?" etc.

Thanks again!

PR #680, opened 2018-08-21
**Feature: ** Add condensedTitle option to Page component.

ImogenF:
## Summary

<blockquote class="trello-card"><a href="https://trello.com/c/tfzDlrV0/377-condensed-page-title-with-tabs">Condensed page title with tabs</a></blockquote>

image

- Tabs moved to inside the title bar if `condensedTitle` is true

## To be tested

Me

Tester 1

- [x] Layout looks good
- [x] No existing examples are broken
  

Tester 2

- [x] Layout looks good
- [x] No existing examples are broken

### DIFF
--- a/src/Page/Page.tsx
+++ b/src/Page/Page.tsx
@@ -26,7 +26,7 @@ export interface PropsWithSimplePage extends BaseProps {
   tabs?: never
   activeTabName?: never
   onTabChange?: never
-  condensedTabs?: never
+  condensedTitle?: never
 }
 
 export interface PropsWithComplexPage extends BaseProps {
@@ -37,7 +37,7 @@ export interface PropsWithComplexPage extends BaseProps {
   tabs?: never
   activeTabName?: never
   onTabChange?: never
-  condensedTabs?: never
+  condensedTitle?: never
 }
 
 export interface PropsWithTabs extends BaseProps {
@@ -60,7 +60,7 @@ export interface PropsWithTabs extends BaseProps {
   areas?: never
   fill?: never
   /** Condensed option */
-  condensedTabs?: boolean
+  condensedTitle?: boolean
 }
 
 export type PageProps = PropsWithSimplePage | PropsWithComplexPage | PropsWithTabs
@@ -172,12 +172,12 @@ class Page extends React.Component<PageProps, Readonly<typeof initialState>> {
   private renderTabsBar() {
     const tabs = this.props.tabs!
     const activeTab = this.getActiveTab(tabs)
-    const { condensedTabs } = this.props
+    const { condensedTitle } = this.props
 
     return (
-      <TabsBar condensed={condensedTabs}>
+      <TabsBar condensed={condensedTitle}>
         {tabs.filter(({ hidden }) => !hidden).map(({ name }, i) => (
-          <Tab condensed={condensedTabs} key={i} active={i === activeTab} onClick={() => this.onTabClick(i, tabs)}>
+          <Tab condensed={condensedTitle} key={i} active={i === activeTab} onClick={() => this.onTabClick(i, tabs)}>
             {name}
           </Tab>
         ))}
@@ -189,7 +189,7 @@ class Page extends React.Component<PageProps, Readonly<typeof initialState>> {
     const tabs = this.props.tabs!
     const activeTab = this.getActiveTab(tabs)
     const currentTabChildren = tabs[activeTab].children
-    const { title, actions, actionsPosition, condensedTabs } = this.props
+    const { title, actions, actionsPosition, condensedTitle } = this.props
 
     return (
       <>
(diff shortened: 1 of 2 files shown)


src/Page/Page.tsx
```diff
@@ -24,6 +24,7 @@ export interface PropsWithSimplePage extends BaseProps {
   tabs?: never
   activeTabName?: never
   onTabChange?: never
+  condensedTabs?: never
```
TejasQ: I think `condensed` is enough, following the convention of `Button` et al.

src/Page/Page.tsx
```diff
@@ -139,55 +151,71 @@ class Page extends React.Component<PageProps, Readonly<typeof initialState>> {
...
-          ))}
-        </TabsBar>
+        {title && (
+          <TitleBar>
+            <Title color="white">{title}</Title>
+            {condensedTabs && this.renderTabsBar()}
+            <ActionsContainer condensed={condensedTabs}>{actions}</ActionsContainer>
+          </TitleBar>
+        )}
+        {!condensedTabs && this.renderTabsBar()}
```
TejasQ: Is it possible to have a bar of tabs _without_ a title?

src/Page/Page.tsx
ImogenF: But in the case of the `Button` component, it is the button itself that is condensed. Here, it's not `Page` that is condensed, but rather just a specific part of the `Page`.

TejasQ: Good point. 👍

src/Page/Page.tsx
ImogenF: Ah. As it stands, it is possible to have a bar of tabs without a title bar, but not possible to have condensed tabs or actions if there is no title.

ImogenF: Is there a use case for actions without a title? Regardless of the tabs? If so, I can enable this, but before this PR it was not possible to have actions without a title, so I assumed that would still be the case.

ImogenF:
[Condensed page title with tabs](https://trello.com/c/tfzDlrV0/377-condensed-page-title-with-tabs)

src/Page/Page.tsx
TejasQ: I wasn't suggesting having tabs without a title: they're _in_ the title. Rather, the opposite: I think this line (185) ought to be inside the `title &&` expression because we ought not have tabs without a page title. Currently, if a page has no title, it has no tabs (as can be observed in this PR's preview).

TejasQ: Currently, no. @micha-f what do you foresee in future designs?

src/Page/Page.tsx
peterszerzo: It seems to me that both `condensedTitle` and `condensedTabs` will leave room for confusion. Wouldn't it be clearer from the outside to read that the entire component is either `condensed` or not, and what that actually means is up to the component itself?

src/Page/Page.tsx
```diff
@@ -139,55 +151,73 @@ class Page extends React.Component<PageProps, Readonly<typeof initialState>> {
...
-          ))}
-        </TabsBar>
+        {title && (
+          <>
+            <TitleBar>
+              <Title color="white">{title}</Title>
+              {condensedTabs && this.renderTabsBar()}
+              <ActionsContainer condensed={condensedTabs}>{actions}</ActionsContainer>
+            </TitleBar>
+            {!condensedTabs && this.renderTabsBar()}
```
TejasQ: [This](https://github.com/contiamo/operational-ui/pull/680/files#diff-88ff81fac534225deda44f519755137cR182) and this current line seem like a bit of a logical paradox, no?

src/Page/Page.tsx
ImogenF: I think using just `condensed` also leaves room for confusion - what if there are no tabs? The user would expect `condensed` to do something to the page but nothing would change. I suppose the question here is whether there are any other parts of the page component that may potentially be condensed as well in the future - if so, then `condensed` makes sense. If not, I think it is clearer to explicitly say what is being condensed. (The tabs are being condensed into the title, so I suppose either `condensedTabs` or `condensedTitle` would work)

src/Page/Page.tsx
ImogenF: How come? If `condensedTabs` is true, the tabs are rendered inside the `TitleBar`, otherwise they're rendered outside of it.

src/Page/Page.tsx
peterszerzo: Sounds like an input from design can clear this up - @micha-f can you read through this discussion and tell us your thoughts?

src/Page/Page.tsx
TejasQ: Ah, I get it. I thought the `TitleBar` _always_ contained the tabs bar, condensed or not. Didn't realize the tabs bar was a separate bar. Makes sense.

src/Page/Page.tsx
micha-f: My feeling is that `condensedTitle` is a bit more clear here. We might add other options to the page layout eventually (like smaller margins for the OLAP editor) and being precise about what this setting does should help to avoid confusion.

PR #682, opened 2018-08-21
Deprecate the explicit use of the CardHeader component

peterszerzo:
## Summary

This PR deprecates the direct use of the `CardHeader` component:

```
<Card>
  <CardHeader title="Title" action={<Button />}>
</Card>
```
in favor of simply writing
```
<Card title="Title" action={<Button />}>
```

This deprecation preps the breaking change in #675.

WIP: typings could use some improvements, which is proving to be challenging because of the rather complex interaction between `Card` and `deprecate` typings.

## To be tested

Me
- [x] `Card` and `CardHeader` examples look the same
- [x] Using `CardHeader` directly throws a deprecation error

Tester 1

- [ ] `Card` and `CardHeader` examples look the same
- [ ] Using `CardHeader` directly throws a deprecation error

Tester 2

- [ ] `Card` and `CardHeader` examples look the same
- [ ] Using `CardHeader` directly throws a deprecation error

### DIFF
--- a/src/Card/Card.tsx
+++ b/src/Card/Card.tsx
@@ -1,6 +1,7 @@
 import * as React from "react"
 import { CardHeader, CardItem } from "../"
 import { DefaultProps } from "../types"
+import deprecate from "../utils/deprecate"
 import styled from "../utils/styled"
 
 export interface CardProps<T extends {} = {}> extends DefaultProps {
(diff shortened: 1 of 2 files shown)


TejasQ:
We don't need this if this PR is **Breaking**. Unfortunately the current master already has breaking changes in it and so if we're going to break the API, we need not add deprecation warnings to a new major version.

I think going forward, we should carefully merge non-breaking changes (easier since PRs are more specific), do a _minor_ release with deprecation warnings, _then_ merge breaking PRs and do a new major, breaking release.

peterszerzo:
This PR is a patch change - I thought the intention was to release a new version with this deprecation warning, which would help upgrading UI's, and merging #675 a bit later. Is this not what you meant?

TejasQ:
Ah, unfortunately that's not what I meant. Sorry for the misunderstanding. I meant, when we know something's going to break,

1. merge all non-breaking changes into master
1. add deprecation warnings for expected breakages in new major version
1. `npm version minor`
1. `npm publish`
1. work on and merge breaking changes
1. remove deprecation warnings
1. `npm version major && npm publish`

The problem is, our current master _already has_ breaking changes in it, and so the next release _will have to be a breaking one_ because we forgot steps 1-4 in this case. 

I _could_ remove breaking changes from the master but I'd rather not mess with its commit history for the sake of better semver. I think following this flow for future releases will help though. 🚀 

Does that make sense? 🤔

peterszerzo:
Makes sense - abandoning this PR then. @fabien0102 please feel free to call me over if you see tricky behavior in `Card`'s while upgrading to the upcoming major version.

PR #687, opened 2018-08-22
**Feature: ** Change actionsPosition options to start, main and end.

ImogenF:
## Summary

`actionsPosition` can now take 3 values: "start", "main" and "end" (defaults to "main").
This makes it more compatible with the condensed title option. 

## To be tested

Me

Tester 1

- [x] Every action position works, regardless of whether title is condensed or not.
  

Tester 2

- [ ] Every action position works, regardless of whether title is condensed or not.

### DIFF
--- a/src/Page/Page.tsx
+++ b/src/Page/Page.tsx
@@ -15,7 +15,7 @@ export interface BaseProps extends DefaultProps {
   /** Page actions, typically `condensed button` component inside a fragment */
   actions?: React.ReactNode
   /** Actions position */
-  actionsPosition?: "start" | "end"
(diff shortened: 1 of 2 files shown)


src/Page/Page.tsx
```diff
@@ -121,12 +115,20 @@ const ActionsContainer = styled("div")<{ actionPosition: PageProps["actionsPosit
   ({ theme, actionPosition }) => ({
     ...(actionPosition === "start"
       ? {
+          order: -1,
```
TejasQ: @ImogenF I did not know this exists you are a genius

PR #694, opened 2018-08-23
**Fix:** Fix form style

fabien0102:
### Before
image

### After
image

## To be tested

Me

Tester 1

Tester 2

### DIFF
--- a/src/Form/Form.tsx
+++ b/src/Form/Form.tsx
@@ -9,11 +9,11 @@ import * as React from "react"
 
 import styled from "../utils/styled"
 
-export const Form = styled("form")({
+export const Form = styled("form")(({ theme }) => ({
   "> *:not(:last-child)": {
-    marginBottom: 34,
+    marginBottom: 34 - theme.space.small,
     display: "block",
   },
-})
+}))
 
 export default Form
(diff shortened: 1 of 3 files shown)


src/Input/Input.tsx
```diff
@@ -65,6 +65,7 @@ const InputFieldContainer = styled("div")<{
   align-items: center;
   justify-content: center;
   ${({ fullWidth, withLabel, theme }) => `
+    margin-bottom: ${theme.space.small}px;
```
peterszerzo: I find this a bit tricky to follow - the value `theme.space.small` should play nicely with the `34 - ${theme.space.small}` value from a different file. Is there a way we can just make sure that this component doesn't apply margins and leave the clean 34 pixel assignment to `Form`?

src/Form/Form.tsx
```diff
@@ -9,12 +9,14 @@ import * as React from "react"
...
-export const Form = styled("form")({
+export const Form = styled("form")(({ theme }) => ({
   "> *:not(:last-child)": {
-    marginBottom: 34,
+    marginBottom: 34 - theme.space.small,
     display: "block",
-    width: "fit-content",
   },
-})
+  "> *:not(:last-child) > label": {
```
peterszerzo: Can we add a comment here to clarify the intent?

PR #713, opened 2018-08-30
**Feature:** Card tabs

peterszerzo:
# Summary

This feature adds card tabs that work identically to page tabs:

screen shot 2018-08-30 at 9 46 56 am

Progress:
- [x] Extract tabs from pages into internals
- [x] Implement card tabs feature with identical API
- [x] Improve typings
- [x] `icon` and `iconColor` feature

# To be tested

Me
- [x] Controlled card tabs work
- [x] Uncontrolled card tabs work
- [x] Other card features work as expected
- [x] Page tabs still work as expected
- [x] No layout regressions

Tester 1

- [x] Controlled card tabs work
- [x] Uncontrolled card tabs work
- [x] Other card features work as expected
- [x] Page tabs still work as expected
- [x] No layout regressions

Tester 2

- [ ] Controlled card tabs work
- [ ] Uncontrolled card tabs work
- [ ] Other card features work as expected
- [ ] Page tabs still work as expected
- [ ] No layout regressions


peterszerzo:
@TejasQ you're totally 100% right - neither `Card` nor `Page` need to hold tab state - they can either control the `Internals/Tabs` component or rely on its local state one level down. Amazing amazing catch!!

TejasQ:
@peterszerzo yeah! In my example, `<Tabs>` was stateful and used render props to solve exactly that!

peterszerzo:
I see, I get it now - I will do that then!

peterszerzo:
[Card header tabs](https://trello.com/c/1By5iUWw/375-card-header-tabs)

src/Internals/CardHeader.tsx
```diff
@@ -38,11 +38,6 @@ const Container = styled("div")(({ theme }) => ({
     alignItems: "center",
     textDecoration: "none",
   },
-  "& svg": {
```
fabien0102: This was to deal with case:
image

@peterszerzo Can you update this example with a real icon name? (`ExternalLink` was renamed by `Open` but this example not)

peterszerzo: I'll take out the example altogether - you'll need a component for the `Learn how + icon` element for your use-case, either in Operational or a project that uses it.

fabien0102: Without a fix or a new component for this, I can't really accept this PR, all my CardHeader will be in a broken state…

peterszerzo: The scope of this PR is to add tabs to cards. I think it's reasonable to assume that there is work involved in upgrading from one version to the next, and as far as that text+icon component is concerned, it can start out in the project at hand and move over later once we know what features it needs.

I'd be wary of prematurely piling components into this project just to support a need that was previously implemented with a problematic CSS rule (remember all the other issues we've had with this selector in the past).

fabien0102: Sure, but the scope of every PR with `**Feature:**` prefix is also to not introduce breaking change or regression. This little icon, even if it's little, it's a regression :wink:

peterszerzo: Wouldn't 85% of our patch changes be regressions by that logic? I would agree with you if the button disappeared or didn't link anymore, but in this case the resulting piece of UI will be perfectly usable despite the profoundly unacceptable aesthetics :)

TejasQ: > The scope of this PR is to add tabs to cards.

You said it best, Peter. Leave the styling in for now and let's discuss how to handle this case of SVGs in `CardHeader`s in an issue – be it a component or something else.

peterszerzo: To summarize an offline discussion with @TejasQ : we've decided to keep the `& svg` styles out, and add a `SimpleLink` component that takes care of the text + link aspect. We could even get fancy and find `Icon` components inside the `action` prop at runtime to render a deprecation warning that would help with the upgrade (let me know)

TejasQ: We can't add the deprecation warning in this PR though because this PR breaks the API :/

TejasQ: To add to what Peter said already,

>  and add a SimpleLink component that takes care of the text + link aspect. 

in this PR.

peterszerzo:
@TejasQ @fabien0102 I've added `SimpleLink` as below:

screen shot 2018-09-05 at 4 42 45 pm

TejasQ:
Great! Can you please re-add [this case](https://github.com/contiamo/operational-ui/pull/713#discussion_r215255784) with the new component?

fabien0102:
@peterszerzo  Since you are on it, you can also do a global replace for `icon="ExternalLink"` to `icon="Open"` :angel:

TejasQ:
+1. That should've been done when the Icons were refactored.

peterszerzo:
Good thing there's wonderful tools like *VIMGREP* to help me with this arduous task 🤓

src/SimpleLink/SimpleLink.tsx
```diff
@@ -0,0 +1,102 @@
...
+      marginLeft: theme.space.base,
+    },
+  }
+}
+
+const Container = styled("p")(containerStyles)
+const ContainerLink = styled("a")(containerStyles)
+
+const SimpleLink: React.SFC<SimpleLinkProps> = ({ to, children, icon, color, onClick, left, right, ...props }) => {
+  const ContainerComponent: React.ComponentType<any> = to ? ContainerLink : Container
```
fabien0102: Is this `: React.ComponentType<any>` really needed?

TejasQ: It can be inferred, no?

peterszerzo:
How does this look:

screen shot 2018-09-05 at 5 14 42 pm

src/SimpleLink/SimpleLink.tsx
peterszerzo: It can be - to be honest, I don't remember why we had them, but taking them out the compiler is happy, so it's all good :) best guess is that emotion typings are better than glamorous ones!

TejasQ:
In response to [this](https://github.com/contiamo/operational-ui/pull/713#issuecomment-418766143) – @lukehefson, can we please have a facepalm emoji here? 
image

TejasQ:
In response to [this comment](https://github.com/contiamo/operational-ui/pull/713#issuecomment-418768856),

there's a bit too much space to the left of the link. The voice in my head reads it with a bit of a pause.

TejasQ:
@lukehefson, actually threads in these conversations could be nice to have.

peterszerzo:
The spacing has been increased @TejasQ

TejasQ:
Increased?!

peterszerzo:
Since this component should work in non-text-based contexts too (next to a button etc.), I think pixel-based margins read clearer here.

Issue #697, opened 2018-08-23
Bug: Paginator is missaligned


fabien0102:
![image](https://user-images.githubusercontent.com/1761469/44537524-5784ea80-a6ff-11e8-965b-c76b17a5269e.png)

nick-barth:
This would appear to be a perfect use-case for flexbox.  It also appears you have already fixed it, judging from

nick-barth:
It appears there is a conditional here that should be removed, while I am unfamiliar with the syntax, and unsure of the history/context behind `role="button",`, I'd still wager that we should be putting flex on the container every time.

https://github.com/contiamo/operational-ui/blob/f59c010633c3503c77787838fa6271756e4f3ace/src/Paginator/Paginator.tsx#L86-L90

fabien0102:
I think that we can totally refactor this component, I also don't have the history of this `[role=button]` but it's totally overkill for this case ^^ (and also not anymore working…).

peterszerzo:
This issue is ok to close, right @fabien0102?

PR #720, opened 2018-09-04
  **Fix:** Alignment in Paginator

nick-barth:
# Summary
Was ugly, now beautiful.
image

# Related issue

#697 

# To be tested

Me

Tester 1

- [ ] Give it the good ol' eyeball

### DIFF
--- a/src/Paginator/Paginator.tsx
+++ b/src/Paginator/Paginator.tsx
@@ -83,12 +83,10 @@ const getRange = ({ page, itemCount, itemsPerPage }: PaginatorProps) => {
 const Container = styled("div")(({ theme }) => ({
   label: "paginator",
   marginTop: theme.space.element,
-  "& [role=button]": {
-    display: "flex",
-    alignItems: "center",
-    justifyContent: "center",
-    minWidth: 25,
-  },
+  display: "flex",
+  alignItems: "center",
+  justifyContent: "center",
+  minWidth: 25,
 }))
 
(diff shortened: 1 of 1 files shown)


nick-barth:
I haven't had any experience using Jest before, but due to good documentation on the part of you & jest it was easy to figure out, I think. 👏TRUST 👏THE 👏PROCESS 👏

src/Paginator/Paginator.tsx
```diff
@@ -83,12 +83,9 @@ const getRange = ({ page, itemCount, itemsPerPage }: PaginatorProps) => {
...
   marginTop: theme.space.element,
-  "& [role=button]": {
-    display: "flex",
-    alignItems: "center",
-    justifyContent: "center",
-    minWidth: 25,
-  },
+  display: "flex",
+  alignItems: "center",
+  minWidth: 25,
```
fabien0102: A little last question ^^ Is it useful?

nick-barth: I assume you are talking about min-width, short answer is I'm unsure, but it my reasoning was that it can't hurt it, and as I'm unaware of the use-cases of this component, it may save it visually in some circumstances I could see, so I kept it.  I can drop it like a bad habit if that's what we want, however.

TejasQ:
@nick-barth a rebase here and we can merge this!

TejasQ:
@fabien0102 pointed out that I could just rebase and merge from here so _teamwork ftw_

Issue #708, opened 2018-08-29
Dealing with unstyled text


TejasQ:
# Problem
So we currently use React components that render regular old text DOM nodes a la `<h1>`, `<span>`, etc.. By default, we _do not_ style these nodes in Operational, but instead expose a suite of Typography components: `<Body>`, `<Heading1>`, etc. that _are_ among the former nodes but with styling.

Because of this, the regular old DOM nodes are unstyled. This looks pretty ugly.

# Proposed Solution
I've thought of the following two solutions:

## Solution 1: `<StyledText>`
A component that wraps _whatever_ is in it, and applies the same styling as `Heading1` does to `h1`s to the underlying DOM node. In terms of implementation, it would be 

```jsx
const StyledText = styled("div")`
  h1 {
    whatever happens in Heading1's styles here
  }

  span {
    you get the idea
  }
`
```

Anything in that `div` would then look as expected. I am not in favor of this approach because:
1. It adds an extra node to the DOM and to the React tree.
1. _Still_ allows breaking Operational's opinionation because we'll have unstyled text in places where this has been forgotten.

## Solution 2: Using `<OperationalUI>` the way it was intended
If we hoist these styling settings to the top-level provider, we have an opinionated library that styles _all text_ out of the box, and does not require wrapping text in React components like React Native does. 

I feel this is more of an idiomatic approach: wrap once, forget about it – your text is always beautiful.

Thoughts?

peterszerzo:
I find solution 1 more in line with the styled components philosophy, where styles are localized and we don't have to worry about which global style fights which other global or local style. Have we identified a need for other contexts in which opinionated static-markup styles are required, other than one in which I put some rendered markdown inside a `<StyledText>` tag? If not, I think it's best to keep these styles scoped. Otherwise, we can look into re-using this component or creating another before adding more to our global styles.

fabien0102:
Why just not using `<Heading1>` and `<Body>` as originally design? For me you should always use a react component for styling text instead of `<h1>` or `<span>`.

I propose to add a ts-lint rule of somebody is using a native dom balise in our project to enforce this rule 🤠

TejasQ:
> So we currently use React components that render regular old text DOM nodes

@fabien0102 , we use React Markdown in some places, that parses Markdown and renders raw `<h1>`s. There is no way around this unless we write our own markdown parser for Operational.

[...]

PR #721, opened 2018-09-04
**Feature:** Add StyledText component.

ImogenF:
# Summary

New `<StyledText>` component that adds operational styling to text that is not otherwise styled by operational, e.g. rendered React Markdown. 

<blockquote class="trello-card"><a href="https://trello.com/c/S40wmFO0/388-improve-display-of-bundle-readme-markdown">Improve display of Bundle Readme (markdown)</a></blockquote>
 
# Related issue

#708 

# Comments

In some cases, it was unclear what exact CSS properties certain HTML elements should have, for example `<h2>` vs `<h3>`. 

The styling is very much open to discussion.

# To be tested

Me

Tester 1

Tester 2


src/StyledText/StyledText.tsx
```diff
@@ -0,0 +1,41 @@
+import * as React from "react"
+import styled from "react-emotion"
+
+import { DefaultProps } from "../types"
+
+export interface StyledTextProps extends DefaultProps {
```
peterszerzo: What do we think about this component (optionally) handling markdown parsing at the same time? I think we favored that approach the last time we talked about this offline.

src/StyledText/StyledText.tsx
```diff
@@ -0,0 +1,41 @@
...
+    fontSize: theme.font.size.title,
+    fontWeight: theme.font.weight.medium,
+    lineHeight: theme.font.lineHeight,
+    color: theme.color.text.dark,
+    width: "100%",
+    margin: `0 0 ${theme.space.element}px 0`,
+    padding: `0 0 ${theme.space.medium}px 0`,
+    borderBottom: `1px solid ${theme.color.separators.default}`,
+  },
+  "& h2, h3, h4, h5, h6": {
```
peterszerzo: There should be at least a color difference between `h2` and `h3` - we can make `h3` a little lighter. The even lower-level tags `h4`, `h5` and `h6` can be the same as `h3` but italic, just to show that we don't support strong distinctions between them.

src/StyledText/StyledText.tsx
```diff
@@ -0,0 +1,41 @@
...
+    fontFamily: theme.font.family.main,
+    fontSize: theme.font.size.body,
+    fontWeight: theme.font.weight.regular,
+    lineHeight: theme.font.lineHeight,
+    color: theme.color.text.default,
+    margin: `0 0 ${theme.space.element}px 0`,
+  },
+}))
+
+const StyledText: React.SFC<StyledTextProps> = ({ children }) => <Container>{children}</Container>
```
peterszerzo: We should do the props spreading here: `({ children, ...props }) => <Container {...props}>{children}</Container>`

src/StyledText/StyledText.tsx
ImogenF: You mean instead of having something like this: 

<StyledText>
  <ReactMarkdown source={markdown} />
</StyledText>

we could have:

<StyledText parseMarkdown>
  {markdown}
</StyledText>

?

src/StyledText/README.md
```diff
@@ -0,0 +1,57 @@
...
+  <h2>jupyter</h2>
+  <p>open the jupyter website</p>
+</StyledText>
+```
+
+### Usage with Markdown
+
+```jsx
+<StyledText
+  markdown={
```
TejasQ: This isn't an API I would expect. What is more idiomatic to me is:

```jsx
<StyledText>
  <ReactMarkdown value={markdown} />
</StyledText>
```

That way, `StyledText` is agnostic to the _type_ of text it renders and just styles everything in there the way we expect it to be.

It also allows us to not have `react-markdown` as a dependency in the project.

ImogenF: This is what I had originally. I just changed it based on @peterszerzo's comments above. As it stands, you can style markdown either using the API you suggest, or using `<StyledText markdown={...}/>`

ImogenF: Easy enough to revert the commit though. I'm pretty indifferent in this case - I feel like both approaches have their merits.

peterszerzo: @TejasQ what do you find unidiomatic about `<StyledText markdown={""} />`? It find that it reads pretty clear, and addresses the most common use-case we'd have for this component.

Generally, I would like to see more markdown supported across this project - it works really well if you need to add tutorial-style `Learn more in our [wiki](http://operationalwiki.com/resource) page` comments to form field hints, error messages etc.

TejasQ: I find this approach unnatural because `StyledText` is expected to style its children, irrespective of what they may be. 

The `markdown` prop seems fairly out of place because all markdown is ultimately just text and can (and I'd argue _should_) be in children. 

I also would argue that the most common use case of this component is not to style _markdown_, but to style _any and all possible text_ inside it: some of it can be markdown, others could be arbitrary headings and paragraphs – the component itself does not care and does its job, well.

We can discuss markdown support in a separate issue. As for this component, I would prefer it to style whatever's inside – which could be some future markdown component of ours even, at a later stage.

TejasQ:
[Improve display of Bundle Readme (markdown)](https://trello.com/c/S40wmFO0/388-improve-display-of-bundle-readme-markdown)

src/StyledText/StyledText.tsx
```diff
@@ -0,0 +1,59 @@
...
+    color: theme.color.text.default,
+    fontStyle: "italic",
+  },
+  "& p": {
+    ...baseStyles(theme),
+    fontFamily: theme.font.family.main,
+    fontSize: theme.font.size.body,
+    fontWeight: theme.font.weight.regular,
+    color: theme.color.text.default,
+  },
```
peterszerzo: Can the `& blockquote` quote get a 2px grey border on the left and a slightly smaller left padding?

src/StyledText/StyledText.tsx
peterszerzo: @ImogenF I just noticed that this got lost in the comments list. Can we make this change, or is there a reasoning behind not forwarding props?

src/StyledText/StyledText.tsx
peterszerzo: If there's a left border, I think it's best to avoid the left margin.

src/StyledText/StyledText.tsx
ImogenF: Oh, good spot. I did implement but it was in the commit with markdown parsing which I've now deleted. Will add that back in.

src/StyledText/StyledText.tsx
ImogenF: So the border left-aligned with the text? Ok.

TejasQ:
Does it make sense to close this in favor of the newer Markdown component we discussed yesterday?

PR #731, opened 2018-09-07
**Fix:** Force the page to never never scroll down

fabien0102:
# Summary
 
1aa19d45-1b20-4d46-8a82-5b5c23d0d344

For a really obscur reason, when I click to this checkbox, the `Page:ViewContainer` scroll down and I loose my modal 😢 

This PR add this workaround to enforce this `ViewContainer` to always be on top.

# To be tested

Me

Tester 1

- [ ] No regression on any Page cases

Tester 2

- [ ] No regression on any Page cases

### DIFF
--- a/src/Page/Page.tsx
+++ b/src/Page/Page.tsx
@@ -4,6 +4,7 @@ import Tabs, { Tab, tabsBarHeight } from "../Internals/Tabs"
 import PageArea from "../PageArea/PageArea"
 import PageContent, { PageContentProps } from "../PageContent/PageContent"
 import { DefaultProps } from "../types"
+import constants from "../utils/constants"
 import styled from "../utils/styled"
 
 export interface BaseProps extends DefaultProps {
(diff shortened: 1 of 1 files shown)


src/Page/Page.tsx
```diff
@@ -83,13 +84,43 @@ const TitleContainer = styled("div")(({ theme }) => ({
...
+ * Page. It includes a bit of hack that prevents it
+ * from being scrolled at all: the component is meant
+ * to have a full-height and a hidden overflow, while
+ * its _children_ scroll, not it itself.
+ *
+ * In order to enforce this, we use the `onScroll` hack.
+ *
+ * More info: https://github.com/contiamo/operational-ui/pull/731
+ */
+class ViewContainer extends React.Component<{ isInTab?: boolean; isTitleCondensed?: boolean }> {
```
TejasQ: Would

```jsx
const ViewContainer: React.SFC<{ isInTab?: boolean; isTitleCondensed?: boolean }> = ({
  isInTab,
  isTitleCondensed,
  ...props
}) => {
  const viewContainerRef = React.createRef<HTMLDivElement>()
  return (
    <div
      style={{
        height: `calc(100% - ${
          isInTab && !isTitleCondensed ? constants.titleHeight + tabsBarHeight : constants.titleHeight
        }px)`,
        overflow: "hidden",
        position: "relative",
      }}
      ref={viewContainerRef}
      onScroll={() => {
        if (viewContainerRef.current) {
          viewContainerRef.current.scrollTop = 0
        }
      }}
      {...props}
    />
  )
}
```

work here instead?

fabien0102: In this case we are creating a `ref` on each render… even I'm not a huge a fan of React.Component, it's actually our best option for this work around.

TejasQ:
@fabien0102 are there any developments here?

fabien0102:
Impossible to reproduce on codesandbox… so not really, I don't really understand the `overflow: hidden` toggle, this div is already in overflow hidden but scroll without any obvious reasons.

I also don't like to not understand what and how it can happens, but this hack it's quite safe because this div must always be on a top (height 100% + overflow hidden)

TejasQ:
I think Peter's right – we should be able to reproduce it before we try to fix it. I'll try to do this and respond here if/when I'm able to have a working reproduction.

PR #736, opened 2018-09-11
**Feature:** Add cmd+enter submit to text area

peterszerzo:
# Summary

Text areas will now call a handler provided to their `onSubmit` prop when cmd+enter is pressed inside.

# To be tested

Me
- [x] Textarea component reacts to cmd+enter on Chrome, Safari and Firefox
- [x] No regressions on Textarea

Tester 1

- [x] Netlify looks good
- [x] Textarea component reacts to cmd+enter in my own environment
- [x] No regressions on Textarea

Tester 2

- [ ] Textarea component reacts to cmd+enter in my own environment
- [ ] No regressions on Textarea

### DIFF
--- a/src/Textarea/Textarea.tsx
+++ b/src/Textarea/Textarea.tsx
@@ -35,6 +35,8 @@ export interface TextareaProps extends DefaultProps {
   resize?: ResizeOptions
   /** Copy text to clipboard on click */
   copy?: boolean
+  /** cmd+enter submit handler */
+  onSubmit?: () => void
 }
 
 export interface State {
@@ -144,6 +146,7 @@ class Textarea extends React.Component<TextareaProps, State> {
       height,
       copy,
       onChange,
+      onSubmit,
       ...props
     } = this.props
     return (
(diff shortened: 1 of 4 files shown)


peterszerzo:
@micha-f how should we handle letting the user know that cmd+enter submit is available? Is this something we should show in e.g. the lower right corner if the `onSubmit` prop is present on the level of this component? Or more like the responsibility of the user to show it somewhere else or in the hint?

src/Textarea/Textarea.tsx
```diff
@@ -162,6 +165,14 @@ class Textarea extends React.Component<TextareaProps, State> {
           isAction={Boolean(action || copy)}
           resize={resize!}
           height={height}
+          onKeyDown={(ev: React.KeyboardEvent<HTMLTextAreaElement>) => {
+            const crossBrowserSafeKeycode = ev.which || ev.keyCode
```
TejasQ: This is incredible!

peterszerzo:
Good catches, @TejasQ, I'll update!

peterszerzo:
@TejasQ I added an identical cmd+enter functionality to the `Form` component as well.

TejasQ:
> Textarea component reacts to cmd+enter in my own environment

What does this mean?

peterszerzo:
Mac if you're on a mac, windows if you're on a PC.

TejasQ:
Interesting. Are we expecting it to behave differently in different environments? Is it because of the cross-browser check?

ImogenF:
I feel like I'm missing something here - cmd + enter doesn't seem to have any impact on the text area for me. What exactly should happen?

Also, the actions bar on the text area seems to have become misaligned at some point (this is true in master as well though, not a regression introduced in this PR)
image

peterszerzo:
@TejasQ I asked for that check in the review because I am not so intimately familiar with modifier keys in key events, and with how it works with the ctrl key on windows. Thanks for checking @fabien0102!

PR #739, opened 2018-09-12
**Feature:** Save my yarn lock

fabien0102:
# Summary

image

No more `yarn.lock` persecution 🤠 

# Related issue

# To be tested

Me

Tester 1

- [ ] I can't kill this poor yarn.lock
- [ ] I can also update the yarn.lock

Tester 2

- [ ] I can't kill this poor yarn.lock
- [ ] I can also update the yarn.lock


fabien0102:
image

TejasQ:
Nice try, but it didn't take much to break this.

image

TejasQ:
I think that doesn't quite make sense because _some_ projects use `yarn`, others `npm`. Because of the open `yarn` issue about postinstall scripts, I'm wondering if we should just use `npm` everywhere.

PR #741, opened 2018-09-13
**Fix:** Add correct type annotation to Form

TejasQ:
# Summary
This PR fixes a regression in #736 that breaks the `Form` component's public interface.

## Regression
image

# To be tested

Me

Tester 1

- [ ] An app that uses `Form` from `@operational/components@canary` compiles.

### DIFF
--- a/src/Form/Form.tsx
+++ b/src/Form/Form.tsx
@@ -21,7 +21,7 @@ const Container = styled("form")(({ theme }) => ({
  * to be re-used across different node types (both `HTMLFormElement` and `HTMLElement`
  * cause issues here).
  */
-const Form: React.SFC = (props: React.HTMLProps<any>) => (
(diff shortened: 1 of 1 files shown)


src/Form/Form.tsx
```diff
@@ -21,7 +21,7 @@ const Container = styled("form")(({ theme }) => ({
  * to be re-used across different node types (both `HTMLFormElement` and `HTMLElement`
  * cause issues here).
  */
-const Form: React.SFC = (props: React.HTMLProps<any>) => (
+const Form: React.SFC<typeof Container> = (props: React.HTMLProps<any>) => (
```
fabien0102: 🤔 `React.SFC<Props>` vs `(props: Props)`

You need to choose 😉

TejasQ: Good catch. Thank you!

TejasQ:
Done. Thanks, y'all!

Issue #684, opened 2018-08-21
Add fallback to `window` in order to support SSR


TejasQ:
Currently, `OperationalUI` uses a lot of properties on `window.` to handle resize, etc. The problem here is that when we try to render Operational UIs server-side, there _is no `window` object_ and so the entire app breaks.

Solving this is a one-liner and would just add a fallback to an empty object to the component.

adeelibr:
Reference: https://stackoverflow.com/a/2671217/4921319

Wouldn't it be better if we removed `window.` all together & instead use `document.body.` to bind event listeners. 

> Webkit and mozilla browsers seem to be happy with either, but IE doesn't respond to most events registered on the window, so you need to use document to work with IE

Also as a fact I can tell that in React-Material-UI they are not using `window.` to bind event handlers. I am assuming it for the same reason as this for SSR support.

TejasQ:
`document.body.` presents the same problem in a node environment unless some type of JSDom abstraction is used, right? I'm not sure that solves the problem.

How does React-Material-UI attach event handlers while maintaining SSR?

I think the easiest solution would be:

`var window = window || { addEventListener: () => {} }` at the top of `OperationalUI.tsx`.

adeelibr:
On checking the `Input` component, `autoComplete` prop is already available. 😅 

Typings are already added

https://github.com/contiamo/operational-ui/blob/67ab23ed89f16e8e92286d0bfeba5cb7b713ed2a/src/Input/Input.tsx#L29

And then used here like

https://github.com/contiamo/operational-ui/blob/67ab23ed89f16e8e92286d0bfeba5cb7b713ed2a/src/Input/Input.tsx#L238

https://github.com/contiamo/operational-ui/blob/67ab23ed89f16e8e92286d0bfeba5cb7b713ed2a/src/Input/Input.tsx#L250

adeelibr:
Although while looking at the code, I think `Input` should be self contained, while the component that take `Input` along with `label` should be a separate component. What do you think?

adeelibr:
> document.body. presents the same problem in a node environment unless some type of JSDom abstraction is used, right? I'm not sure that solves the problem.

I don't think so, since in SSR when you hydrate the ReactDOM you set it with `document.getElementById('app`)` So I think `document.body` is available on a Node instance as well. I went through the entire Material UI code base. They are simply attaching event's to `document.body` without any check like.

> var window = window || { addEventListener: () => {} }

[...]

PR #754, opened 2018-09-23
**Feature:** OperationalUI SSR Compliant

adeelibr:
# Summary
* Added support for SSR in OperationalUI component

# Related issue

https://github.com/contiamo/operational-ui/issues/684

# To be tested

Me
- [X] No error or warning in the console on `localhost:1234`

### DIFF
--- a/src/OperationalUI/OperationalUI.tsx
+++ b/src/OperationalUI/OperationalUI.tsx
@@ -124,7 +124,7 @@ class OperationalUI extends React.Component<OperationalUIProps, State> {
    *  whether any messages need to be removed from state
    */
 
-  public messageTimerInterval: ReturnType<typeof setTimeout> | null = null
+  public messageTimerInterval: ReturnType<typeof setInterval> | null = null
 
   public removeOutdatedMessages() {
     if (this.props.hideMessageAfter === 0) {
--- a/src/OperationalUI/OperationalUI.tsx
+++ b/src/OperationalUI/OperationalUI.tsx
@@ -112,8 +112,8 @@ class OperationalUI extends React.Component<OperationalUIProps, State> {
 
   public state: State = {
     windowSize: {
-      width: window.innerWidth,
-      height: window.innerHeight,
+      width: 0,
+      height: 0,
     },
     messages: [],
     isLoading: false,
@@ -124,7 +124,7 @@ class OperationalUI extends React.Component<OperationalUIProps, State> {
    *  whether any messages need to be removed from state
    */
 
-  public messageTimerInterval: number | null = null
+  public messageTimerInterval: number | null | any = null
 
   public removeOutdatedMessages() {
     if (this.props.hideMessageAfter === 0) {
@@ -137,7 +137,7 @@ class OperationalUI extends React.Component<OperationalUIProps, State> {
 
     // If we're out of messages, clear the interval.
     if (!filteredMessages.length && this.messageTimerInterval) {
-      window.clearInterval(this.messageTimerInterval)
+      clearInterval(this.messageTimerInterval)
       this.messageTimerInterval = null
     }
     // Only run a setState if any message(s) were removed.
@@ -152,16 +152,17 @@ class OperationalUI extends React.Component<OperationalUIProps, State> {
    * @todo look into making this unnecessary.
    */
   public handleResize: (() => void) & Cancelable = debounce(() => {
-    this.setState(() => ({
-      windowSize: {
-        width: window.innerWidth,
-        height: window.innerHeight,
-      },
-    }))
+    this.onSetWindowSize()
   }, 200)
 
   public setLoading = (isLoading: boolean) => {
     this.setState(() => ({ isLoading }))
+  }
+
+  public onSetWindowSize = () => {
+    this.setState(() => ({
+      windowSize: { width: window.innerWidth, height: window.innerHeight },
+    }))
   }
 
   public componentDidCatch(error: Error) {
@@ -175,14 +176,15 @@ class OperationalUI extends React.Component<OperationalUIProps, State> {
     if (!this.props.noBaseStyles) {
       injectGlobal(baseStylesheet(constants))
     }
-    window.addEventListener("resize", this.handleResize)
+    this.onSetWindowSize()
+    document.body.addEventListener("resize", this.handleResize)
   }
 
   public componentWillUnmount() {
     if (this.messageTimerInterval) {
-      window.clearInterval(this.messageTimerInterval)
+      clearInterval(this.messageTimerInterval)
     }
-    window.removeEventListener("resize", this.handleResize)
+    document.body.removeEventListener("resize", this.handleResize)
   }
 
   public render() {
@@ -203,7 +205,7 @@ class OperationalUI extends React.Component<OperationalUIProps, State> {
 
                 // If we don't yet have an interval, start one.
                 if (!this.messageTimerInterval) {
-                  this.messageTimerInterval = window.setInterval(() => this.removeOutdatedMessages(), 2000)
+                  this.messageTimerInterval = setInterval(() => this.removeOutdatedMessages(), 2000)
                 }
               },
               loading: this.state.isLoading,
--- a/src/OperationalUI/OperationalUI.tsx
+++ b/src/OperationalUI/OperationalUI.tsx
@@ -124,7 +124,7 @@ class OperationalUI extends React.Component<OperationalUIProps, State> {
    *  whether any messages need to be removed from state
    */
 
-  public messageTimerInterval: number | null | any = null
+  public messageTimerInterval: ReturnType<typeof setTimeout> | null = null
 
   public removeOutdatedMessages() {
     if (this.props.hideMessageAfter === 0) {


src/OperationalUI/OperationalUI.tsx
```diff
@@ -112,8 +112,8 @@ class OperationalUI extends React.Component<OperationalUIProps, State> {
 
   public state: State = {
     windowSize: {
-      width: window.innerWidth,
-      height: window.innerHeight,
+      width: 0,
+      height: 0,
```
adeelibr: `windowSize` for width & height was initially kept 0, because on initial server side render `window` object is not available. `windowWidth` is set up after first render in `componentDidMount` life cycle.

src/OperationalUI/OperationalUI.tsx
```diff
@@ -152,18 +152,19 @@ class OperationalUI extends React.Component<OperationalUIProps, State> {
...
-      },
-    }))
+    this.onSetWindowSize()
   }, 200)
 
   public setLoading = (isLoading: boolean) => {
     this.setState(() => ({ isLoading }))
   }
 
+  public onSetWindowSize = () => {
```
adeelibr: Made a generic method to be used in `componentDidMount` along with when `handleResize` is called.

src/OperationalUI/OperationalUI.tsx
```diff
@@ -175,14 +176,15 @@ class OperationalUI extends React.Component<OperationalUIProps, State> {
     if (!this.props.noBaseStyles) {
       injectGlobal(baseStylesheet(constants))
     }
-    window.addEventListener("resize", this.handleResize)
+    this.onSetWindowSize()
+    document.body.addEventListener("resize", this.handleResize)
```
adeelibr: `document` instead of `window` because of better support with Internet Explorar

src/OperationalUI/OperationalUI.tsx
```diff
@@ -124,7 +124,7 @@ class OperationalUI extends React.Component<OperationalUIProps, State> {
    *  whether any messages need to be removed from state
    */
 
-  public messageTimerInterval: number | null = null
+  public messageTimerInterval: number | null | any = null
```
adeelibr: Since this is no longer explicitly referring to the `setInterval` of object `window` or the node 'global` instance` I had to set the type to `any`, I am open to suggestions for this?

TejasQ: So I would get more creative about typing this. A union with `any` is debatably loose enough that you could just use `any`, which may not be ideal.

What is it? It's either: `setInterval` on `window`, which is a `number`, or `null`, or what else? What are the possible values it can have?

TejasQ:
Thanks so much for your contribution! I'm glad you were able to work with this repository. Was it easy?

I love this PR! It looks really promising. One thing I'd add is maybe leave comments _in code_ as you're leaving comments on the PR in order to better describe intent. I think that could be a value add since they're free, memory-wise, and help readers of the code years from now where the PR would be a little less readily accessible.

Besides that, your work is amazing! Let's review a bit and get it merged! 🎉

src/OperationalUI/OperationalUI.tsx
adeelibr: That's the thing now, since it's not binded with `window.setTimeout` or `global.setTimeout` Typescript can't tell what type it is. I did `typeof messageTimerInterval` and the value is `number` nothing else. So I am guessing `any` can be for now, and in the future we can come up with a better strategy to go for instead of using `setTimeout`

> So I would get more creative about typing this. A union with any is debatably loose enough that you could just use any, which may not be ideal.

I was going to keep it only `any` but I wanted your review on this particular matter.

adeelibr: Oh sorry! `binded` might not be the appropriate word. I updated my PR for this.

src/OperationalUI/OperationalUI.tsx
```diff
@@ -124,7 +124,7 @@ class OperationalUI extends React.Component<OperationalUIProps, State> {
    *  whether any messages need to be removed from state
    */
 
-  public messageTimerInterval: number | null = null
+  public messageTimerInterval: ReturnType<typeof setTimeout> | null = null
```
adeelibr: Reference taken from https://stackoverflow.com/a/51040768/4921319

@TejasQ your thoughts?

src/OperationalUI/OperationalUI.tsx
peterszerzo: I wonder if it'd be safer and clearer if we used `undefined` here - I ran into situations on other projects that rendered something like a `This screen size is not supported` banner if the screen size was smaller than a threshold, which the zero values obviously satisfy. If we used `undefined`, on the other hand, such a comparison would be disallowed by the compiler, and we can prevent bugs where a wrong screen size banner is shown for a flash until the window size is set.

Maybe the example is a little silly, but these context props are used by e.g. the `Tooltip` component, and may result in strange visual consequences nonetheless, and preventable if the codebase were forced to handle unavailable window dimensions explicitly.

src/OperationalUI/OperationalUI.tsx
TejasQ: Shouldn't it be `typeof setInterval`?

adeelibr: Sorry silly mistake, I'll fix this. 😅

src/OperationalUI/OperationalUI.tsx
adeelibr: @peterszerzo I am sorry I didn't understand it correctly. How would the compiler dis allow such a comparison? Assuming if it was undefined initially, `render` was called in the lyfe cycle hooks after the constructor was called. The render would still show the UI with width, height undefined and would override the the default contexts of

https://github.com/contiamo/operational-ui/blob/6fe657f27ecae5bf32324d104d514e8dee6db766/src/OperationalContext/OperationalContext.init.tsx#L33

```
windowSize: {
    width: 1080,
    height: 640,
  },
```

So in Tooltip 

https://github.com/contiamo/operational-ui/blob/6fe657f27ecae5bf32324d104d514e8dee6db766/src/Tooltip/Tooltip.tsx#L149

So assuming

https://github.com/contiamo/operational-ui/blob/6fe657f27ecae5bf32324d104d514e8dee6db766/src/Tooltip/Tooltip.tsx#L158


1 < 0 => true
1 < undefined => false

and the `if` branch will never get executed right?

peterszerzo: You're totally right, it's perfectly valid: https://www.typescriptlang.org/play/index.html#src=const%20screenSize%3A%20number%20%7C%20undefined%20%3D%20480%0D%0A%0D%0Aconst%20shouldRender%20%3D%20screenSize%20<%201000

So what I suggested doesn't actually make anything safer. Thanks for investigating it nonetheless!

src/OperationalUI/OperationalUI.tsx
adeelibr: Fixed, can you review again? @TejasQ

TejasQ:
> I see many of the components are smart components (stateful) rather then dumb. Ideally the UI library should all stateless (dumb) except maybe for a couple of edge cases. 

> Every one of them should be a separate module (not a utility) and the Input component should just be Input, So every component should be composable and the rest of the effort should be spent in documentation on communicating how to make composable components that are isolated completely & where the end-user can actually "build on top of it"

The goal with Operational is that we want to provide an _opinionated_ set of UI components: meaning, we abstract away a _lot_, _**even**_ composition where we can, allowing our users (and ourselves) to just put stuff together fairly idiomatically.

This is the justification for some state within components. We've been using them and they've been working quite comfortable in our products, actually.

Are there solid arguments in favor of an alternative? If so, I'd love to hear them. Typically, I dislike "this should x" statements without much reasoning about _why_.

> I really want to get more involved. If that's okay :)

adeelibr:
> Are there solid arguments in favor of an alternative? If so, I'd love to hear them. Typically, I dislike "this should x" statements without much reasoning about why.

> The goal with Operational is that we want to provide an opinionated set of UI components: meaning, we abstract away a lot, even composition where we can, allowing our users (and ourselves) to just put stuff together fairly idiomatically.

The UI approach I like to follow is completely opposite of this, by making it stateless I am trying to make an argument that the user should be in control all the time. I am used to designing/structuring my components that way. But I see that you have a different view on this. (Which I can understand, somewhat..)

As another option, just another 2 cent 😋 . We can use a pattern called "state-reducers", a bit intro on it here. https://blog.kentcdodds.com/the-state-reducer-pattern-%EF%B8%8F-b40316cfac57 This basically illustrates a pattern where if no props are provided by the user, the component itself uses it's internal state. So it can work both way, controlled/uncontrolled. Does this make sense? 

The reason I want to argue on this is, `Operational-UI` this way makes absolute sense because you are using it internally in your company, and it's ideal for it. I feel user's outside would be more up for it, if they had more control on the API of the component.

TejasQ:
> I am trying to make an argument that the user should be in control all the time.

I agree, for a user. For a developer, even _as_ a developer, I find I want to build UIs quickly that _get a product out_, without a lot of mental overhead.

I appreciate your arguing here – it is a _giant_ part of any engineering job and being able to argue well is _quite_ a skill to have. 🙂 

That said,

> Operational-UI this way makes absolute sense because you are using it internally in your company, and it's ideal for it

I don't understand this. The question I find myself asking is _why_.  Why is it that it

> makes absolute sense

?

We're using it internally in _quite a lot of places_, and we're able to develop our products quite rapidly when the complexity of handling a lot of props on a Checkbox, for example, is abstracted away and I'm able to just _build a UI_ quickly and allow it to do its job well, you know?

The goal is to allow us to develop internally and breathe exhales instead of sighs.

TejasQ:
@adeelibr any progress here?

adeelibr:
@TejasQ the PR itself is ready, just waiting for your review. the thing as you said above. I haven't added code documentation here. But I feel the code is self explaining, nothing complicated. Would you like me to document in code the changes I made? :)

TejasQ:
Oh, cool. I just wasn't sure what was next there since there was only a +1 on my comment and no new commit. I'll take a look at the PR again. 🙂

adeelibr:
@peterszerzo The PR is in sync with the latest master.

PR #757, opened 2018-09-24
 **Feature:** [Splash] API for logo, logoSize

adeelibr:
# Summary

This is a work in progress, I just wanted to create a PR .This is in no way ready yet. I'll keep on updating the PR. 

**Splash Features**
- logo prop => This is optional, It expects a react node. You can pass in a svg|jpeg|png or image of any sort along with props for it
   - If `logo` passed down is an SVG some default props are added while preserving the `props` user has set
- logoSize prop => for changing the splash logo size
- if no logo is provided the default logo is shown
- change color of logo if an SVG provided

**Code related**
- rotation logic moved to spash.logo.tsx

**In Progress**
[X] ability to change the background based current theme of splash light | dark
[X] remove setInterval for rotation, and instead implement css keyframe animations, for better performance

f4at70cbcj
 
The GIF is pretty bad, i'll record a better GIF with the next update in this PR 👍 

# Related issue

https://github.com/contiamo/operational-ui/issues/650

Me
- [ X] No error or warning in the console on `localhost:6060`


adeelibr:
Progress:
- Moved the rotation logic from state/setInterval to CSS keyframes
- Added support for logo in any format e.g, svg|png|jpeg|gif etc

src/Splash/Splash.Logo.tsx
```diff
@@ -1,30 +1,84 @@
...
-      <g>
-        <path d="M381.163,687.88c1.029,0.09 1.291,0.09 2.31,0.27c11.663,2.057 19.88,16.053 14.891,27.618c-0.41,0.95 -0.553,1.168 -1.044,2.078l-100,173.205c-0.542,0.88 -0.66,1.113 -1.278,1.943c-5.541,7.444 -16.663,10.126 -25.018,5.93c-9.66,-4.851 -13.758,-17.833 -8.345,-27.873l100,-173.205c3.916,-6.348 11.029,-10.183 18.484,-9.966Z" />
-        <path d="M721.163,98.983c1.03,0.09 1.291,0.09 2.31,0.27c11.748,2.071 19.816,16.201 14.891,27.617c-0.409,0.951 -0.553,1.169 -1.043,2.079l-100.001,173.205c-0.542,0.88 -0.66,1.113 -1.278,1.943c-5.542,7.444 -16.663,10.126 -25.018,5.93c-9.66,-4.852 -13.758,-17.834 -8.345,-27.873l100.001,-173.205c3.915,-6.348 11.029,-10.183 18.483,-9.966Z" />
-      </g>
-    </svg>
-  )
+const Container = styled("div")`
```
adeelibr: Decided to add keyframes instead of doing it using setInterval & updating state.

adeelibr:
@TejasQ  @peterszerzo would love your feedback on this, regarding the animation implementation for the logo.

Todo Till Yet:
[X] - Text colors in the Splash component are currently hard-coded to white, which won't be readable if the background is set to a very bright color. Implement color in a way that it should toggle between white and dark using the `readableTextColor` utility. This will involve tweaking animation colors, logo colors and backdrop colors.

adeelibr:
Added ability to change background color.

kfvmukkrwh

src/Splash/Splash.Logo.tsx
TejasQ: Why?

src/Splash/Splash.Logo.tsx
```diff
@@ -1,30 +1,84 @@
...
+class OperationalLogo extends React.Component<Props, {}> {
+  public static defaultProps = {
+    color: "currentColor",
+    logo: null,
+  }
+
+  public render() {
+    const { size, color, logo: Logo } = this.props
+
+    const defaultStyle = {
```
TejasQ: Why is this in render?

src/Splash/Splash.Logo.tsx
```diff
@@ -1,30 +1,84 @@
...
+      display: "inline-block",
+      animation: "rotation 8s 0s infinite linear",
+    }
+
+    const svgProps = {
+      width: size,
+      height: size,
+      viewBox: "0 0 1000 1000",
+      style: Object.assign({}, defaultStyle),
```
TejasQ: Why `Object.assign` instead of spread?

src/Splash/Splash.Logo.tsx
```diff
@@ -1,30 +1,84 @@
...
+      return (
+        <Container>
+          <Logo.type {...logoProps} />
+        </Container>
+      )
+    }
+
+    return (
+      <Container>
+        <svg {...svgProps}>
```
TejasQ: What's this if it's not a logo? Can we express this a little more descriptively than an arbitrary `svg`?

src/Splash/Splash.Logo.tsx
adeelibr: The result is same, you can now convert the Splash.Logo.tsx into a dumb component. Using keyframes adds more possibility to add various kind of animations if needed be,(in the future) which with the prior technique with `setInterval` would have been a bit complicated I think. (correct me if I am wrong here)

Rather then explicitly changing the state of rotation after every 8s with `setInterval` Using `keyframes` is better, because it implicitly iterates an animation after ever 8s.

This is somewhat relevant I think https://developer.mozilla.org/en-US/docs/Web/Apps/Fundamentals/Performance/CSS_JavaScript_animation_performance (I am not sure)

Also why would you pass an unnecessary rotation `variable` in the styles ?
https://github.com/contiamo/operational-ui/blob/67ab23ed89f16e8e92286d0bfeba5cb7b713ed2a/src/Splash/Splash.Logo.tsx#L17

Where instead the same can be achieved using `key-frames`

src/Splash/Splash.Logo.tsx
adeelibr: I'll update it to spread, Although instead of the `spread` syntax being less verbose. There is no performance implications with using `Object.assign` over `rest`. But since `{...rest}`is easy on the eyes 😄 So i'll change it to rest. 👍

src/Splash/Splash.Logo.tsx
adeelibr: Line no 69, executes only when we provide no logo prop, and since the default logo to be returned is an SVG, which is hard coded, that is why the above implementation is that way.

> What's this if it's not a logo?

Can you please explain on this more, since if it's not a `logo` then the purpose of `Splash.logo.tsx` changes.

src/Splash/Splash.tsx
```diff
@@ -13,15 +13,18 @@ export interface SplashProps {
   /** Main content */
   children: React.ReactNode
   /** Backdrop color, dark enough to support white text as in the example. */
-  color?: string
+  color: string
+  /* Splash Logo */
+  logo?: React.ReactElement<any>
+  /* Size of splash logo */
+  logoSize: number
```
peterszerzo: I don't think this should be configurable - the logo should equal the height of the title + the buttons area underneath (the 50px value), giving it some aesthetic guarantees.

TejasQ:
Why? Our library really shouldn’t bundle images since they’re meant to be used alongside any client-provided images instead of giving them to a library user: Operational UI ought to provide components, not necessarily images. 

What’s your use case? I’d love to understand it better.

src/Splash/Splash.Logo.tsx
TejasQ: Sure. 

My point is that the non-early return is fairly ambiguous. If it’s a fallback, it can be expressed as a fallback. 

Currently in code, there is no indicator of what the `svg` actually is: one would have to look at the component to see what happens when you don’t specify a logo. Instead, I propose making clear what happens if a logo is not specified.

adeelibr:
@TejasQ 

>What’s your use case? I’d love to understand it better.

This is for development purposes, while testing I was trying to pass an `img` element in the `Splash` component to test if it worked accordingly with an `img` element as wel as an `svg` element. I don't propose that you add it in the library, just enable it during development. So that a case like this can be handled.

src/Splash/Splash.tsx
adeelibr: This is an additional param, just in case someone want to modify it. It will work the default way if you don't provide the `logoSize` prop. Should it still be removed? It doesn't change the default behavior, just provided a way if someone wants to. @peterszerzo

src/Splash/Splash.Logo.tsx
adeelibr: > Can we express this a little more descriptively than an arbitrary svg?

> Instead, I propose making clear what happens if a logo is not specified.

I am sorry I don't follow 😅 Can you please elaborate more how you'd want this to be like? 😄

src/Splash/Splash.tsx
peterszerzo: We should remove it I'd say - otherwise we give the user the power to create a bad-looking screen :)

adeelibr: Sure, I'll update my PR. As soon as I get home. Office at the moment. 😅 Give me a couple of hours.

adeelibr:
@peterszerzo sure! i'll update my PR accordingly. :)

src/Splash/Splash.tsx
adeelibr: @peterszerzo removed the `logoSize` prop, and reverted it back to `110`

adeelibr:
@peterszerzo I have tweaked the keyframes, they now perfectly align with the background animation, can you kindly review.

adeelibr:
Todo:
- Test a <Splash /> component `logo` prop with an `img`, 
Blockers
- Current parcel configurations don't allow me to import an image in the `dev-server` folder. (correct me if I am wrong).

src/Splash/Splash.Logo.tsx
```diff
@@ -1,30 +1,88 @@
...
+    }
+    100% {
+      transform: rotate(0deg);
+    }
+  }
+`
+
+const defaultStyle = {
+  display: "inline-block",
+  animation: "rotation 11s 0s infinite linear",
```
TejasQ: The animation delay does not need to be `0` explicitly in this case.

src/Splash/Splash.Logo.tsx
TejasQ: Absolutely.

Currently, the code is like this:

```tsx
const MyComponent = () => {
  if (someCondition) {
    return <MyLogo />;
  }
  return <svg>BLAH BLAH WHAT AM I I HAVE NO IDEA! BLAH BLAH WHAT AM I I HAVE NO IDEA! BLAH BLAH WHAT AM I I HAVE NO IDEA! BLAH BLAH WHAT AM I I HAVE NO IDEA! BLAH BLAH WHAT AM I I HAVE NO IDEA! BLAH BLAH WHAT AM I I HAVE NO IDEA! BLAH BLAH WHAT AM I I HAVE NO IDEA! BLAH BLAH WHAT AM I I HAVE NO IDEA! BLAH BLAH WHAT AM I I HAVE NO IDEA! BLAH BLAH WHAT AM I I HAVE NO IDEA! BLAH BLAH WHAT AM I I HAVE NO IDEA! BLAH BLAH WHAT AM I I HAVE NO IDEA! BLAH BLAH WHAT AM I I HAVE NO IDEA! BLAH BLAH WHAT AM I I HAVE NO IDEA! BLAH BLAH WHAT AM I I HAVE NO IDEA! BLAH BLAH WHAT AM I I HAVE NO IDEA! BLAH BLAH WHAT AM I I HAVE NO IDEA! BLAH BLAH WHAT AM I I HAVE NO IDEA! BLAH BLAH WHAT AM I I HAVE NO IDEA!</svg>;
}
```

The problem I see here is that the SVG in the late return is quite ambiguous. What is it? It's a little hard to tell. Do you understand? Does that make sense? One possible solution could be to just extract the SVG into a component with a more descriptive name: `<ThisFallbackLogo />` that abstracts on top of the SVG essentially.

Did that clear things up?

TejasQ:
You can test the `logo` prop with an image URL,

> Blockers
> - Current parcel configurations don't allow me to import an image in the `dev-server` folder. (correct me if I am wrong).

I don't understand this. I'm not sure you would need to import an image. Why?

src/Splash/Splash.Logo.tsx
adeelibr: It's just a practice I like to focus, give explicit delay time, I know it is by default 0.

src/Splash/Splash.Logo.tsx
adeelibr: Yes this makes sense, can I instead of making separate component `<ThisFallbackLogo />` make a method inside the `render()` call in the same component like;

```
render() {
   const renderFallBackLogo = (
      <svg>
          <path />
          <path />
      </svg>    
   );
   
   if (someCondition) {
     return <MyLogo />;
   }
   return renderFallBackLogo;
}
```

Your thoughts? @TejasQ

adeelibr:
@TejasQ 

> I don't understand this. I'm not sure you would need to import an image. Why?

So currently the `logo` prop takes in an HTML element, It does not handle strings, which I think it should. So i'll make some changes to the internal implementation of `Splash.logo.tsx` component

src/Splash/Splash.Logo.tsx
TejasQ: Why? I’m interested to know the benefits of this practice.

src/Splash/Splash.Logo.tsx
adeelibr: Sounds cool! Implemented,

src/Splash/Splash.Logo.tsx
adeelibr: So I couldn't find a reference to back up my argument 😅 In my defense

- I like writing explicit code,
- since I write code for react native, react, node etc along with html & css, it's hard to sometimes remember the syntax for everything. (i know this might not be a good excuse)
- the syntax for me is readable `rotation 11s 0s infinite linear` this says run a rotation animation for 11s without any delay infinity with a linear  speed from beginning to end
- I know that no delay is applied by default, but this is just my opinion that the above statement makes it more readable

I can also understand your point of view (i think).
- there is a redundant addition value there which is not required.
- since it's more code, this is a size factor (even just a very minute one)

Maybe you have other concerns as well. This is what I assumed.

If you want I can remove the `0s` delay value from above, It just that this `rotation 11s 0s infinite linear` is more readable to me. (just my personal opinion)

adeelibr:
> how about it takes either a string (an image URL) or a React Node?

Done implemented :D Now this `logo` prop accepts HTML element or a string, and works both ways.

> That would remove the need to import images while still providing powerful image inclusion.

 Yes, I don't feel that we need the import images feature any more.

src/Splash/Splash.Logo.tsx
TejasQ: I understand. We can leave it in. I just was wondering what other developers collaborating on the project would think if they saw this.

TejasQ:
Looks good to me. @peterszerzo can I get an extra set of eyes on this?

src/Splash/Splash.Logo.tsx
adeelibr: > I just was wondering what other developers collaborating on the project would think if they saw this.

I hope they think good on this approach. 😅

peterszerzo:
@adeelibr can the animation get an ease-in-out? The current movement feels a bit rigid. We can then rebase and merge - looks fantastic otherwise.

adeelibr:
@peterszerzo Sure, give me a couple of hours to update my PR. I'll have this done by today.

peterszerzo:
@adeelibr I tried rebasing myself but it still says it's out-of-date, would you mind giving it a rebase yourself?

adeelibr:
@peterszerzo done!

adeelibr:
@TejasQ happy to contribute. If in the future there is an issue you'd like me to be part of. Just let me know. I'll be more then happy to dig into it.

TejasQ:
You are one of a kind, sir!

adeelibr:
This way I can learn more, by working with other devs who have different school of thought then mine + be more in OSS, which hopefully gives me more visibility & better chances of finding a good company out their someday 💃  @TejasQ All I can do is work hard at my goal to learn more.

PR #761, opened 2018-09-28
**Feature:** Add List and Stepper components

TejasQ:
# Summary
This PR adds a Stepper component _and_ a List component that will allow Labs to have an easier import flow.

# To be tested

Me

Tester 1

Tester 2


fabien0102:
image

The number is not well aligned

fabien0102:
image

Maybe something to prevent this case with large image

fabien0102:
image

The list example is broken

src/Stepper/Stepper.tsx
```diff
@@ -0,0 +1,123 @@
...
+  public state: StepperState = {
+    activeSlideIndex: 0,
+  }
+
+  public static defaultProps = {
+    stepColor: "primary",
+    activeSlideIndex: 0,
+  }
+
+  public static getDerivedStateFromProps({ activeSlideIndex }: Partial<StepperProps>) {
```
peterszerzo: Is the intention behind this setting is to have the active state be handled both in a controlled way and through local state? I just tried an uncontrolled stepper on the netlify link (just `<Steps steps={[/** a few steps in here */]} />`, and it wasn't changing the active step. If that's expected behavior, local state should be removed entirely.

src/Stepper/Stepper.tsx
```diff
@@ -0,0 +1,123 @@
...
+    font-weight: ${({ theme, isActive }) => (isActive ? "bold" : theme.font.weight.regular)};
+  flex: 1 0 auto;
+  font-size: ${({ theme }) => theme.font.size.body}px;
+  color: ${({ theme, isActive }) => (isActive ? theme.color.text.dark : theme.color.text.lighter)};
+  cursor: pointer;
+
+  /* Number with circles */
+  ::before {
+    content: "${({ number }) => number}";
+    width: 30px;
```
peterszerzo: I'm noticing this is a bit bigger than in the design, where I measured 24px. The font size should also be the same as the title text right next to the blue circle as in `theme.font.size.body`, whereas here it is `theme.font.size.title`.

src/List/List.tsx
```diff
@@ -0,0 +1,61 @@
...
+  flex: 0 0 170px;
+
+  img {
+    max-width: 100%;
+  }
+`
+
+const StyledIcon = styled(Icon)`
+  fill: ${({ theme }) => theme.color.border.default};
+  width: 30px;
```
peterszerzo: These look a bit too big - I think 24px would be better.

fabien0102:
image
Still broken on windows, but fixable by removing this strange `padding-top` and adding a `line-height` 😉

peterszerzo:
Why do we need paddings set if we're centering things with flexbox?

src/Stepper/Stepper.tsx
```diff
@@ -0,0 +1,123 @@
...
+    height: 30px;
+    display: flex;
+    flex: 0 0 30px;
+    align-items: flex-start;
+    justify-content: center;
+    font-weight: ${({ theme }) => theme.font.weight.bold};
+    font-size: ${({ theme }) => theme.font.size.title}px;
+    padding-top: 3px;
```
fabien0102: So to fix the alignment of the number into the circle, just remove this line and add `line-height: ${({ theme }) => theme.font.size.title}px`. And it should work everywhere.

(I'm curious about the origin of this `padding-top: 3px` btw 🤔, not really the best way to vertical-align a text)

src/Stepper/Stepper.tsx
TejasQ: The intention here is to allow someone to specify initial (or derived) state by specifying them in the prop passed to this component: as in, the intention behind this setting is to have the active state be only fully controlled.

There is no concept or support for an uncontrolled stepper.

TejasQ:
> (I'm curious about the origin of this padding-top: 3px btw 🤔, not really the best way to vertical-align a text)

> I also don't know why we are in flex-start here

@fabien0102 great to hear that the solutions you've proposed work on your machine. The reason behind the `padding-top` + `flex-start` hack is because it appeared to be the most pixel-perfect approach on mine. 🙂 

Consider,

> So to fix the alignment of the number into the circle, just remove this line and add line-height: ${({ theme }) => theme.font.size.title}px. 

Okay. This is what a direct copy and paste of the code in this comment yields.

image

> And it should work everywhere.

Unfortunately, in this case, it is not as simple as we think. 

> align-items: center; also works (instead of line-height), 

I'm actually not sure this can be considered "working". Notice how it's a couple of pixels too low on macOS:

image

with guides,

image

I'm not sure what the best solution is in this case: `align-items: center` is the cleanest, with barely noticeable misalignment on macOS, and apparently perfect alignment on windows. Another strong option is a hardcoded, concrete line-height instead of a theme value. 

Here's what it looks like with `line-height: 29px;` (or `theme.font.size.title - 1`).

image

I think the line height solution might be the best option actually. 

What does the team prefer? @peterszerzo @fabien0102

TejasQ:
[Stepper component](https://trello.com/c/OsMTjkSO/465-stepper-component)

peterszerzo:
The chevron icons in the list component are still too big - measuring 15 x 25 on the current branch, compared to 10 x 18 in the designs.

screen shot 2018-10-02 at 9 45 16 am

TejasQ:
@peterszerzo you are _the man_. Will fix.

Issue #751, opened 2018-09-21
Autocomplete component


fabien0102:
### Summary
![image](https://user-images.githubusercontent.com/1761469/45871463-dcf6cb80-bd8d-11e8-918b-60406c08fd4a.png)

We need a component to have autocompletion from a server response.

### API proposal
```typescript
export interface AutocompleteProps<T> {
  /**
   * Label text, rendering the input inside a tag if specified.
   * The `labelId` props is responsible for specifying for and id attributes.
   */
  label?: string
  /**
   * Should the input fill its container?
   */
  fullWidth?: boolean
  /**
   * A search can show a loading indicator.
   */
  loading?: boolean
  /**
   * Message to display when there are no results.
   */
  noResultsMessage?: React.ReactNode
  /**
   * Called when a result is selected.
   */
  onResultSelect: (result: { label: string; value: T }) => void
  /**
   * Called on search input change.
   */
  onSearchChange: (search: string) => void
  /**
   * Search results
   */
  results?: Array<{ label: string; value: T }>
  /**
   * Renders the SearchResult contents.
   * We render the `result.label` by default.
   */
  resultRenderer?: (result: { label: string; value: T }) => React.ReactNode
  /**
   * Minimum characters to query for results.
   */
  minCharacters?: number
}
```

### Usage
```jsx
<Get path={`?search=${this.state.search}`}>
  {(data, { loading }) => (
    <Autocomplete
      value={this.state.value}
      loading={loading}
      results={data}
      noResultsMessage="No result Found"
      onResultSelect={value => this.setState({ value })}
      onSearchChange={search => this.setState({ search })}
    />
  )}
</Get>
```

TejasQ:
This looks great. Great to see that you’ve thought of:

a) a lot of edge cases,
b) being verbose/explicit in the prop naming. 

I like this API and have no issues with it.

TejasQ:
Added in #762

PR #762, opened 2018-10-01
**Feature:** Add Autocomplete component

TejasQ:
# Summary
This PR adds support for an `Autocomplete` component that works in tandem with https://github.com/contiamo/restful-react.

# Related issue
#751 

# Open questions
- @fabien0102 do we need debounce on this? I saw [the implementation on your branch](https://github.com/contiamo/operational-ui/blob/feat-autocomplete/src/Autocomplete/Autocomplete.tsx#L75-L80) (this is a fork of your branch). To my understanding, Restful React's `Get` already has a debounce so I'm wondering if this is necessary here?

# To be tested

Me

Tester 1

- [ ] No regression on `Input`
- [ ] No regression on `Progress`
  

Tester 2

- [ ] No regression on `Input`
- [ ] No regression on `Progress`

### DIFF
--- a/src/Autocomplete/Autocomplete.tsx
+++ b/src/Autocomplete/Autocomplete.tsx
@@ -12,6 +12,10 @@ export interface AutocompleteProps<TValue> {
    */
   label?: string
   /**
+   * An Icon to append to each result
+   */
+  resultIcon?: Item["icon"]
+  /**
    * Should the input fill its container?
    */
   fullWidth?: boolean
@@ -57,7 +61,7 @@ const initialState = {
 type AutocompleteState = Readonly<typeof initialState>
 
 const Container = styled(ContextMenu)<Partial<AutocompleteProps<any>>>`
-  width: ${({ fullWidth }) => (fullWidth ? "100%" : "initial")};
+  width: ${({ fullWidth }) => (fullWidth ? "100%" : "fit-content")};
   display: flex;
   align-items: center;
 `
@@ -66,6 +70,7 @@ function makeItems<TValue>({
   searchValue,
   minCharacters,
   results,
+  resultIcon,
   noResultsMessage,
 }: Partial<AutocompleteProps<TValue> & AutocompleteState>) {
   if (!searchValue || !(searchValue.length > (minCharacters || 0))) {
@@ -73,7 +78,9 @@ function makeItems<TValue>({
   }
 
   if (results && results.length) {
-    return results
+    return results.map(
+      result => (typeof result === "string" ? { label: result, icon: resultIcon } : { ...result, icon: resultIcon }),
+    )
   }
 
   if (!results || !results.length) {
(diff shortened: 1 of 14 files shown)


fabien0102:
@TejasQ For the debounce, the idea was to debounce early to avoid all the re-rendering part where we can. I know that `Get` have also a debounce props, but since this kind of component always need a debounce, I was wondering to put this by default in this component.

After we need to test if the global reactivity of the component feel good or not.

src/Autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,133 @@
...
+`
+
+function makeItems<TValue>({
+  searchValue,
+  minCharacters,
+  results,
+  resultIcon,
+  noResultsMessage,
+}: Partial<AutocompleteProps<TValue> & AutocompleteState>) {
+  if (!searchValue || !(searchValue.length > (minCharacters || 0))) {
```
fabien0102: I think you can remove the `|| 0`, `minCharacters` already have a `defaultProps`

TejasQ:
Great catch! The interesting thing is that this component really doesn't do any data fetching but instead just rerenders `Get` when the state (search string) changes. 

It rerenders on text input, because it's controlled, so `Get` always rerenders and it fetches data.

Do you have an idea here how to implement `minCharacters`? Should that be the responsibility of this controlled component?

fabien0102:
This is why I had a `onSearchChange` (https://github.com/contiamo/operational-ui/blob/feat-autocomplete/src/Autocomplete/Autocomplete.tsx#L75-L80) after I'm agree that it's a very tricky component.

I think that if we decoupled a bit this `value` and the `onSearchChange`, we should be able to construct a nice component, not to hard to control and quite flexible

TejasQ:
I saw that, but [the `onSearchChange` looked a lot like `onChange`](https://github.com/contiamo/operational-ui/blame/feat-autocomplete/src/Autocomplete/README.md#L76). Let's try that.

fabien0102:
> I saw that, but the onSearchChange looked a lot like onChange. Let's try that.

Yes but because many things can change on this complexe component, I prefer to have `onSearchChange` instead of `onChange` to be more explicit about the behaviour.

This component is my initial inspiration => https://react.semantic-ui.com/modules/search/ you can have a look the API, this one is really well construct 💯

TejasQ:
I think it might be better to implement this without `Get` in mind. We can use the `debounce` prop on `Get` where we use them together, but ideally, I'd like this component to exist and work correctly irrespective of the data fetching layer of an application.

Do you think that makes sense?

TejasQ:
[User-selection (auto-complete) component](https://trello.com/c/bAHI64rU/453-user-selection-auto-complete-component)

TejasQ:
@fabien0102 what do you think of this? @peterszerzo I'd love your thoughts too.

src/Autocomplete/Autocomplete.tsx
```diff
@@ -0,0 +1,101 @@
...
+   */
+  loading?: boolean
+  /**
+   * Message to display when there are no results.
+   */
+  noResultsMessage?: string | Item
+  /**
+   * Called when a result is selected.
+   */
+  onResultClick: (item?: string | Item) => void
```
peterszerzo: Why is the argument optional in this function?

TejasQ: >  if you didn't want to use it, you'd just ignore the argument

That's exactly why the argument is optional. If you didn't want to use it and ignored the argument, TypeScript would yell at you and not compile, even though you have no use for the argument.

peterszerzo: Oh I see - that makes sense!

src/ContextMenu/ContextMenu.tsx
```diff
@@ -54,63 +60,110 @@ const MenuContainer = styled("div")<{
...
-                key={`contextmenu-${index}`}
-                condensed={condensed}
-                align={align}
-                width={width || "100%"}
-                item={item}
-              />
-            )
-          })}
-        </MenuContainer>
+        {this.state.isOpen && (
```
TejasQ: We were putting this on the DOM even though it wasn't open...

fabien0102:
Just amazing! Good job @TejasQ 🎉

Issue #759, opened 2018-09-26
Button styles cannot be extended


TejasQ:
Currently, it is impossible to create a styled variant of `Button`, while it _is_ possible to do this with other components:

```tsx
const NotNormalButtonCase = styled(Button)`
  color: ${({ theme }) => theme.color.primary};
  background-color: white;
  border: 0;
`;
```

doesn't work.

@peterszerzo any info about this?

peterszerzo:
The reason this doesn't work is this piece of styling code: https://github.com/contiamo/operational-ui/blob/master/src/Button/Button.tsx#L90

Just to write it out, buttons apply color this way (accommodating an edge case where if the button is rendered as an `<a/>` tag, the browser's default styling wins by specificity):

```
"&, a:link&, a:visited&": {
  textDecoration: "none",
  color: loading ? "transparent" : expandColor(theme, textColor) || foregroundColor,
},
```

Which prevents `emotion` from doings its own internal logic of overriding styles, and unpredictably packs it in the wrong order. If the colors were inline:

```
textDecoration: "none",
color: loading ? "transparent" : expandColor(theme, textColor) || foregroundColor,
```

then color overrides work, as do any other style overrides that don't conflict with these two rules (those work already).

In light of all of that, the other solution would be to only apply the color in the `a:visited, a:link` selector if the button is actually a link, fixing color overrides for most buttons. The ones rendered as a link (with a `to` prop) can be customized as `styled(Button)`color: yellow !important;`. Not the most savory, but it works, and without the extra prop.

TejasQ:
Does the solution in the PR work?

peterszerzo:
It does work, but it's strange to not have `styled(Button)({ color: "red" })` work in just this one case.

TejasQ:
I definitely agree. Maybe we should reopen the issue?

PR #768, opened 2018-10-04
**Feature:** Allow configuration of Button's text color

TejasQ:
# Summary
It's a little hard to extend buttons currently by adding a text color. This PR fixes that.

# Related issue
#759 

# To be tested

Me

Tester 1

- [x] Messing with the `textColor` prop renders expected Button.
  

Tester 2

- [ ] Messing with the `textColor` prop renders expected Button.

### DIFF
--- a/src/Button/Button.tsx
+++ b/src/Button/Button.tsx
@@ -16,6 +16,8 @@ export interface ButtonProps extends DefaultProps {
   to?: string
   /** Button color theme (hex or named color from `theme.color`) */
   color?: string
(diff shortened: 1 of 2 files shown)


src/Button/Button.tsx
```diff
@@ -135,9 +138,13 @@ const Button: React.SFC<ButtonProps> = ({
           {...props}
           color_={color}
           loading={loading}
+          disabled={disabled}
```
fabien0102: 🤔 I don't get it, it was in props, it's spread, so…

PR #773, opened 2018-10-05
**Feature:** Autocomplete/Input Prefill & Clear

TejasQ:
# Summary
This PR adds support for `Input` and `Autocomplete` to support prefilled values, and allows them to be `clear`ed via a prop. 

# Related issue
image
https://trello.com/c/bAHI64rU

# To be tested

Me

Tester 1

- [ ] No regression on `Input`
  

Tester 2

- [ ] No regression on `Input`

### DIFF
--- a/src/Input/Input.tsx
+++ b/src/Input/Input.tsx
@@ -77,7 +77,6 @@ const InputFieldContainer = styled("div")<{
   align-items: center;
   justify-content: center;
   max-width: 100%;
-  min-width: ${inputWidth}px;
   ${({ fullWidth, withLabel, theme }) => `
     margin-right: ${withLabel ? 0 : theme.space.small}px;
     display: ${withLabel ? "flex" : "inline-flex"};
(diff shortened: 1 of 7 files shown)


src/Input/Input.tsx
```diff
@@ -230,25 +267,81 @@ class Input extends React.Component<InputProps, State> {
...
+            {...commonInputProps}
+            clear={clear}
+            preset={Boolean(preset)}
+            id={forAttributeId}
+            autoComplete={autoComplete}
+            withIconButton={withIconButton}
+          />
+          {this.props.clear &&
+            this.props.value && (
+              <ClearButton disabled={Boolean(disabled)} preset={Boolean(preset)} onClick={this.props.clear}>
```
fabien0102: 🤔 I don't try yet, but I'm quite curious of the reason of this huge implementation for the "ClearButton", I have the same component hardcode on another project and I have only this:

```jsx
const UnselectIcon = styled(Icon)`
  position: absolute;
  right: 12px;
`;


<UnselectIcon
   color="#909090"
   name="No"
   onClick={() => setConfirmState({ user: {} as UserResp, search: "" })}
/>
```

Of course it's not the same context and I probably not cover all the functionnalities that we need, but I'm curious

src/Input/Input.tsx
```diff
@@ -230,25 +267,81 @@ class Input extends React.Component<InputProps, State> {
...
+          <InputField
+            {...commonInputProps}
+            clear={clear}
+            preset={Boolean(preset)}
+            id={forAttributeId}
+            autoComplete={autoComplete}
+            withIconButton={withIconButton}
+          />
+          {this.props.clear &&
+            this.props.value && (
```
fabien0102: Even after 21 lines of `this.props` destructuration, you are using `this.props.value` and `this.prop.clear`😆

src/utils/constants.ts
```diff
@@ -211,6 +212,9 @@ export const expandColor = (
   if (!colorToBeExpanded) {
     return null
   }
+  if (colorToBeExpanded.includes(".")) {
+    return get(theme, colorToBeExpanded, "red")
```
fabien0102: 👍 Very good idea!

peterszerzo:
@TejasQ I think the clear button would be simpler if we just positioned it absolute, made it a bit smaller than the height of the input itself, and didn't give it any border. That way the original input field's border takes care of the boundary, avoiding the jumps we saw on @fabien0102 gif above.

micha-f:
It's important that you can only clear once you have selected (clicked) an item from the dropdown list.

src/Input/Input.tsx
TejasQ: You're right. It was a quite convoluted. The main reasoning behind it was that I wanted it to be its own button, a sibling to the `input`, like the button when there's copy paste available, which would solve a lot of possibly unforeseen problems like text overflow.

After looking at it with fresh eyes, I'm not sure the convolution is worth the alleged safety. I also think that screenshot you posted might not be entirely valid because users typically don't browse the web so zoomed in unless there's special accessibility needs involved and so, as a workaround, I think that could work as a last resort, which at this point is unnecessary – we have better resorts, like `position: absolute`.

I've updated the implementation to be more straightforward. That will be a few more rules than your snippet though. I've added [a breakdown of why they're necessary in the PR](https://github.com/contiamo/operational-ui/blob/5700583d1828ce80cc83d62a84ab174b81abfa2b/src/Input/Input.tsx#L145-L162).

TejasQ:
@fabien0102 Thanks for the great feedback! I have a question: are you sure [about this](https://github.com/contiamo/operational-ui/pull/773#issuecomment-427347367)?

image

@micha-f is [this behavior (in the second example)](https://deploy-preview-773--operational-ui.netlify.com/#!/Autocomplete) somehow different than [your comment](https://github.com/contiamo/operational-ui/pull/773#issuecomment-427352685)?

micha-f:
@TejasQ That looks good!

src/Input/Input.tsx
fabien0102: I've zoomed because it was looking bad (and you always ask to be pixel perfect), but it seams to be only on windows, it was looking good on the @peterszerzo screen

fabien0102:
image

For the disable diff on windows ;)

TejasQ:
About [this](https://github.com/contiamo/operational-ui/pull/773#issuecomment-427347367) -> [this](https://github.com/contiamo/operational-ui/pull/773#issuecomment-427371654) -> and [this](https://github.com/contiamo/operational-ui/pull/773#issuecomment-427377458), you've got to be kidding me: this is some specific user-agent-specific stylesheet stuff that Chrome adds differently on Windows than it does on Mac. I don't know why.

Here's my `user agent stylesheet`. It doesn't have that selector at all that exists in your screenshot:
image

I wonder what the reasoning is behind this. Maybe we can ask the Chrome team? (cc @kylix-rd @madmath)

Either way, setting `background-color` explicity will (did) fix this in this PR.

Issue #727, opened 2018-09-06
`LabelText` is inline, but really should be a `block` 🤔


TejasQ:
I'm really not sure why we use a `span` for `LabelText`. It is inline, but all of our designs and implementations have labels on separate lines above other elements. Consider,

# Status Quo
![image](https://user-images.githubusercontent.com/9947422/45148204-0547cd00-b1c7-11e8-8fe8-e1d18211a460.png)

# Expected
![image](https://user-images.githubusercontent.com/9947422/45148331-52c43a00-b1c7-11e8-8b38-ceae5b53a5b4.png)

This issue humbly requests fixing this by making `LabelText` a block-level element.

adeelibr:
Can I work on this?

peterszerzo:
I think it makes sense to take care of both this issue and #726 by both converting `LabelText` from https://github.com/contiamo/operational-ui/blob/master/src/utils/mixins.ts#L37 to a block-style element, as well as moving it into its own folder `src/LabelText` and making it a public export.

adeelibr:
Can there be 2 PR for this? One for the label to block conversion for `LableText` and the other where we move it into it's seperate folder `src/LabelText` or should both of them be done in the same PR?

peterszerzo:
@adeelibr that's completely up to you. It is a teeny bit safer to split up into two PR's, but for a change this small, I'm not sure it's worth the bureaucracy :)

(also, thanks for giving us a hand on this!)

adeelibr:
@peterszerzo I don't mind doing it in 2 PR's 😅 Once this PR is done. I'd be more then happy to create a new pull request for `LabelText` to move into it's own folder.

adeelibr:
@peterszerzo The first PR was merged, can I now make another PR to move it into a separate module. If no one is working on it?

adeelibr:
The first PR was done https://github.com/contiamo/operational-ui/pull/753  where I converted the element to `block` level. I am still waiting for go ahead to make the next PR to move this to a separate file. @TejasQ

TejasQ:
Ah, you're waiting on @peterszerzo. Cool!

adeelibr:
yup :sw

peterszerzo:
@adeelibr you're welcome to go ahead with the other branch.

PR #777, opened 2018-10-06
**Feature:** LabelText Moved To src/LabelText

adeelibr:
# Summary

Moved `LabelText` from `src/utils/mixins.ts` to `src/LabelText`

# Related issue

https://github.com/contiamo/operational-ui/issues/727#issuecomment-427306988


adeelibr:
@peterszerzo Can you kindly review?

Also should there be `README.md` file in the `src/LabelText` folder? If yes, then what should be the initial content like?

peterszerzo:
@adeelibr code looks great. Yes, we should have a `README.md` explaining why we have this (input fields use these internally, but sometime there is a need to use them standalone with other UI elements to guarantee consistency with surrounding input fields). Just add one usage snippet.

adeelibr:
Also while we are at it, Shouldn't `LabelText` be of `<label/>` element instead of `<span/>`?

src/LabelText/README.md
```diff
@@ -0,0 +1,12 @@
+This component renders label text, The `<LableText />` element is used to associate with a form field in `<input />` field. The `LabelText` tells the users the value that should be enterd in the associated input field. This is typically used within the following
```
peterszerzo: Some suggestions for the wording here:
- `renders label text -> renders a semantic label tag` - if we do this, we don't need to talk about how they associate with input fields, because folks knowing HTML already know that.
- `typically used -> used internally in the following components if the label prop is specified:`
- we should explain why this component is exposed: though it is used internally, sometimes you want to use it standalone with other components to achieve visual consistency with surrounding labeled input fields.

src/LabelText/LabelText.tsx
```diff
@@ -0,0 +1,13 @@
+import styled from "react-emotion"
+
+export const labelTextHeight = 15
+
+export const LabelText = styled("span")(({ theme }) => ({
```
adeelibr: @peterszerzo shouldn't this be a `label` tag instead of `span`  for better usability with form elements.

src/LabelText/README.md
adeelibr: I have updated the README file, is this better now? 😄

src/LabelText/LabelText.tsx
TejasQ: Good idea, Adeel! Let's try that!

adeelibr: Done.

peterszerzo:
README looks great, @adeelibr - from my end, we can merge after a rebase

adeelibr:
@peterszerzo it is in sync with `master` branch now.

TejasQ:
Thanks @adeelibr! We'll merge when checks pass.

PR #780, opened 2018-10-09
**Feature:** Recalculate tree open paths when depth changes

### LINKED ISSUE #765: Tree item's `initiallyOpen` prop doesn't work when tree is updated
When the tree changes, e.g. after it's been initially empty and some data arrived from the server, nodes that are marked `initiallyOpen` don't actually open. This is because open states are calculated once *on mount* and are controlled by the user from that point onwards. When there's an update, nothing happens based on these props.

Expected:
screen shot 2018-10-02 at 10 02 02 am

Actual:
screen shot 2018-10-02 at 10 02 07 am

Recalculating on update is a bit tricky, though: whenever the component updates, it contains event listeners that won't be referentially equal to the previous values, so it's hard to check for tree changes that would have to retrigger open states, undoing all manual open and close operations the user has done so far.

I can envision some kind of checksum-based solution that recalculates the open states when the shape of the tree changes:
- if it was previously empty and now it isn't, recalculate. This is like duh, but it also solves most production cases.
- if it changes a little bit (one item less in the children of the third child), open states shouldn't recalculate, because the user would expect items already closed to stay closed.
- if it changes a lot (three top-level items disappear and the max depth goes from 6 to 2), it should probably change.
- in-between cases are tricky to distinguish, and any effort figuring out a fancy logic might still run into edge cases.

Any thoughts on how we should proceed?

peterszerzo:
# Summary

This PR fixes #765 (refer to issue for a detailed description). The chosen solution recalculates initially open paths when the maximum depth of the tree changes.

# Test like so

Replace the tree snippet with the following:

```js
const createTrees = () => [
    {
      label: "Store",
      childNodes: [
        {
          label: "Region",
          initiallyOpen: true,
          childNodes: [
            {
              label: "City",
              tag: "D",
              disabled: true,
              childNodes: [],
            },
            {
              label: "County",
              color: "primary",
              tag: "D",
              childNodes: [],
            },
          ],
        },
      ],
    },
    {
      label: "Legal Entity",
      initiallyOpen: true,
      childNodes: [
        {
          label: "Limited Liability Company",
          tag: "D",
          childNodes: [],
          draggable: true,
          onDragStart: ev => {
            ev.dataTransfer.effectAllowed = "move"
          },
        },
        {
          label: "Inc.",
          tag: "D",
          color: "#2C363F",
          childNodes: [],
        },
      ],
    },
  ]

initialState = { 
  trees: []
}

setTimeout(() => {
  setState(() => ({
    trees: createTrees()
  }))
}, 3000)

;<Tree
  trees={state.trees}
/>
```

This snippet updates the tree from empty to non-empty (the most common use-case when this bug occurs). After the 3s timeout, the new tree appears in the expected open state based on the `initiallyOpen` flags in the individual nodes.

# Related issue

#765

# To be tested

Me
- [x] No regressions on tree

Tester 1

Tester 2

### DIFF
--- a/src/Tree/Tree.utils.ts
+++ b/src/Tree/Tree.utils.ts
@@ -1,6 +1,29 @@
 import { Tree } from "./Tree.types"
 
-const pathsEqual = (path1: number[], path2: number[]) => path1.join("-") === path2.join("-")
+const arePathsEqual = (path1: number[], path2: number[]) => path1.join("-") === path2.join("-")
+
+/**
+ * Returns the largest element of an array of numbers, or undefined if the array is empty.
+ */
+const getMaxFromList = (nos: number[]): number | undefined => {
+  if (nos.length === 0) {
+    return undefined
+  }
+  return nos.reduce((accumulator, current) => (accumulator < current ? current : accumulator), -100000)
+}
+
+export const getDepth = (tree: Tree): number => {
+  if (tree.childNodes.length === 0) {
+    return 1
+  }
+  // Type-casting is necessary because at this point in the function there is a guarantee
+  // that tree.childNodes is not empty and therefore it has a maximum value.
+  return 1 + (getMaxFromList(tree.childNodes.map(getDepth)) as number)
+}
+
+export const getMaxDepth = (trees: Tree[]): number | undefined => {
+  return getMaxFromList(trees.map(getDepth))
+}
 
 export const getInitialOpenPaths = (basePath: number[]) => (tree: Tree): number[][] => {
   return [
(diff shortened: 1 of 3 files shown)


src/Tree/Tree.utils.ts
```diff
@@ -2,6 +2,29 @@ import { Tree } from "./Tree.types"
 
 const pathsEqual = (path1: number[], path2: number[]) => path1.join("-") === path2.join("-")
 
+/**
+ * Returns the largest element of an array of numbers, or undefined if the array is empty.
+ */
+const maxFromList = (nos: number[]): number | undefined => {
```
TejasQ: I'd suggest a _small_ change to the naming of these functions in this file:

```diff
- maxFromList
+ getMaxFromList
```

in order to make them a little more declarative. The following functions `getDepth` and `getMaxDepth` already follow this convention, though I'd propose the same change to `pathsEqual` -> `arePathsEqual`. Man, I can already see that showing up in a ternary and looking FREAKING AMAZING:

`return arePathsEqual ? doSomething() : doSomethingElse()` 

It reads like English. 😍

peterszerzo: Makes sense - thanks for the catch :)

PR #781, opened 2018-10-10
**Breaking:** opinionated table actions

### LINKED ISSUE #766: Give a proper API to `Table.__experimentalRowActions`
In order to produce this kind of table:
image

we need to use `__experimentalRowActions`, since it's a quite commun case, I think it's time to give a proper API to this feature.

peterszerzo:
# Summary

Adds an opinionated, `ActionMenu `-based table row actions solution, replacing `__experimentalRowActions` with `rowActions`. The API is the same as the `items` prop of `ActionMenu` so it should be familiar both on an API level and visually.

screen shot 2018-10-10 at 10 04 31 am

# Related issue

Fixes #766.

# To be tested

Me
- [x] Table actions work
- [x] Click propagation is stopped so that default row click events are not triggered 

Tester 1

- [x] No regressions on tables.
  

Tester 2

- [ ] No regressions on tables.

### DIFF
--- a/src/Table/README.md
+++ b/src/Table/README.md
@@ -1,12 +1,23 @@
 Tables simply render a semantic HTML table structure based on raw data.
 
-### Usage
+### Simple usage
+
+You can render a simple use-case of the table by specifying a list of records and supplying the columns as a list of strings which would serve both as column heading and access keys for fields.
 
 ```js
-<Table columns={["Name", "Title"]} rows={[["Max", "Carpenter"], ["Moritz", "Baker"]]} />
+<Table
+  data={[{ name: "Max", profession: "Carpenter" }, { name: "Moritz", profession: "Baker" }]}
+  columns={["name", "profession"]}
+/>
 ```
 
-### With external data to format
+While this approach is convenient, it is not recommended because it makes it too easy to re-use record keys as table headings, and adds a strong coupling between data and view concerns.
+
+We suggest taking the time to think about the best way to describe the data fields and then specifying it explicitly as column headers. The following, slightly more verbose version of the API demonstrates how this can be achieved:
+
+### With explicit data formatting
+
+In this case, simple functions taking an individual record as an argument specify how cells should be rendered in a given column:
 
 ```jsx
 const data = [
@@ -45,21 +56,90 @@ const data = [
   },
 ]
 ;<Table
-  columns={["", "Name", "Last updated", "Tags", "Collaborators"]}
-  rows={data.map(item => [
-    <Icon name="Box" color="info" />,
-    item.name,
-    item.lastUpdated,
-    item.tags.map((tag, i) => <Chip key={i}>{tag}</Chip>),
-    <AvatarGroup avatars={item.collaborators} />,
-  ])}
-  onRowClick={(row, i) => console.log({ row, i })}
-  rowActionName="view details"
+  data={data}
+  columns={[
+    { heading: "", cell: dataEntry => <Icon name="Box" color="info" /> },
+    { heading: "Name", cell: dataEntry => dataEntry.name },
+    { heading: "Last updated", cell: dataEntry => dataEntry.lastUpdated },
+    { heading: "Tags", cell: dataEntry => dataEntry.tags.map((tag, tagIndex) => <Chip key={tagIndex}>{tag}</Chip>) },
+    { heading: "Collaborators", cell: dataEntry => <AvatarGroup avatars={dataEntry.collaborators} /> },
+  ]}
+  onRowClick={(dataEntry, i) => console.log({ dataEntry, i })}
+/>
+```
+
+### With icons
+
+```jsx
+const data = [
+  {
+    name: "Mega Deal Dev",
+    lastUpdated: "2018-06-06",
+    tags: ["agent-view", "production"],
+  },
+  {
+    name: "Mega Deal Dev",
+    lastUpdated: "2018-06-06",
+    tags: ["agent-view", "production"],
+  },
+  {
+    name: "Toy Bundle",
+    lastUpdated: "2018-06-06",
+    tags: ["agent-view", "production"],
+  },
+]
+;<Table
+  data={data}
+  icon={dataEntry => "Building"}
+  columns={[
+    { heading: "", cell: dataEntry => <Icon name="Box" color="info" /> },
+    { heading: "Name", cell: dataEntry => dataEntry.name },
+    { heading: "Last updated", cell: dataEntry => dataEntry.lastUpdated },
+    { heading: "Tags", cell: dataEntry => dataEntry.tags.map((tag, tagIndex) => <Chip key={tagIndex}>{tag}</Chip>) },
+  ]}
+  onRowClick={(dataEntry, i) => console.log({ dataEntry, i })}
+/>
+```
+
+### With default row action
+
+```jsx
+const data = [
+  {
+    name: "Mega Deal Dev",
+    lastUpdated: "2018-06-06",
+    tags: ["agent-view", "production"],
+  },
+  {
+    name: "Mega Deal Dev",
+    lastUpdated: "2018-06-06",
+    tags: ["agent-view", "production"],
+  },
+  {
+    name: "Toy Bundle",
+    lastUpdated: "2018-06-06",
+    tags: ["agent-view", "production"],
+  },
+]
+;<Table
+  data={data}
+  icon={dataEntry => "Building"}
+  iconColor={dataEntry => (dataEntry.name.indexOf("Mega") > -1 ? "primary" : "gray")}
+  columns={[
+    { heading: "", cell: dataEntry => <Icon name="Box" color="info" /> },
+    { heading: "Name", cell: dataEntry => dataEntry.name },
+    { heading: "Last updated", cell: dataEntry => dataEntry.lastUpdated },
+    { heading: "Tags", cell: dataEntry => dataEntry.tags.map((tag, tagIndex) => <Chip key={tagIndex}>{tag}</Chip>) },
+  ]}
+  onRowClick={(dataEntry, i) => console.log({ dataEntry, i })}
+  rowActionName="View Details"
 />
 ```
 
 ### With row actions
 
+Row actions are specified as a function of an individual record, returning action items conforming to the [ContextMenu](/#ContextMenu) API:
+
 ```jsx
 const data = [
   {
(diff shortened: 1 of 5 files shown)


src/Table/Table.tsx
```diff
@@ -83,29 +84,26 @@ const Actions = styled(Td)(({ theme }) => ({
...
+const Table: React.SFC<TableProps> = ({ rows, columns, onRowClick, rowActionName, rowActions, ...props }) => {
   return (
     <Container {...props}>
       <thead>
         <Tr>
           {columns.map((title, i) => (
             <Th key={i}>{title}</Th>
           ))}
-          {Boolean(__experimentalRowActions || rowActionName) && <Th />}
+          {Boolean(rowActions || rowActionName) && <Th />}
```
TejasQ: What is `rowActionName`?

TejasQ: Yeah, I agree. Can we remove it?

peterszerzo: On it :)

src/Table/README.md
```diff
@@ -106,12 +106,20 @@ const data = [
     <AvatarGroup avatars={item.collaborators} />,
   ])}
   onRowClick={(row, i) => console.log({ row, i })}
-  __experimentalRowActions={data.map(item => (
-    <>
-      <Button color="info">details</Button>
-      <Button color="error">delete</Button>
-    </>
-  ))}
+  rowActions={data.map(item => [
```
TejasQ: Actually, why `data.map` here?

peterszerzo: This is how the API is currently set up - `rowActions` is an array of actions React nodes, one per row.

TejasQ: That is not at all what I expected. 😕
Certainly not what I expected reading [this](https://github.com/contiamo/operational-ui/pull/781/files#diff-11b59e1c788d31182aad220b72a6f5beR16).

I expected an array of `IContextMenuItem`s, oddly enough. Consider,

```tsx
<Table
  rowActions={[
    { label: "sup", onClick: () => alert("boi") },
    { label: "show me my friend", onClick: () => window.location = "https://twitter.com/coleturner" }
  ]}
/>
```

I'm not sure why it needs to be linked to the `data` in the table...

TejasQ: As an aside, can we please link to Cole's twitter the `Table` README? :D

peterszerzo: You're right, it's not an array of react nodes, more like the props needed to construct them.

The API you've copied is similar to what is, you simply have one for each table row. This way you can have table actions change dynamically based on what the data responsible for the row is, e.g. redirect to the show page of a single record.

peterszerzo: The API expects you to compute the props responsible to get an ActionMenu going for each row, and stick those props (list of menu items) into an array. A function that takes a record and gives you a single list of items would be an API alternative - but if we go with that, we should change the way rows are calculated also (which is what the current API is consistent with).

TejasQ: > A function that takes a record and gives you a single list of items would be an API alternative

Would this also be an alternative?

```tsx
<Table
  rowActions={(allTheRows, thisIndex) => [
    { label: "sup", onClick: () => alert(allTheRows[thisIndex].sup) },
    { label: "show me my friend", onClick: () => window.location = "https://twitter.com/coleturner" }
  ]}
/>
```

peterszerzo: I don't think you should have access to all the data points inside the function, I feel like it would encourage muddling up rows with each other. I was thinking of going full-on data table like this:

```
<Table
  data={/** array of some data type */}
  columns={[
    {
      heading: "Name",
      cell: datum => datum.name
    },
    {
      heading: "Age",
      cell: datum => datum.age < 18 ? <span>Young</span> : <span>Old</span>
    }
  ]}
  actions={datum => [
    {
      label: "View",
      onClick: () => { redirect(datum.id) }
    }
  ]}
/>
```

So it's like the `DataTable` implementation we discussed previously, just with explicit row headers.

TejasQ: I like it, _except_ I think the API would be nicer if `columns` also accepted `string[]` in case we wanted the cells to be displayed exactly as they are.

peterszerzo: You mean something like this:
```js
<Table
  data={/** array of some data type */}
  columns={[
    "name",
    {
      heading: "Age",
      cell: datum => datum.age < 18 ? <span>Young</span> : <span>Old</span>
    }
  ]}
  actions={datum => [
    {
      label: "View",
      onClick: () => { redirect(datum.id) }
    }
  ]}
/>
```
Which would just take the string `"name"` as both column heading and accessor key right? I am very much concerned about how much this couples how backend names fields and how frontend displays them - the frontend should be explicit about how it displays data, and we should take extra care to name table headings sensibly and not take the lazy way out.

TejasQ: From your example, I would understand that the heading for the first column in the table is "name" and the value in each cell is `data.name`, yes. I quite like that in terms of syntax sugar.

src/Table/README.md
```diff
@@ -45,20 +54,21 @@ const data = [
...
+    { heading: "Collaborators", cell: datum => <AvatarGroup avatars={datum.collaborators} /> },
+  ]}
+  onRowClick={(datum, i) => console.log({ datum, i })}
 />
 ```
 
-### With \_\_experimentalRowActions
+### With row actions
+
+Row actions are specified as a function of an individual record, returning action items conforming to the `ContextMenu` API:
```
TejasQ: Can we link to the `ContextMenu` here? It should be as simple as

```suggestion
Row actions are specified as a function of an individual record, returning action items conforming to the [`ContextMenu`](/#ContextMenu) API:
```

src/Table/README.md
```diff
@@ -97,26 +107,36 @@ const data = [
...
-  onRowClick={(row, i) => console.log({ row, i })}
-  __experimentalRowActions={data.map(item => (
-    <>
-      <Button color="info">details</Button>
-      <Button color="error">delete</Button>
-    </>
-  ))}
+  data={data}
+  columns={[
+    { heading: "", cell: datum => <Icon name="Box" color="info" /> },
```
TejasQ: Maybe we can call `datum` `row` in order to not confuse our German-speaking newcomers and also _use_ the row somehow in the example.
```suggestion
    { heading: "", cell: row => <Icon name="Box" color={row.color} /> },
```

src/Table/Table.tsx
```diff
@@ -83,48 +75,60 @@ const Actions = styled(Td)(({ theme }) => ({
...
-  ...props
-}) => {
+function Table<T>({ data, columns, onRowClick, rowActions, ...props }: TableProps<T>) {
+  const standardizedColumns: Array<Column<T>> =
+    Boolean(columns[0]) && typeof columns[0] === "string"
+      ? (columns as Array<Extract<keyof T, string>>).map(columnName => ({
+          heading: columnName,
+          cell: (datum: T) => datum[columnName],
+        }))
+      : (columns as Array<Column<T>>)
```
TejasQ: Maybe a newline after this for readability? 👼 
```suggestion
      : (columns as Array<Column<T>>)
      
```

src/Table/Table.tsx
```diff
@@ -83,48 +75,60 @@ const Actions = styled(Td)(({ theme }) => ({
...
           ))}
-          {Boolean(__experimentalRowActions || rowActionName) && <Th />}
+          {Boolean(rowActions) && <Th />}
         </Tr>
       </thead>
       <tbody>
-        {rows.length ? (
-          rows.map((row, i) => (
+        {data.length ? (
+          data.map((datum, i) => (
```
TejasQ: Can we be more explicit about the intent of `i` in this API?

```suggestion
          data.map((row, index) => (
```

src/Table/README.md
```diff
@@ -1,12 +1,21 @@
...
 ```js
-<Table columns={["Name", "Title"]} rows={[["Max", "Carpenter"], ["Moritz", "Baker"]]} />
+<Table
+  data={[{ name: "Max", profession: "Carpenter" }, { name: "Moritz", profession: "Baker" }]}
+  columns={["name", "profession"]}
+/>
 ```
 
-### With external data to format
+While this approach is convenient, it is not recommended because it makes it too easy to re-use record keys as table headings, instead of taking the time to think about the best way to describe fields and supplying that explicitly. The following, slightly more verbose API makes this possible:
```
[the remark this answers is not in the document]
peterszerzo: I'll mash the two up - I would like to avoid subjective judgments on what is `proper`, but point out the specific harm that can happen instead.

src/Table/README.md
peterszerzo: It is not a row, though, and marking it as such breaks the link between variable names that I would argue currently makes the code easier to follow. If the latin singular is of concern (I know, it sounds awful pretentious), maybe we can go with `dataEntry` instead?

src/Table/README.md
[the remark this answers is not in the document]
TejasQ: Yes, definitely.

src/Table/README.md
TejasQ: I think `dataEntry` would be more approachable.

fabien0102:
@peterszerzo Nice job, and the API is better than before, but do we really want to break this API now?

This a lot of works to migrate in all our apps and we are quite short in time before the launch ^^

I propose three options:
1. keep this for later
2. don't break the API with some deprecated props for a smooth migration
3. A codemod to insure a nice and easy migration

peterszerzo:
@fabien0102 I don't think upgrade will take more than half an hour per project, and will make things safer and nicer to work with, plus it will remove custom table actions code across the board.

I realize this may still sound steep, and I should also note that this wasn't the original intention of this PR. Neither of us liked an API that wasn't steering the component in this direction already, and ending up with a hybrid (arrays for rows, functions for actions) is a clearer outcome.

TejasQ:
How about we release it under the `canary` tag and install it in a project and evaluate the migration work? We can do that before we merge this. 🙂  If it's a huge amount of effort, we'll need to weight it against the 9th November deadline. If not, we can merge it now. What do y'all think?

We've spent quite a bit of time making the API a lot nicer so it'd be nice to play with it, at least for me with the IdP.

AFAIK there's also a requirement for this in Pantheon, right @peterszerzo?

peterszerzo:
It is - it was the plan to get standardized table actions in there. Otherwise I'll have to implement that on the side.

peterszerzo:
Upgrading a production application with 8 reasonably complex tables took 20 minutes, adding table actions and eliminated risks of misaligned cells.

I should also note that for projects that used `DataTable`, the process is even simpler.

TejasQ:
That's good! Let's do [this](https://github.com/contiamo/operational-ui/pull/781#issuecomment-431013756) then. I need to use this in the IdP as well fairly soon so let's merge it at the very latest my _Tuesday next week_ after using canary, if there are no major problems on `canary`.

I will dogfood it in IdP and Labs. @fabien0102 if you'd dogfood `@canary` in Hub, that would be helpful too.

If it's a <30 minute migration, I'd say we go live with `@operational/components` v11 early next week.

TejasQ:
I was _this_ close to merging this but I have a really interesting use case where `rowActions` are a little hard to implement: 

- each item needs access to the datum
- each item needs a function passed down from restful-react, with the datum available in there.

This calls for an implementation similar to 

```tsx
rowActions={realm => (
                      <UpdateRealm tenantId={tenantId} id={realm.id}>
                        {updateRealm => (
                          <DeleteRealm tenantId={tenantId}>
                            {deleteRealm => (
```

But now I have JSX in there and I'm not sure how to map it to ContextMenu-style objects. Would it make sense for easier migration to allow a `ReactNode` in there as well? 🤔

peterszerzo:
@TejasQ what if you just passed a function to `rowActions` that output a react node that ultimately renders an `ActionMenu`? That should allow you to wrap it with whatever you please and still get a consistent node in there. The only modification we do on the `ActionMenu` element that is rendered internally is to override its display to inline-flex, which we could investigate (whether it is necessary), or just add with a child selector on the parent. Long story short, implementation-wise it seems very straightforward.

TejasQ:
LETS DO IT 🚀

peterszerzo:
Done already!

TejasQ:
Peter

src/Table/README.md
```diff
@@ -45,20 +56,89 @@ const data = [
...
+  },
+  {
+    name: "Toy Bundle",
+    lastUpdated: "2018-06-06",
+    tags: ["agent-view", "production"],
+  },
+]
+;<Table
+  data={data}
+  icon={dataEntry => "Building"}
```
TejasQ: How does one deal with the color of an Icon as in Labs?

image

src/Table/README.md
```diff
@@ -1,12 +1,23 @@
...
-### Usage
+### Simple usage
+
+You can render a simple use-case of the table by specifying a list of records and supplying the columns as a list of strings which would serve both as column heading and access keys for fields.
 
 ```js
-<Table columns={["Name", "Title"]} rows={[["Max", "Carpenter"], ["Moritz", "Baker"]]} />
+<Table
+  data={[{ name: "Max", profession: "Carpenter" }, { name: "Moritz", profession: "Baker" }]}
+  columns={["name", "profession"]}
```
fabien0102: The idea of `<DataTable />` was to provide `columnsFormatter` and `formatters` to be able to customize headers and values. If we remove these two props, I don't really see the advantage of this API (as well explain in the comment bellow).

The main idea behind this `columnsFormatter` was to use a translation system to be able to customize any header regarding the language of the user.

```tsx
<DataTable
    data={data}
    onRowClick={i => pushState(`usecases/${i.id}`)}
    columns={["id", "createdAt", "name"]}
    columnsFormatter={name =>
      intl.formatMessage({
        id: `usecases.headers.${name}`,
        defaultMessage: title(name)
      })
    }
  />
```

But it's true that this API is the most user friendly, so I propose to remove totally the `columns={["name", "profession"]}` syntax in favor of the other one propose :wink:

peterszerzo: The idea behind the change is to encourage explicitly specifying table headers, and removing shortcuts that blindly uppercase fields and give it to the user. This information should be directly visible and customizable, and we should spend a bit of extra work writing out descriptive and well-formatted table headers.

I am surprised that you don't see the advantage of this API - it is type-safe the same way data table is.

peterszerzo:
I would cook the pagination idea a bit more - you are always welcome to keep the abstraction in a dependent project and see how it fares. The scope of this PR is large enough as it is.

fabien0102:
@peterszerzo Sure, I will open another PR for the paginator, you're right, this PR is large enough 💯

Issue #786, opened 2018-10-11
Upgrade Progress's constants


TejasQ:
[`Progress` uses the legacy constants](https://github.com/contiamo/operational-ui/blob/master/src/Progress/Progress.tsx) from back when we used glamorous. `theme.deprecated.` needs to be replaced with `theme.` in order to modernize it and make it customizable via runtime configuration and general theming.

Issue #785, opened 2018-10-11
z-index issues :(


TejasQ:
So we have a bit of a z-index conflict going on: 
![image](https://user-images.githubusercontent.com/9947422/46798318-e0102680-cd51-11e8-9b5c-10b0d69fbc3c.png)

In order to fix it, either:

- `ContextMenu` or `HeaderMenu` needs a higher z-index, or
- `Progress` needs a lower one.

I think the first option makes the most sense because progress always needs to be visible and thus lowering its z-index might mask it in some cases.

Thoughts? @peterszerzo @ImogenF

peterszerzo:
@TejasQ the progress bar shouldn't be there in the first place from a design perspective - we should either replace it with a spinner for pages or find a way to put the progress bar all the way up like on github's interface. Portals may help, though they'll complicated setup, unless the `OperationalUI` renders the portal content straight off the bat. I realize this is murky territory, but maybe we can come up with something :)

TejasQ:
Waaaaait so the progress bar is only supposed to be on the very top?

peterszerzo:
You're right :)

TejasQ:
Awesome! I'll add that to #788!

Issue #782, opened 2018-10-10
Invalid CSS on `Table`


TejasQ:
[This is invalid CSS](https://github.com/contiamo/operational-ui/blob/master/src/Table/Table.tsx#L54) on Table. 

[This is the documentation of the declaration](https://developer.mozilla.org/en/docs/Web/CSS/vertical-align).

Ideally, we use correct CSS and allow our tables to properly center the contents of their cells.

Issue #778, opened 2018-10-08
Improve `HeaderMenu` and `ContextMenu` types


fabien0102:
### Problem

```tsx
<HeaderMenu
  items={[
    { label: "My Organization", value: "organization" },
    ...me.realms.map(realm => ({ label: realm.name, value: realm })),
  ]}
  onClick={selected => {
    /* selected is string | IContextMenu<any> | undefined */
  }}
  withCaret
>
  /* I can't put a string here => Type 'string' is not assignable to type 'ReactNode[]' */
</HeaderMenu>
```

Ideally, `selected` should be `IContextMenu<Realm | string>` and I should be able to put a simple `string` into `children` (the current workaround is to do `{["my string"]}` as `children`…)

TejasQ:
Can I please work on this?

TejasQ:
I don't understand the use case of `children` for this component though. The `value` is set in a prop and it really shows nothing inside it...

PR #788, opened 2018-10-11
**Breaking:** Adjustments

TejasQ:
# Summary
This PR composes a bunch of fixes to various open issues (see commits). It contains [_one_ breaking change](https://github.com/contiamo/operational-ui/pull/788/files#diff-85f3197de18a071f349f7f5a257a93e3R25) to the `onClick` prop of `ContextMenu`: in terms of API, it only takes one argument: a `ContextMenuItem` instead of a `string | ContextMenuItem`.

# Related issue
#746 
#778 
#782 
#785 
#786 
#757 

# To be tested

Me

Tester 1

Tester 2


src/Autocomplete/util/makeItems.ts
```diff
@@ -0,0 +1,22 @@
+import { IContextMenuItem } from "../../ContextMenu/ContextMenu.Item"
```
peterszerzo: Does this method need to be in its own folder in a separate file? Other components use the `ComponentName/ComponentName.utils.ts` pattern, where utility methods are piled up (see e.g. `DatePicker`).

src/ContextMenu/ContextMenu.tsx
```diff
@@ -37,6 +37,7 @@ export interface ContextMenuProps extends DefaultProps {
 export interface State {
   isOpen: boolean
   focusedItemIndex: number
+  items: IContextMenuItem[]
```
peterszerzo: Is there another reason this is in state other than a way to convert between `string`'s and `Item`'s? If not, converting from `string[]` to `Item[]` should happen on the fly in render in order to avoid duplicating state.

src/ContextMenu/ContextMenu.tsx
```diff
@@ -103,16 +104,34 @@ class ContextMenu extends React.Component<ContextMenuProps, State> {
...
     focusedItemIndex: 0,
+    items: [],
   }
 
   public static defaultProps: Partial<ContextMenuProps> = {
     align: "left",
     embedChildrenInMenu: false,
   }
 
+  /**
```
peterszerzo: See comment above in this file.

src/HeaderMenu/HeaderMenu.tsx
```diff
@@ -5,7 +5,7 @@ import styled from "../utils/styled"
 
 export interface HeaderMenuProps extends DefaultProps {
   /** Clickable component(s) from which menu appears  */
-  children: React.ReactNode[]
+  children: React.ReactNode | React.ReactNode[]
```
peterszerzo: What is the reason behind this change?

src/ContextMenu/ContextMenu.tsx
```diff
@@ -103,16 +104,34 @@ class ContextMenu extends React.Component<ContextMenuProps, State> {
...
+   */
+
+  public static getDerivedStateFromProps(props: ContextMenuProps) {
+    return {
+      items: props.items.map(item => {
+        if (typeof item === "string") {
+          return { label: item }
+        } else {
+          return item
+        }
```
ImogenF: This would be nicer as a ternary operation.

src/Autocomplete/util/makeItems.ts
TejasQ: Sure! I can do that!

src/ContextMenu/ContextMenu.tsx
TejasQ: My reasoning here is because `state.items` is actually _derived state_ from the props given to this component. Consider,

`<ContextMenu items={["your", "mom", "is", "amazing"]}`

those are initial props. We then [derive the state](https://github.com/contiamo/operational-ui/blob/f450ad06dc3b17016912ee00c801ef27bfd64071/src/ContextMenu/ContextMenu.tsx#L123), which _cannot and should not_ be plain strings, but objects that implement the `IContextMenuItem` interface.

That's the chief reason I put them in state: it communicates intent (to me) a little more clearly. And from that point on, `render` just renders the state: it is essentially a function of the `state` without side effects, maintaining its purity.

src/ContextMenu/ContextMenu.tsx
TejasQ: [Responded.](https://github.com/contiamo/operational-ui/pull/788#discussion_r225105596)

src/HeaderMenu/HeaderMenu.tsx
TejasQ: Previously, one could not have a single child in `HeaderMenu`. This became problematic when you wanted `<Icon name="MyIcon" /> My Project` as a child. There is [more info in the related issue](https://github.com/contiamo/operational-ui/issues/778).

src/ContextMenu/ContextMenu.tsx
peterszerzo: This isn't derived state, it is simply props transformed into something else one-to-one. As in the docs https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops, the `getDerivedStateFromProps` method is intended for pieces of state that depend on changes in props over time e.g. setting intermediate states in transitions.

I also disagree that the intent is clearer: putting things in local state sends a false signal to another developer that the state contains additional information than what is passed in from props, or worse, that it is alterable by `this.setState`, whereas it should at all times just reflect what it has been given through props.

TejasQ: You make a lot of sense. Thanks for pointing this out. Will refactor!

src/ContextMenu/ContextMenu.tsx
```diff
@@ -103,7 +103,10 @@ class ContextMenu extends React.Component<ContextMenuProps, State> {
     }
   }
 
-  public state: State = {
+  private makeItem = (itemFromProps: ContextMenuProps["items"][-1]) =>
```
peterszerzo: Can we add the comment from before, the one that said that the `getDerivedStateFromProps` solution was meant to standardize the public API, only that here we're achieving the same thing differently?

PR #800, opened 2018-10-23
**Fix:** Improve tree component layout

peterszerzo:
# Summary

Some improvements on sizes, spacings and hover effects on the tree component. Also adding a feature where if the tree component is flat (the elements have no children), the left button isn't rendered at all, adding more space on the left, as in the screenshots below:

screen shot 2018-10-23 at 3 31 53 pm

screen shot 2018-10-23 at 3 31 16 pm

# To be tested

Me

Tester 1

Tester 2


src/Tree/Tree.tsx
```diff
@@ -27,7 +27,7 @@ const TreeContainer = styled("div")`
 `
 
 const TreeChildren = styled("div")`
-  margin-left: 16px;
+  margin-left: 14px;
```
TejasQ: Hmmmm. How does this fit into `theme.space` and our design guidelines? @micha-f

src/Tree/Tree.tsx
```diff
@@ -38,65 +38,62 @@ const TreeItem = styled("div")<{
...
  * This is a single-use close button with hard-coded padding to ensure the close icon inside stays readable.
  * @todo look into re-using and formalizing this element.
  */
-const IconButton = styled("div")<{ hidden_?: boolean }>`
+const IconButton = styled("div")<{ hidden_?: boolean; hoverEffect?: boolean }>`
   cursor: pointer;
   width: 20px;
   height: 20px;
+  padding: 4px;
   ${({ hidden_ }) => (hidden_ ? "visibility: hidden;" : "")} ${({ theme }) => `
```
TejasQ: I find this a little hard to read.

src/Tree/Tree.tsx
```diff
@@ -38,65 +38,62 @@ const TreeItem = styled("div")<{
...
-    :not(:last-child) {
-      margin-right: ${theme.space.base}px;
-    }
-  `} :hover {
-    ${({ theme }) => `background-color: ${theme.color.background.light};`};
+  `}
+  ${({ theme, hoverEffect }) =>
+    hoverEffect
+      ? `
+  :hover {
```
TejasQ: I find this and its surrounding block a little hard to read. Maybe we can use object style instead of a template literal here if it helps?

src/Tree/Tree.utils.ts
```diff
@@ -21,8 +21,8 @@ export const getDepth = (tree: Tree): number => {
   return 1 + (getMaxFromList(tree.childNodes.map(getDepth)) as number)
 }
 
-export const getMaxDepth = (trees: Tree[]): number | undefined => {
-  return getMaxFromList(trees.map(getDepth))
+export const getMaxDepth = (trees: Tree[]): number => {
+  return getMaxFromList(trees.map(getDepth)) || 0
```
TejasQ: Is there a reason why this return isn't implicit?

src/Tree/Tree.tsx
peterszerzo: Would you prefer `visibility: ${something => };`? The reason I don't go with that is that I would have to mention the default value "visible" explicitly - arguably not too much trouble for this one case, but if we did it everywhere it would be quite an overhead.

src/Tree/Tree.tsx
peterszerzo: I was thinking about it also - template string styles are just hard to read, and spreading conditionally is not that much trouble. I'll update.

src/Tree/Tree.utils.ts
peterszerzo: Nope, it's a leftover from a logging exercise - I'll update.

src/Tree/Tree.tsx
TejasQ: Es ergiebt Sinn mein Bruder I was just a little concerned with the nesting of the backticks. I wonder if object spreading would be clearer in this case.

src/Tree/Tree.tsx
```diff
@@ -38,74 +38,70 @@ const TreeItem = styled("div")<{
...
-  `} :hover {
-    ${({ theme }) => `background-color: ${theme.color.background.light};`};
-  }
-`
+const IconButton = styled("div")<{ hidden_?: boolean; hoverEffect?: boolean }>(({ theme, hidden_, hoverEffect }) => ({
+  cursor: "pointer",
+  width: 20,
+  height: 20,
+  padding: 4,
+  ...(hidden_ ? { visibility: "hidden" } : {}),
```
TejasQ: TINY 
TINY 
TINY NIT

but can we swap the next line with this one? That way the spreads are grouped together. 😄

PR #803, opened 2018-10-24
Monospaced card items

peterszerzo:
# Summary

Adding a monospace flag to `CardItem` which renders its `value` or `children` prop in a monospaced font, ideal for technical stuff like ID's, SHA's and other information that also shows up in code (resource name used as a variable someplace else).

# To be tested

Me

Tester 1

Tester 2


src/CardItem/CardItem.tsx
```diff
@@ -21,17 +23,22 @@ const CardItemTitle = styled("div")(({ theme }) => ({
...
+const CardItemBody = styled("div")<{ monospace?: boolean }>(({ theme, monospace }) => ({
   color: theme.color.text.default,
   fontFamily: theme.font.family.main,
   fontSize: theme.font.size.body,
   marginTop: theme.space.base,
+  ...(monospace
+    ? {
+        fontFamily: theme.font.family.code,
+      }
+    : {}),
```
ImogenF: Would it not make more sense to combine this with the other fontFamily declaration in L28?

peterszerzo: It absolutely would - updating shortly

PR #805, opened 2018-10-25
**Feature:** Tree reordering

peterszerzo:
# Summary

Adds reordering capabilities to trees:

swap

READ THIS FIRST - the implementation is a bit dense, might be helpful to read this example first:

```js
<Tree
  trees={[
    {
      label: "apples",
      childNodes: []
    },
    {
      label: "oranges",
      childNodes: []
    },
  ]}
  onReorder={(source, target) => console.log(source, target)}
/>
```

What will happen when I play with this?
- if I move "apples" over "oranges", it logs `[ 0 ]` and `[ 1 ]`, indicating that I dragged the 0th child onto the 1st, which would add apples on top of oranges, hence it would stay in place. This is a bit lame for a 2-member list, but for any realistic situation is very predictable, and there is a large wiggle room in which the node can move around and find a destination easily. This is an attempt to improve on the behavior in tools like sketch or adobe XD, although we should stress-test it to make sure it's not too off-the-beaten-path.
- if I move "apples" to the end of the list, it logs `[ 0 ]` and `[ 2 ]`, indicating that I dragged over a fictitious 3rd node above which the node shall be inserted, the result of which would be that apples and oranges will reverse.

Reordering itself has to be done by the client based on these indices, I outlined why I went with this behavior in the component's readme.

# Related issue

# To be tested

Me

Tester 1

Tester 2

### DIFF
--- a/src/Tree/README.md
+++ b/src/Tree/README.md
@@ -53,3 +53,48 @@ The tree component renders a tree structure with collapsable nodes in a filetree
   ]}
 />
 ```
+
+### Reordering
+
+Tree items can be reordered using drag-and-drop. The API for this feature is low-level and assumes that the tree nodes are generated from data available in a different shape in state, making it dangerous and error-prone to update based on a readily reordered set of nodes.
+
+Instead, it provides the paths of the source and target where the nodes are inserted. The client would have to do the actual state update themselves, which is fairly straightforward for a simple flat list, and doable for more complicated use-cases (beware of edge cases).
+
+In the future, this library may provide helpers or sample code to deal with reordering with less boilerplate.
+
+```jsx
+initialState = {
+  courses: ["Appetizer", "Soup", "Spaghetti", "Cake"],
+}
+
+const remove = i => list => list.filter((_, j) => i !== j)
+;<Tree
+  trees={state.courses.map(course => ({
+    tag: "C",
+    label: course,
+    childNodes: [],
+  }))}
+  onReorder={(source, target) => {
+    const sourceIndex = source[0]
+    const targetIndex = target[0]
+
+    /**
+     * The following logic swaps the entries at source and target, catering to an edge case where the element is dragged
+     * to the end of the list. This logic is for illustration purposes only, and needs a bit more work in order to replicate
+     * similar behavior in tools like Photoshop.
+     */
+    let newCourses
+
+    if (targetIndex === state.courses.length) {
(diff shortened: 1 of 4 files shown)


peterszerzo:
[Query editor - reordering](https://trello.com/c/ONao7MMM/521-query-editor-reordering)

src/Tree/README.md
```diff
@@ -53,3 +53,48 @@ The tree component renders a tree structure with collapsable nodes in a filetree
...
+Instead, it provides the paths of the source and target where the nodes are inserted. The client would have to do the actual state update themselves, which is fairly straightforward for a simple flat list, and doable for more complicated use-cases (beware of edge cases).
+
+In the future, this library may provide helpers or sample code to deal with reordering with less boilerplate.
+
+```jsx
+initialState = {
+  courses: ["Appetizer", "Soup", "Spaghetti", "Cake"],
+}
+
+const remove = i => list => [...list.slice(0, i), ...list.slice(i + 1)]
```
fabien0102: Just for the style 😎 

```suggestion
const remove = i => list => list.filter((_, j) => i !==j)
```

src/Tree/README.md
```diff
@@ -53,3 +53,48 @@ The tree component renders a tree structure with collapsable nodes in a filetree
...
+;<Tree
+  trees={state.courses.map(course => ({
+    tag: "C",
+    label: course,
+    childNodes: [],
+  }))}
+  onReorder={(source, target) => {
```
fabien0102: I would like to simplify a bit the API to have only `nextTrees` or something like this to hide the reordering algo into operational-ui.

And also maybe rename `trees` by `data`, I find `trees` a bit redondant inside `Tree` component.
 
```jsx
<Tree
  data={this.state.data}
  onReorder={nextData => this.setState({data: nextData})}
/>
```

peterszerzo: The data that the tree gets is typically not stored in state, because it contains like mouse event listeners and remove listeners. This means that the data that is consumed by the tree is created on the fly from raw data available in state, which may be in a different format or generated algorithmically (as far as my use of this component is concerned). This means that shipping back data in this generated format would require the user to write code that figures out how changes in a generated structure map to changes in the raw structure. This is a process that I would like to discourage with this API. An added advantage is that the user is in control of how the swapping should take place (maybe other elements should also be deleted when any one element is dragged away from its parent).

PR #807, opened 2018-10-26
Website performance

peterszerzo:
# Summary

Fixes the 3s reload an the jumpiness of the website by not rendering examples if sections are out of view. Comes with the tradeoff of fixing the height of the examples section, but at least the navigation is sane and the fix is quick. We can build on top of this in the future.

# To be tested

Me

Tester 1

Tester 2

### DIFF
--- a/styleguide/ReactComponentRenderer.tsx
+++ b/styleguide/ReactComponentRenderer.tsx
@@ -20,10 +20,10 @@ const ReactComponentRenderer: React.SFC<ReactComponentRendererProps> = ({
   tabBody,
 }) => (
   <Consumer>
-    {({ updateActiveComponent }) => (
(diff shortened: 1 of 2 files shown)


ImogenF:
scroll

I don't agree that this solves a jumpiness problem. Now, when you scroll, you have to have a significant amount of the example for a given section in view before the content appears, and you can see the content disappearing from any other partially visible sections.

ImogenF:
Can you display the full contents of any section that is at least partially visible on the screen?

peterszerzo:
I'd love to, as soon as I figure out how `react-waypoint` works.

TejasQ:
Updated the title since this doesn't affect the library.

TejasQ:
Temporarily closing this because of inactivity.

PR #816, opened 2018-10-31
**Fix:** List component

fabien0102:
# Summary

Some little fix during the first integration of this component.

image

# To be tested

Me

Tester 1

- [x] All cards are working as before
  

Tester 2

- [ ] All cards are working as before

### DIFF
--- a/src/Card/Card.tsx
+++ b/src/Card/Card.tsx
@@ -116,6 +116,7 @@ const Container = styled("div")(({ theme }) => ({
 
 const Content = styled("div")`
   padding: ${({ theme }) => theme.space.element}px;
+  height: 100%;
 `
 
 const SectionsContainer = styled("div")<{ stackHorizontal: boolean }>`
--- a/src/Card/Card.tsx
+++ b/src/Card/Card.tsx
@@ -14,6 +14,10 @@ export interface BaseProps extends DefaultProps {
   activeTabName?: string
   /** Callback fired on tab change */
   onTabChange?: (newTabName: string) => void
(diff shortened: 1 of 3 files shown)


src/Card/Card.tsx
```diff
@@ -116,6 +116,7 @@ const Container = styled("div")(({ theme }) => ({
 
 const Content = styled("div")`
   padding: ${({ theme }) => theme.space.element}px;
+  height: 100%;
```
TejasQ: @fabien0102 we noticed some strange behavior (maybe because of this) on your Labs UI with the patched `node_modules/@operational/components`. Care to take a look?

src/Card/Card.tsx
```diff
@@ -14,6 +14,10 @@ export interface BaseProps extends DefaultProps {
   activeTabName?: string
   /** Callback fired on tab change */
   onTabChange?: (newTabName: string) => void
+  /**
+   * Fill all the height of the parent.
+   */
+  fullSize?: boolean
```
peterszerzo: I'd vouch for calling this `fullHeight`, I feel like otherwise people will be wondering why removing `fullSize` still makes the element cover the width of the container (elements like these are typically narrower).

fabien0102: It's to fit with this implementation -> https://github.com/contiamo/operational-ui/blob/master/src/Internals/ControlledModal.tsx#L30

peterszerzo: That makes sense!

PR #818, opened 2018-11-02
**Feature:** TopBar components

peterszerzo:
# Summary

Adding `Topbar`, `TopbarButton` and `TopbarSelect` to components.

screen shot 2018-11-02 at 10 55 05 am

screen shot 2018-11-02 at 10 55 29 am

# To be tested

Me

Tester 1

Tester 2


peterszerzo:
[Add Toolbar Component to operational (extract from Pantheon UI)](https://trello.com/c/rl1V1ArS/540-add-toolbar-component-to-operational-extract-from-pantheon-ui)

ImogenF:
Looks great!
Just a few comments, but all personal opinion so feel free to disagree 😉 

- The TopbarSelect component styling seems a bit off to me. I feel like there should be more distinction between the label and the selected value.
image
Maybe the label could be moved a bit further left, or the font made a bit darker?

- The "x" icon looks really large compared to the others
image

src/TopbarSelect/TopbarSelect.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+const TopbarSelectValueSpan = styled("span")<{ active?: boolean }>`
+  color: ${props => (props.active ? props.theme.color.text.dark : props.theme.color.text.lighter)};
+`
+
+const TopbarSelectLabel = styled("p")`
+  margin: 0px ${props => props.theme.space.base}px 0px 0px;
+  font-size: ${props => props.theme.font.size.fineprint}px;
+  color: ${props => props.theme.color.text.lighter};
+  font-weight: 600;
```
TejasQ: Would the following work?

```suggestion
  font-weight: {$props => props.theme.font.weight.medium};
```

peterszerzo:
@TejasQ the reason we have subcomponents is that said subcomponents can compose in a much more flexible way, either on the left side of the top bar or the right, in any order, repeated any number of times. Contrast that with a card header that is always at the top, and where it is obvious to put header contents in a `title` and `actions` prop. If we went down the same road for top bar, it would turn out pretty confusing.

TejasQ:
Makes sense. I'm also a little unsure of the variants of `Select` and `Button`. It seems a little smelly to me that our own Buttons and Selects don't compose comfortably into this component and we need whole new components that essentially serve the same function. I'm not sure what the right solution might be, but this feels little off because it is currently a duplicated responsibility.

What if we do a major button overhaul and we miss these because the `TopBar` isn't as widely used as generic `Button`s? This is the reason for my concern.

Finally, I'm also a little bit concerned about the naming. Naming things is an age old problem, but I'm wondering if `TopBar` can get confused with `HeaderBar` since `HeaderBar` is _actually_ at the top (and so what does that make `TopBar`?) It gets a little _more_ confusing when a `Page`'s `title` is thrown into the mix. I'd suggest naming this `PageToolbar` or `Toolbar` in order to more clearly describe its intended usage.

TejasQ:
[Improve toolbar dropdowns](https://trello.com/c/GyZ7eHen/568-improve-toolbar-dropdowns)

peterszerzo:
`Button`'s and `TopbarButton`'s are very different both visually and in the way they interact with siblings (same goes with `Select` and `TopbarSelect`), to the point when we'll set up API's that cater to both places, it'll turn into a maintainability nightmare (think of all the ternaries in styles, the discriminating union interfaces etc.). Having separate components keeps both usage and readability sane, and allows components to evolve independently.

As far as naming goes, clarity and conciseness will always fight a battle, and at this point I'd rather stick to the way we call them during the design process. I'll clear up the possible naming confusions in the documentation.

TejasQ:
> Button's and TopbarButton's are very different both visually and 

Ah, I see. I misunderstood: I thought the blue thing that looked like a `Button` from the screenshot attached to this PR was a `TopbarButton` in this case. My bad. 

image

That makes sense.

Awesome. It looks good to me so far. One final thought I have is this:
image

We have different padding (revealed via hover) on the elements of the _other_ `Topbar`, `HeaderMenu` where the selects/buttons (Avatar in this case) have the padding _inside them_, instead of the actual _bar_ itself padding its children. 

image

image

Design-wise, I don't have the context, but these two cases aren't consistent and I'm wondering if that's the direction we're intentionally wanting to go here.

peterszerzo:
@TejasQ the blue thing is a barebones regular `Button` component, whereas `TopbarButton` is used to render the undo-redo buttons.

The idea is that there are buttons and selects that blend in seamlessly into top bar, taking up the full height and having nice opinionated borders, and, when needed, other components can be added.

fabien0102:
@TejasQ Is this ready to merge?

TejasQ:
It looks good to me. If you approve, I say we can merge it. :sligh

PR #789, opened 2018-10-11
**Fix:** Remove lodash omit from Code.tsx

adeelibr:
# Summary

Remove `lodash/omit` from `Code.tsx`

# Related issue

https://github.com/contiamo/operational-ui/issues/752

PR #819, opened 2018-11-04
**Fix:** Remove lodash uuid from Checkbox.tsx #789

adeelibr:
Summary

1.  Remove lodash/uuid from Checkbox.tsx
2.  Convert Checkbox.tsx to dumb component

After this PR remove lodash from:

https://github.com/contiamo/operational-ui/blob/8c0d472a09a8f56d5fa57e1add31a299e0e341a5/src/utils/constants.ts#L217

https://github.com/contiamo/operational-ui/blob/8c0d472a09a8f56d5fa57e1add31a299e0e341a5/src/ContextMenu/ContextMenu.tsx#L131

And from `OperationalUI`

https://github.com/contiamo/operational-ui/blob/8c0d472a09a8f56d5fa57e1add31a299e0e341a5/src/OperationalUI/OperationalUI.tsx#L3

https://github.com/contiamo/operational-ui/blob/8c0d472a09a8f56d5fa57e1add31a299e0e341a5/src/OperationalUI/OperationalUI.tsx#L4

https://github.com/contiamo/operational-ui/blob/8c0d472a09a8f56d5fa57e1add31a299e0e341a5/src/OperationalUI/OperationalUI.tsx#L5

### DIFF
--- a/src/Checkbox/Checkbox.tsx
+++ b/src/Checkbox/Checkbox.tsx
@@ -94,9 +94,10 @@ export interface CheckboxProps {
 }
 
 const Checkbox: React.SFC<CheckboxProps> = ({ value, onChange, label, disabled }) => {
-  const uuid = Math.random()
+  const uuid = `checkbox_${Math.random()
     .toString(36)
-    .slice(2)
+    .slice(2)}`
+
   return (
     <div style={disabled ? { opacity: 0.6 } : {}}>
       <Input id={uuid} type="checkbox" checked={Boolean(value)} onChange={() => onChange(!value)} disabled={disabled} />
--- a/src/Checkbox/Checkbox.tsx
+++ b/src/Checkbox/Checkbox.tsx
@@ -1,4 +1,3 @@
-import uniqueId from "lodash/uniqueId"
 import * as React from "react"
 import { keyframes } from "react-emotion"
 
@@ -94,25 +93,16 @@ export interface CheckboxProps {
   disabled?: boolean
 }
 
-class Checkbox extends React.Component<CheckboxProps> {
-  private uuid = uniqueId("checkbox_")
-
-  public render() {
-    const { value, onChange, label, disabled } = this.props
-
-    return (
-      <div style={disabled ? { opacity: 0.6 } : {}}>
-        <Input
-          id={this.uuid}
-          type="checkbox"
-          checked={Boolean(value)}
-          onChange={() => onChange(!value)}
-          disabled={disabled}
-        />
-        <Label htmlFor={this.uuid}>{label}</Label>
-      </div>
-    )
-  }
+const Checkbox: React.SFC<CheckboxProps> = ({ value, onChange, label, disabled }) => {
+  const uuid = Math.random()
+    .toString(36)
+    .slice(2)
+  return (
+    <div style={disabled ? { opacity: 0.6 } : {}}>
+      <Input id={uuid} type="checkbox" checked={Boolean(value)} onChange={() => onChange(!value)} disabled={disabled} />
+      <Label htmlFor={uuid}>{label}</Label>
+    </div>
+  )
 }
 
 export default Checkbox


adeelibr:
I am trying to remove `lodash/get` from `constants` file

https://github.com/contiamo/operational-ui/blob/8c0d472a09a8f56d5fa57e1add31a299e0e341a5/src/utils/constants.ts#L217

This is the approach I have come up with, can you kindly give some feedback on it.

## Approach 1

```
export const get = (obj: object, path: string, defaultValue: string) => {
  const fullPath = path
    .replace(/\[/g, ".")
    .replace(/]/g, "")
    .split(".")
    .filter(Boolean);

  const isNestedValid = (step: string) => {
    return !(step && (obj = obj[step]) === undefined);
  }

  return fullPath.every(isNestedValid) ? obj : defaultValue;
}
```

This works both ways with dot notation `.` and `[]` notation.

#### Usage

```
get (theme, 'color.text.default', 'red');
get (theme, 'color[text].default', 'red');
```

## Approach 2

```
export const get = (object: object, path: string, defaultValue: string) => {

  const deepAccessUsingString = (obj: object, p: string) => {
    return p.split(".").reduce((nestedObject, key: string) => {
      if (nestedObject && key in nestedObject) {
        return nestedObject[key];
      }
      return undefined;
    }, obj);
  }

  try {
    const value = deepAccessUsingString(object, path);
    if (!value) {
      throw new Error("Bad Path");
    }
    return value;
  } catch (error) {
    return defaultValue;
  }
}
```

#### Usage

```
get (theme, 'color.text.default', 'red');
```

This works for only dot notations. The approach I take in here is if the path is valid I return the found value otherwise I throw an error which in turns returns the default value.

### Approach 3

This will result in an overall bigger change

```
const get = (fn) => {
    try {
        return fn();
    } catch (error) {
        return undefined;
    }
} 
```

#### Usage

```
get (() => theme.color.text.primary.foo.baz) || "red";
```

But this will have a bigger impact on the entire API, since this change will have to be done everywhere where `expandColor` is being used. In which we pass in the `theme` and `colorToBeExpanded` (the path inside the `theme` object);

I think this might be the least preferred approach, but I still wanted to mention this here.

<hr />

After I am done with removing `get` I'll move on to the `ContextMenu.tsx` file to remove `isEqual` method.

<hr />

I would really appreciate if you guys could pitch in your feedback here, that even if this is the right approach to do it or not.

CC @peterszerzo @TejasQ  @fabien0102 😄

adeelibr:
In `Approach 1` where I do this

```
const isNestedValid = (step: string) => {
    return !(step && (obj = obj[step]) === undefined);
}
return fullPath.every(isNestedValid) ? obj : defaultValue;
```

This can be problematic, since it is mutating the data `obj`.

This can be replaced by `reduce` method, which goes like this.

```
return fullPath.reduce((acc, cv) => {
    try {
      acc = acc[cv]
      if (!acc) {
        throw new Error("Bad Path");
      }
    } catch (e) {
      return defaultValue
    }
    return acc
}, obj)
``` 

Here `fullPath` is an array, e.g, =>  `[ 'color',  'text', 'default' ]`

fabien0102:
I really don't see the point of removing lodash, especially with our current utilisation. We are importing only few small parts, not the entire library.
Lodash is not bad or evil, it's well tested and works out of the box! Some methods can be replace by native one (as `map` for an array, I agree) but should we really want to reimplement `get` or `debounce`? 🤔 It's already well implemented and well tested, so why reinventing the weel?

Winning some Kb here it's not free, we have to test and maintain the piece of code that we introduce, so I don't really this "remove lodash" as a "win" for us.

### Some stats
https://bundlephobia.com/result?p=lodash.get@4.4.2 lodash.get is actually 4.5kb gzipped, 
http://bytesizematters.com/ your snippet in `Approche 2` is 565 B

It's actually not a big cost for the safety of using a well tested and production ready method :wink:

To conclude, I'm not in favor of this pre-mature optimization, we are not fighting against kb (yet) when every components and everything will be perfect, and we have a hard requirement to deliver a product on mobile with bad internet connectivity, why not, but for now the cost is bigger than the value. (also before this, we should maybe migrate to preact if we have this bundle size as goal)

@adeelibr I hope that you understand my opinion, I also understand yours and I want to thanks you for every contributions that you made on this project 👍 I just prefer clean and tested code rather than winning some kB 😉

fabien0102:
After, I don't say, it can be fun to reimplement a version of lodash size oriented 😉

adeelibr:
Sure I can understand. So should I delete this PR?

fabien0102:
Yep, let's close this PR! Thanks for your contributions again 💯

PR #820, opened 2018-11-06
 **Breaking:** Making page prop non-optional.

eugengarkusha:
# Summary
   making "page" parameter non-optional in PaginatorProps

The motivation for this change is the decrease in the cognitive load  for :
the users:
  In case if the page was not specified user had to read the source code or the doc of the component 
  to be sure that it takes page 1 by default

and maintainers:
  removed Paginator.defaultProps
  removed unsafe '!' operators from the code

# Related issue

# To be tested

Me

Tester 1

Tester 2

### DIFF
--- a/src/Paginator/Paginator.tsx
+++ b/src/Paginator/Paginator.tsx
@@ -1,5 +1,3 @@
-let isUserWarnedAboutDeprecation = false
-
 import * as React from "react"
 
 import Button from "../Button/Button"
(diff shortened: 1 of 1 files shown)


fabien0102:
@TejasQ Should we merge this PR today and release a new major version soon? 😈

TejasQ:
Let's do a new major release on Monday, but let's maybe get a minor/patch out today. 🙂

PR #822, opened 2018-11-06
*Feature:* Improve the pushMessage stacking

fabien0102:
# Summary

In operational-ui, we have a nice pushMessage that stacks our errors in the corner of the screen.

Actually, if we have a lot of errors (as 500 on polling) it's not really userfriendly and quite paintful to dismiss the stack.

The idea is to introduce a `count`, so if we have the same error, again and again, we will have `(20) Error - oh no` instead of a stack of 20 messages

## Demo
a59fc615-bec5-4cff-9988-da941df43821

# To be tested

Me

Tester 1

Tester 2

### DIFF
--- a/src/OperationalUI/OperationalUI.tsx
+++ b/src/OperationalUI/OperationalUI.tsx
@@ -234,15 +234,16 @@ class OperationalUI extends React.Component<OperationalUIProps, State> {
   }
 
   private pushMessage = (message: IMessage) => {
-    this.setState(prevState => {
-      const previousMessageWithSamePayload = prevState.messages.find(
-        m => m.message.body === message.body && m.message.type === message.type,
-      )
+    const hasSamePayload = (m: { message: IMessage }) =>
+      m.message.body === message.body && m.message.type === message.type
 
-      if (previousMessageWithSamePayload) {
+    this.setState(prevState => {
+      const hasPreviousMessageWithSamePayload = Boolean(prevState.messages.find(hasSamePayload))
+
+      if (hasPreviousMessageWithSamePayload) {
         return {
           messages: prevState.messages.map(m => {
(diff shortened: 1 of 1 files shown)


fabien0102:
[Avoid errors spamming](https://trello.com/c/KYYRyEEz/572-avoid-errors-spamming)

src/OperationalUI/OperationalUI.tsx
```diff
@@ -239,6 +232,35 @@ class OperationalUI extends React.Component<OperationalUIProps, State> {
...
+  private pushMessage = (message: IMessage) => {
+    this.setState(prevState => {
+      const previousMessageWithSamePayload = prevState.messages.find(
+        m => m.message.body === message.body && m.message.type === message.type,
+      )
+
+      if (previousMessageWithSamePayload) {
+        return {
+          messages: prevState.messages.map(m => {
+            if (m.message.body === message.body && m.message.type === message.type) {
```
TejasQ: This check can be hoisted since it's common.

src/OperationalUI/OperationalUI.tsx
```diff
@@ -239,6 +232,35 @@ class OperationalUI extends React.Component<OperationalUIProps, State> {
       </ThemeProvider>
     )
   }
+
+  private pushMessage = (message: IMessage) => {
+    this.setState(prevState => {
+      const previousMessageWithSamePayload = prevState.messages.find(
```
TejasQ: We're not really using this value but instead using it as a check. I would propose the following adjustment in order to communicate intent a little clearer:

```suggestion
      const isPreviousMessageWithSamePayload = Boolean(prevState.messages.find(
```

**(Be sure to add the closing `)` on line `240` if you accept this)**

PR #836, opened 2018-11-19
**Fix:** Make select filter case insensitive

TejasQ:
# Summary
This PR makes filtering through dropdowns case _insensitive_.

# To be tested

Me

Tester 1

Tester 2


TejasQ:
[Filtering in Operational Dropdown should not be case sensitive](https://trello.com/c/KYQ1HXkH/587-filtering-in-operational-dropdown-should-not-be-case-sensitive)

PR #841, opened 2018-11-21
Add a breaking change warning for Paginator in v12

TejasQ:
# Summary
Before we do our next minor release, let's add a deprecation warning to warn about the breaking change in v12.

# Related issue
#820 

# To be tested

Tester 1

Tester 2

### DIFF
--- a/src/Paginator/Paginator.tsx
+++ b/src/Paginator/Paginator.tsx
@@ -1,3 +1,15 @@
+/**
+ * @todo remove this in v12.
+ */
+console.warn(`[operational-ui]: Deprecation Warning
+===
+
+<Paginator />'s page prop is now _mandatory_, meaning in the next major
+release, your code (if TypeScript) will not compile. Please add an explicit
+\`page\` prop to your Paginator if you'd like to stay up to date.
+
+Issue: https://github.com/contiamo/operational-ui/pull/820`)
+
 import * as React from "react"
 
 import Button from "../Button/Button"
--- a/src/Paginator/Paginator.tsx
+++ b/src/Paginator/Paginator.tsx
@@ -1,3 +1,5 @@
+let isUserWarnedAboutDeprecation = false
+
 import * as React from "react"
 
 import Button from "../Button/Button"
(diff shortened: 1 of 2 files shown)


fabien0102:
To avoid useless I will prefer to have this `console.warn` only if we have instance of `Page` without explicit `page`

fabien0102:
And regarding the implementation and usage of `utils/deprecate` I think we can delete this piece of code.
Since we package the app, I'm not really sure of the `process.env.NODE_ENV !== production` (is it really execute when operational-ui is execute in another project? 🤔 )

TejasQ:
> To avoid useless noise I would prefer to have this console.warn only if we have instance of `Page` without explicit `page`

Do you mean `Paginator` without an explicit `page`? That's a little bit hard to check given the `defaultProps`. 🤔 Do you have an idea about this?

> And regarding the implementation and usage of `utils/deprecate` I think we can delete this piece of code.
> Since we package the app, I'm not really sure of the `process.env.NODE_ENV !== production` (is it really execute when operational-ui is execute in another project? 🤔 )

Good catch. I'm not sure actually. We'll have to look into that. 🙂

fabien0102:
Of course `Paginator` ^^, my bad.

mmm, I think we can remove the `defaultProps`, keep the props optional, and in the constructor, if `page` is null -> console.warn (and assign a default value)

TejasQ:
Great, great, _great_ idea! I'll do this. Thanks!

fabien0102:
Just for information, react have this kind of entry point to deal with development mode or not:

```js
'use strict';

if (process.env.NODE_ENV === 'production') {
  module.exports = require('./cjs/react.production.min.js');
} else {
  module.exports = require('./cjs/react.development.js');
}
```

fabien0102:
https://github.com/contiamo/operational-scripts/issues/31

TejasQ:
Done.

src/Paginator/Paginator.tsx
```diff
@@ -87,15 +87,32 @@ const Container = styled("div")(({ theme }) => ({
   alignItems: "center",
 }))
 
-const Paginator: React.SFC<PaginatorProps> = ({ itemCount, itemsPerPage, page, onChange, ...props }) => {
+const Paginator: React.SFC<PaginatorProps> = ({ itemCount, itemsPerPage, page: explicitPage, onChange, ...props }) => {
+  const page: number = explicitPage || 1
+
+  if (!explicitPage) {
```
fabien0102: Just a last comment, can you add a little boolean to avoid multiple warnings? Something like `isWarningAlreadyDisplayed`

PR #851, opened 2018-12-04
**Feature:** Support ReactElements in ContextMenu

TejasQ:
- Contact does not manipulate margins and correctly wraps
- ContextMenu supports React Elements as label and icon props
- Hint\'s API now allows moving its toolbar around
- Topbar needed an update to be compatible with ContextMenu

# Summary
In order to allow `ContextMenu`s to be a little more expressive, we have decided to allow React Elements in a given `ContextMenuItem`'s `icon` and `label` props. 

This will enable more sophisticated menus, like this:

image

# To be tested

Me

Tester 1

Tester 2

### DIFF
--- a/src/ContextMenu/ContextMenu.Item.tsx
+++ b/src/ContextMenu/ContextMenu.Item.tsx
@@ -106,10 +106,10 @@ const Content: React.SFC<{ value: StringOrItem }> = ({ value }) => {
   )
 }
 
-const icon = (props: Props) => {
+const ContextMenuItemIcon: React.SFC<Pick<Props, "item" | "iconLocation">> = props => {
   // If item is just a string,
   if (typeof props.item === "string") {
-    return null
+    return <></>
   }
 
   // If it's an object with an icon property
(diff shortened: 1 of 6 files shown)


src/ContextMenu/ContextMenu.Item.tsx
```diff
@@ -18,9 +18,9 @@ export interface Props {
 }
 
 export interface IContextMenuItem<TValue = any> {
-  label: string
+  label: string | React.ReactElement<any>
```
peterszerzo: Can we discriminate here? Forcing an absent icon when the label is a react element seems like a good constraint.

src/ContextMenu/ContextMenu.Item.tsx
```diff
@@ -106,21 +106,33 @@ const Content: React.SFC<{ value: StringOrItem }> = ({ value }) => {
...
-const InPlaceIcon = (props: Props) =>
-  typeof props.item !== "string" ? (
-    <ContextMenuIcon
-      iconlocation_={props.iconLocation}
-      color={props.item.iconColor}
-      left={props.iconLocation === "left" || !props.iconLocation}
-      name={props.item.icon as IconName}
-    />
-  ) : null
+const icon = (props: Props) => {
```
peterszerzo: Uppercase the name here I'd say - even if you don't intend it to be a proper react component, it still is one.

peterszerzo:
The centerline of the help tooltip should line up with the centerline of the plus signs, and should be the same size actually, otherwise it gets lost.

src/ContextMenu/ContextMenu.Item.tsx
TejasQ: I don't think that's a good constraint because the screen I've posted above has both an icon _and_ a React Element. 😄

src/ContextMenu/ContextMenu.Item.tsx
TejasQ: Good catch! 😄

src/ContextMenu/ContextMenu.Item.tsx
peterszerzo: Oh never mind I thought the icon specified by the icon prop was rendered on the left.

src/Hint/Hint.tsx
```diff
@@ -3,7 +3,7 @@ import { DefaultProps } from "../types"
 import styled from "../utils/styled"
 
 import Icon from "../Icon/Icon"
-import Tooltip from "../Tooltip/Tooltip"
+import DiscriminatedTooltip from "../Tooltip/Tooltip"
```
fabien0102: 🤔

TejasQ:
@fabien0102 updated. Thanks! 🚀

Issue #809, opened 2018-10-26
Create a `Foldable` component


fabien0102:
### Summary
![image](https://user-images.githubusercontent.com/1761469/47575415-557b1a00-d942-11e8-94d4-c745c6f3e122.png)

The idea is to have a nice foldable component

### API Proposal

```tsx
<Foldable title="Jobs"><Code /></Foldable>

**Comment by TejasQ** [comment]
@micha-f maybe this would help even our card sections and resource-role-assigner.

What if we had a 

```jsx
<Foldable>
  {({ Toggler }) => {
    return <Card title={<>Hello <Toggler />}>My Card</Card>
  }}
</Foldable>
```

that would make a card foldable? This could be applied to any component that we would like to be foldable. 😄

TejasQ:
This was merged.

PR #854, opened 2018-12-06
**Feature:** Foldable component

TejasQ:
# Summary
This PR hijacks #809 and adds a `Foldable` component that we can later use for #809's specific use case. It adds `Foldable` because according to our designs, we need to fold:

- `CardSection`s, and
image

- `Group`s which will be an upcoming component in a newer PR.
image

- `Code` per #809 which will _also_ be an upcoming PR.

# To be tested

Me

Tester 1

Tester 2


TejasQ:
[Add collapsible option to card section](https://trello.com/c/E1pPsP4E/649-add-collapsible-option-to-card-section)

PR #860, opened 2018-12-11
**Feature:** Topbar improvements

peterszerzo:
# Summary

Toolbar improvements:

* [x] Toolbar background color should be #f8f8f8
* [x] dropdown width should always be at least that of the trigger element
* [x] include the trigger element in the dropdown shadow when expanded

screen shot 2018-12-11 at 11 45 50 am

# To be tested

Me

Tester 1

Tester 2

### DIFF
--- a/src/TopbarSelect/TopbarSelect.tsx
+++ b/src/TopbarSelect/TopbarSelect.tsx
@@ -16,13 +16,26 @@ export interface TopbarSelectProps {
   onChange?: (newLabel: string | React.ReactElement<any>) => void
 }
 
-const TopbarSelectContainer = styled("div")`
+export interface State {
+  renderedWidth?: number
+}
+
+const TopbarSelectContainer = styled("div")<{ isActive: boolean }>`
   height: ${props => props.theme.topbarHeight}px;
   display: flex;
   align-items: center;
   padding: 0px ${props => props.theme.space.medium}px;
+  box-shadow: ${props => (props.isActive ? props.theme.shadows.popup : "none")};
+  border-bottom: ${props => (props.isActive ? "1px" : "0px")};
+  border-color: ${props => props.theme.color.border.default};
+  cursor: pointer;
+  background-color: ${props => (props.isActive ? props.theme.color.white : "transparent")};
   :hover {
-    background-color: ${props => props.theme.color.background.lighter};
+    background-color: ${props => (props.isActive ? props.theme.color.white : props.theme.color.background.lighter)};
+  }
+  & svg {
+    /** Icons are purely presentational and click events are handled upstream */
+    pointer-events: none;
   }
 `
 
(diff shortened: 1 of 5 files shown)


peterszerzo:
[Improve toolbar dropdowns](https://trello.com/c/GyZ7eHen/568-improve-toolbar-dropdowns)

src/TopbarSelect/TopbarSelect.tsx
```diff
@@ -16,14 +16,25 @@ export interface TopbarSelectProps {
   onChange?: (newLabel: string | React.ReactElement<any>) => void
 }
 
-const TopbarSelectContainer = styled("div")`
+export interface State {
+  renderWidth?: number
```
TejasQ: What does this mean? `renderWidth`?

src/TopbarSelect/TopbarSelect.tsx
```diff
@@ -16,14 +16,25 @@ export interface TopbarSelectProps {
...
   padding: 0px ${props => props.theme.space.medium}px;
+  box-shadow: ${props => (props.isActive ? props.theme.shadows.popup : "none")};
+  border-bottom: ${props => (props.isActive ? "1px" : "0px")};
+  border-color: ${props => props.theme.color.border.default};
+  cursor: pointer;
   :hover {
     background-color: ${props => props.theme.color.background.lighter};
   }
+  & svg {
+    cursor: pointer;
```
TejasQ: Would `pointer-events: none` make more sense in this case since we want the container of the SVG to be clicked/handle clicks and the SVG's click events to bubble to its container?

peterszerzo: I agree, that makes much more sense.

src/TopbarSelect/TopbarSelect.tsx
TejasQ: To be honest, I'd expect `renderWidth` to be an identifier of a function, since `render` is a verb. We also have `renderTabs` in this codebase.

`renderedWidth` or `containerWidth` would make a little more sense to me.

peterszerzo:
@TejasQ yes that's correct - since the context menu's container is outside, it is hard to control its width reliably.

About the centering: that has to do with the design of the icons and how it looks with chevron down vs. chevron up. Right now, the endpoints of both chevrons are at the same place, which means that it's the bounding box of the icons are centered and not their perceived centerline. We can revisit the icons on a separate branch if you like.

TejasQ:
> @TejasQ yes that's correct - since the context menu's container is outside, it is hard to control its width reliably.

The context menu's container is outside of `TopbarSelect` you mean? 🤔

peterszerzo:
That's right!

TejasQ:
This is quite nice the way it is but I'm wondering if we can avoid having local component state around just for this. Maybe using the newer way that the React community suggest for working with refs? `React.createRef`? 🤔

peterszerzo:
I don't see callback refs being deprecated: https://reactjs.org/docs/refs-and-the-dom.html#callback-refs, they are even mentioned as a way to have more fine-grained control over them. Is there something coming that would deprecate them?

I rewrote using `createRef` but it turned out to be longer:

```
export class TopbarSelect extends React.Component<TopbarSelectProps, Readonly<State>> {
  public state: State = {
    renderedWidth: undefined,
  }

  private containerRef = React.createRef()

  public componentDidMount() {
    this.updateRenderedWidth()
  }

  public componentDidUpdate() {
    this.updateRenderedWidth()
  }

  private updateRenderedWidth() {
    const node = this.containerRef.current as HTMLElement
    if (!node) {
      return
    }
    const renderedWidth = node.clientWidth
    if (renderedWidth !== this.state.renderedWidth) {
      this.setState(() => ({
        renderedWidth,
      }))
    }
  }

...
...
```

Is there a shorter way?

TejasQ:
I'm not sure. Turning this component into a class-based component just to keep track of the width of an HTML element seems a little smelly which is why I'm thinking about it.

If this is the only/best way we can see, sure let's merge it but I think there might be a simpler/more efficient way I'm not seeing right now because I'm caught up with another task.

TejasQ:
That said, _correctness_ always comes after _serving_ our customers so if this is needed to ship something urgent, let's move faster towards merging it.

peterszerzo:
I don't see anything incorrect about it. Not everything can be done in CSS (which is why we have things like `ResizeObserver`, and which is why `clientWidth` is exposed in the first place), and this accommodates an API preference which allows reusing the context menu. And keeping it in state is the simplest and clearest way you can guarantee that a change in a local value results in a re-render.

TejasQ:
There's certainly nothing _incorrect_ about it. What I meant to question was whether it's "correct" in the sense of being the most optimal solution as well. This is still something I'm thinking about...

peterszerzo:
Opened https://github.com/contiamo/operational-ui/issues/865 to track a conversation about it.

TejasQ:
So after thinking about this some more, might I propose creating a ref and then passing it to the relevant node, without keeping track of it in the state? Kind of like:

```ts
const selectElement = React.createRef<HTMLDivElement>()

const TopbarSelect: React.SFC<TopbarSelectProps> = ({ label, selected, items, onChange, ...props }) => {
  const haveSelectElement = () => selectElement !== null && selectElement.current !== null

  return (
    <ContextMenu
      condensed
      items={items}
      width={haveSelectElement() ? selectElement.current!.clientWidth : 0}
      onClick={newItem => {
        if (onChange) {
          onChange(newItem.label)
        }
      }}
    >
      {isActive => (
        <TopbarSelectContainer {...props} isActive={isActive} innerRef={selectElement}>
          <TopbarSelectLabel>{label}</TopbarSelectLabel>
          <TopbarSelectValue>
            <TopbarSelectValueSpan active={Boolean(selected)}>{selected}</TopbarSelectValueSpan>
            <Icon name={isActive ? "ChevronUp" : "ChevronDown"} size={12} />
          </TopbarSelectValue>
        </TopbarSelectContainer>
      )}
    </ContextMenu>
  )
}
```

That would make the component quite simpler and also prepare it for Hooks whenever it stabilizes, while avoiding the complexities of having a class-based component.

Thoughts?

peterszerzo:
It looks like the snippet won't work reliably if there are multiple `TopbarSelect`'s on the page - since there will be a single ref, only a single width will be recorded, forcing all the others to that width.

TejasQ:
You're right. Great, let's merge it then.

Issue #862, opened 2018-12-11
Group: Header's bottom margin is too darn high


TejasQ:
![image](https://user-images.githubusercontent.com/9947422/49795480-5b825a00-fd3a-11e8-9498-bfff0d101942.png)

The bottom margin of the title needs to be `16`px and _not_ `20`px that it is right now.

Issue #861, opened 2018-12-11
Foldable: Toggler's typing is too strong


TejasQ:
![image](https://user-images.githubusercontent.com/9947422/49795413-37bf1400-fd3a-11e8-98e6-5256773d1b90.png)

Unfortunately, I only tested this API in JavaScript (in the playground in the docs) *when used inside `children`* and missed that some props should be optional. Will fix soon. My sincerest apologies, everyone.

PR #863, opened 2018-12-11
**Fix:** Foldable and Group

TejasQ:
# Summary
This PR closes issues #862 and #861, while making the `Toggler` exposed by `Foldable` to its children a little more flexible: it can now do other things `onClick` and reflect on a different value for `isFolded`, allowing it to be used in more circumstances, like an accordion as we have on the IdP.

# Related issue
#862 
#861 

# To be tested

Me

Tester 1

Tester 2


src/Foldable/README.md
```diff
@@ -127,3 +127,80 @@ It is also possible to control the initial state of this component like below. I
...
 </Foldable>
 ```
+
+## Advanced Case: An Accordion
+
+In some cases, you'd want a one component's `isFolded` property to be aware of its surroundings. We might solve this with local component state in React as so.
+
+```jsx
+initialState = {
+  active: 0,
```
peterszerzo: The example would be clearer if this were a boolean I'd say.

TejasQ: It needs to be an index reference to a currently active section though...

peterszerzo: `activeSection` then maybe?

TejasQ:
> Since more and more use-cases are popping up when folded state is managed from the outside, I'm wondering whether this component is worth having around at all. Managing one folded state case-by-case is not that much extra work, and adds a lot of clarity and control.

You might be right. 😄 Let's examine in an issue. The initial reason for creating this component is because `CardSection` and `Group` both needed identical behavior that was well tested. I figured a single component that handles this behavior reliably might be helpful. Of course, I could be wrong. Nothing a little discussion can't solve.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@operational/components",
-  "version": "12.3.0",
+  "version": "12.3.1-rra.1",
```
fabien0102: Please avoid to commit this ^^

peterszerzo:
Added https://github.com/contiamo/operational-ui/issues/869 to discuss removing foldable, as I suggested halfway through this discussion.

peterszerzo:
We agreed offline that despite deprecation discussions, we're merging this PR to meet an internal need.

PR #866, opened 2018-12-12
**Fix:** Header menu

TejasQ:
# Summary
We currently have a regression on `HeaderMenu` throughout our platform: 
image

This PR learns from #860 and implements identical behavior to correctly size and position `HeaderMenu`s context menu.

# Related issue
#860 

# To be tested

Me

Tester 1

Tester 2

### DIFF
--- a/src/HeaderMenu/HeaderMenu.tsx
+++ b/src/HeaderMenu/HeaderMenu.tsx
@@ -16,8 +16,16 @@ export interface HeaderMenuProps extends DefaultProps {
   align?: ContextMenuProps["align"]
 }
 
+interface HeaderMenuState {
+  menuWidth: number
+}
+
 const backgroundColor = "hsla(0, 0%, 100%, 0.1)"
 const boxShadow = "0 3px 6px rgba(0, 0%, 0%, 0.3)"
+
+const HeaderContextMenu = styled(ContextMenu)`
+  height: 100%;
+`
 
 const Container = styled("div")<{
   align: HeaderMenuProps["align"]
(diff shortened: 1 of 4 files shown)


src/HeaderMenu/HeaderMenu.tsx
```diff
@@ -68,21 +76,40 @@ const Container = styled("div")<{
...
+  }
+
+  public componentDidMount() {
+    if (this.menuNode === null) {
+      return
+    }
+    if (this.menuNode.current === null) {
+      return
+    }
+    this.setState(() => ({ menuWidth: this.menuNode.current!.clientWidth }))
```
peterszerzo: Can we compare this implementation with the way `TopbarSelect` does it and have a consistent implementation? Doing this sync only once on mount may leave the layout inconsistent in case widths change, in case things jump around slightly because of rendering artifacts, or if the component's implementation ever changes so that widths vary for one reason or another.

TejasQ: I understand. 

Would you suggest I add a `componentDidUpdate` handler just like on `TopbarSelect`?

peterszerzo: Let's discuss potential improvements to https://github.com/contiamo/operational-ui/blob/master/src/TopbarSelect/TopbarSelect.tsx#L65-L90 and copy-paste it in here I'd say.

TejasQ: I think asserting `as HTMLElement` on something that's possibly `null` is dangerous:

https://github.com/contiamo/operational-ui/blob/82599cbaa94156c081ca3dc29009b6a954d23da2/src/TopbarSelect/TopbarSelect.tsx#L80

Instead, I think

```ts
  private updateRenderedWidth() {
    if (!this.containerRef) {
      return
    }
    if (this.containerRef.current === null) {
      return
    }

    const node = this.containerRef.current // the type should be inferred correctly now
    const renderedWidth = node.clientWidth
    if (renderedWidth !== this.state.renderedWidth) {
      this.setState(() => ({
        renderedWidth,
      }))
    }
  }
```

PR #873, opened 2018-12-14
**Feature:** Add Toggle component

fabien0102:
# Summary

image

## API

```tsx
export interface Value {
  label: React.ReactNode
  value: string
}

export interface ToggleProps {
  /** Options available */
  options: [Value, Value]
  /** Current value */
  value: Value["value"]
  /** Callback trigger on any changes */
  onChange: (newValue: Value["value"]) => void
  /** Condensed option */
  condensed?: boolean
}
```

## Usage

```tsx
initialState = {
  toggle: "visual",
}
;<Toggle
  value={state.toggle}
  onChange={value => setState({ toggle: value })}
  options={[{ label: "Visual", value: "visual" }, { label: "Code", value: "code" }]}
/>
```

# Related issue

https://trello.com/c/payZqE8i

# To be tested

Me

Tester 1

Tester 2


src/Toggle/Toggle.tsx
```diff
@@ -0,0 +1,70 @@
...
+  /** Current value */
+  value: Value["value"]
+  /** Callback trigger on any changes */
+  onChange: (newValue: Value["value"]) => void
+  /** Condensed option */
+  condensed?: boolean
+}
+
+const Container = styled("div")`
+  button {
```
peterszerzo: Two observations here:
* this can go into the `Button` styled component below so it's more localized.
* if you're overriding the padding, I don't think we should be using `Button` anyhow. I think we'd be better off to apply the font sizes manually than fighting the internals of another component.

src/Toggle/Toggle.tsx
```diff
@@ -0,0 +1,70 @@
...
+    padding: 0 ${props => props.theme.space.element}px;
+  }
+
+  > :not(:last-child) {
+    margin-right: 0;
+    border-radius: ${({ theme: { borderRadius } }) => `${borderRadius}px 0 0 ${borderRadius}px`};
+  }
+
+  > :not(:first-child) {
+    left: -2px; /* Merge buttons */
```
peterszerzo: Does this work in IE11 for example? If not, I suspect Safari may give trouble also. Negative margins of -1 tend to work well also, not sure what is better here.

fabien0102: It depend of the zoom, how it looks in safari? Can you screenshot? (I'm working on chrome btw 😅 )

peterszerzo:
screen shot 2018-12-14 at 3 33 34 pm

Compared to the zeplin screenshot, you have a bit too much blur. I'd keep it at 6-8px.

src/Toggle/Toggle.tsx
fabien0102: 1. 👍 
2. Except this little padding, we want all the feature of `Button` (hover, condensed…) this is why I prefer use the `Button` component instead of copying 20 lines of css 😉

src/Toggle/Toggle.tsx
```diff
@@ -31,29 +31,39 @@ const Container = styled("div")`
...
-      return `box-shadow: ${originalBoxShadow}, ${innerShadow};`
+      const innerShadow = condensed ? `0 0 5px 1px #B1B1B1 inset` : `0 0 7px 1px #B1B1B1 inset`
+      return `${originalBoxShadow}, ${innerShadow};`
     }
-    return ""
+    return originalBoxShadow
   }};
+
+  :hover {
+    background-color: ${props => darken(props.theme.color.white, 5)};
```
peterszerzo: The background shouldn't darken here if the button is selected.

src/Toggle/Toggle.tsx
```diff
@@ -0,0 +1,78 @@
...
+  /** Current value */
+  value: Value["value"]
+  /** Callback trigger on any changes */
+  onChange: (newValue: Value["value"]) => void
+  /** Condensed option */
+  condensed?: boolean
+}
+
+const Container = styled("div")`
+  > :not(:last-child) {
```
TejasQ: Why are we using `:not` rules here for first and last children given that container only _always_ contains 2 children at any given time? 🤔 

`:not(:last-child)` could be refactored as

```suggestion
  > :first-child {
```

since in this case, if it is not last, it is _definitely_ first.

src/Toggle/Toggle.tsx
fabien0102: This should be fixed, the hover and selected background color are the same

src/Toggle/Toggle.tsx
fabien0102: It's true, it was in prevision of the future (2+ cases) for now it's a bit over-enginering 😅 Since it works and the performance is not really impacted here (only 2 elements), I propose to keep this like this for now 👼

fabien0102: It's more about the intension -> if you are not the last one: remove the radius

src/Toggle/Toggle.tsx
```diff
@@ -0,0 +1,78 @@
...
+  font-size: ${props => props.theme.font.size.small}px;
+  font-weight: ${props => props.theme.font.weight.medium};
+  display: inline-flex;
+  align-items: center;
+  justify-content: center;
+  border-radius: ${props => props.theme.borderRadius}px;
+  border: 0;
+  cursor: pointer;
+  outline: none;
+  margin-right: ${props => props.theme.space.small}px;
```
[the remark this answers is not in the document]
fabien0102: It's needed if we have other elements around 😉

src/Toggle/Toggle.tsx
```diff
@@ -0,0 +1,78 @@
...
+  background-color: ${props => (props.selected ? darken(props.theme.color.white, 5) : props.theme.color.white)};
+  padding: 0 ${props => props.theme.space.element}px;
+  color: ${props => (props.selected ? props.theme.color.primary : props.theme.color.text.default)};
+  box-shadow: ${({ selected, condensed, theme: { color } }) => {
+    const originalBoxShadow = `0 0 0 1px ${color.border.disabled} inset`
+    if (selected) {
+      const innerShadow = condensed ? `0 0 5px 1px #B1B1B1 inset` : `0 0 7px 1px #B1B1B1 inset`
```
peterszerzo: Make the inner shadow color `#C8C8C8` instead

src/Toggle/Toggle.tsx
peterszerzo: `margin-left: -2px` is better because it won't affect the spacing between the toggle and neighboring elements to the right. Unnoticeable in most cases, but sometimes it can look sloppy.

While we're at it, I tried with `-2px` and I'm seeing a strong solid line in the middle that goes away with `-1px`. I think it's better to support default zoom in the best-looking way and allow for the tiny artifact when zoomed.

PR #874, opened 2018-12-17
**Fix:** outside click in `ContextMenu`

peterszerzo:
# Summary

Fix close-on-outside-click feature on `ContextMenu`, fixing the `TopbarSelect` bug where multiple can be open at the same time. The implementation removes the `InvisibleOverlay` component and manages document click listeners instead (the way `DatePicker` does it currently).

# To be tested

Me

Tester 1

Tester 2

### DIFF
--- a/src/ContextMenu/ContextMenu.tsx
+++ b/src/ContextMenu/ContextMenu.tsx
@@ -64,32 +64,37 @@ const MenuContainer = styled("div")<{
   minWidth: "fit-content",
 }))
 
-/**
- * Overlay to prevent mouse events when the context menu is open
- */
-const InvisibleOverlay = styled("div")(({ theme }) => ({
-  position: "fixed",
-  top: 0,
-  bottom: 0,
-  right: 0,
-  left: 0,
-  cursor: "default",
-  zIndex: theme.zIndex.selectOptions - 1,
-}))
-
 class ContextMenu extends React.Component<ContextMenuProps, Readonly<State>> {
   private menu: HTMLDivElement | null = null
+  private container: HTMLDivElement | null = null
 
-  private toggle = (e: React.MouseEvent<HTMLDivElement>) => {
-    e.stopPropagation()
-    this.setState(prevState => ({
-      isOpen: !prevState.isOpen,
+  public componentDidMount() {
+    document.addEventListener("click", this.onDocumentClick)
+  }
+
+  public componentWillUnmount() {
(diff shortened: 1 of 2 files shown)


src/ContextMenu/ContextMenu.tsx
```diff
@@ -64,32 +64,37 @@ const MenuContainer = styled("div")<{
...
-      isOpen: !prevState.isOpen,
+    this.setState(() => ({
+      isOpen: false,
     }))
   }
 
   private focusElement = () => {
-    if (this.menu && this.menu.querySelector('[tabindex="0"]')) {
-      setTimeout(() => (this.menu!.querySelector('[tabindex="0"]') as HTMLDivElement).focus())
+    if (this.menu && this.menu.querySelector(`[tabindex="0"]`)) {
```
TejasQ: @peterszerzo why did you replace quotes with backticks here? 🤔

micha-f:
@TejasQ that's correct!

src/ContextMenu/ContextMenu.tsx
peterszerzo: My linter was complaining.

peterszerzo:
@TejasQ fixed!

peterszerzo:
I don't understand, why should the click to the props & methods button not occur?

peterszerzo:
I understand now - in that case, I have concerns with the spec itself @micha-f. If the user clicks outside the context menu, in most cases they do so because they intended it, and expect whatever they're clicking to be clicked. If that's not what they intended, they can open the context menu again. I can see the appeal of an extra level of safety, but it's not worth having the UI seem frozen.

TejasQ:
@micha-f would love your feedback here. Personally, I disagree. I think the spec really helps guide the user by focusing them on a given dropdown.

micha-f:
Our discussion in the past was that click outside closes the dropdown - and does not trigger the underlying action (if you click on a button for example). In fact all hover effects and cursor changes (to a pointer) should also be disabled in the background page. This is what the invisible overlay does - correct?
I'd vote to keep it as it is a really simple and robust way to implement closing of dropdowns. We don't really want to force the user to click on the dropdown title or to look for and find some place safe to click on in the ui.

peterszerzo:
Makes sense, abandoning this PR then since it is based on removing the invisible overlay, which is indispensable to the direction we've agreed on.

PR #875, opened 2018-12-17
**Enhance:** context menu child sizes on window resize

peterszerzo:
# Summary

Updating menu widths on resize for `TopbarSelect` and `HeaderMenu`. Also fixing a bug where `window.addEventListener("resize")` got replaced with `document.body.addEventListener("resize")`, which doesn't fire normally.

# To be tested

Me

Tester 1

Tester 2

### DIFF
--- a/src/HeaderMenu/HeaderMenu.tsx
+++ b/src/HeaderMenu/HeaderMenu.tsx
@@ -1,4 +1,7 @@
+import { Cancelable } from "lodash"
+import debounce from "lodash/debounce"
 import * as React from "react"
+
 import ContextMenu, { ContextMenuProps } from "../ContextMenu/ContextMenu"
 import { DefaultProps } from "../types"
 import styled from "../utils/styled"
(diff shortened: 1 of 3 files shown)


peterszerzo:
@TejasQ isn't the `HeaderMenu`'s sized `250px` because it is intended to be that width? If so, why are we catering to cases where we forcefully change the width of the element?

The resize example is intended to guarantee sizes going out of sync in case things get squished (as an extra layer of protection if the UI breaks for an unrelated reason).

peterszerzo:
Just for context: `TopbarSelect` works with custom width settings, but even there, we don't typically set custom widths.

TejasQ:
@peterszerzo I think it would just be nice to demonstrate the behavior in the documentation... If you think we don't need an example, then let's just merge it. What do you think?

peterszerzo:
What I'm suggesting is that if we want `250px` components, then we should just say that that's it and we're not supporting anything else. I think it's an expectation throughout the library that if you hack around with `style` attributes and `styled()` variations, your results may vary.

(this reservation of mine came out of the implementation: sure, I should remove the `250px` limit, but should I really? Isn't that erasing the design intent here? The situation is different for `TopbarSelect`, but there everything is working)

TejasQ:
You're right. 😄

Issue #705, opened 2018-08-27
Spacings strategy


peterszerzo:
Currently, we have a few ways to specify spacings:
1. components set a margin on themselves, e.g. right margin on the `Button`.
2. parent sets margins on children, e.g. `Form`.
3. component sets a margin on itself if it receives the appropriate prop, e.g. `<Icon left />`.

This issue is meant as a discussion space to make this strategy more predictable and consistent.

There are advantages and drawbacks to each of these: the first approach involves fighting with overrides, the second avoids this but makes new components necessary with the sole responsibility to provide spacings, and the third introduces redundancy in the markup.

peterszerzo:
After an offline discussion, the decision is to not change the current spacings strategy. Feel free to revisit this issue in the future.

PR #876, opened 2018-12-18
**Feature:** Add Message component

fabien0102:
# Summary

image

# Related issue

https://trello.com/c/xlCBv6gh

# To be tested

Me

Tester 1

Tester 2


fabien0102:
[Add a message component](https://trello.com/c/xlCBv6gh/662-add-a-message-component)

src/Message/Message.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+export interface MessageProps {
+  title?: React.ReactNode
+  body: React.ReactNode
+  type: "warning" | "info" | "error"
+}
+
+const Container = styled("div")<{ type: MessageProps["type"] }>`
+  padding: ${props => props.theme.space.content}px;
+  margin-bottom: ${props => props.theme.space.big}px;
```
peterszerzo: What motivates this margin? Are we using `big` spacing consistently/in other components?

peterszerzo: Well, making the component work for that design screen will make us battle existing styling in other cases, making the component less usable and exposing spacing bugs. I would abstain from doing margins altogether, at least in new components.

fabien0102: I check all screens that we have, everything screens have a big margin after the error

peterszerzo: It is not the responsibility of this component to deal with these spacings, and we battle spacing-related bugs constantly as a result of setting them in the first place. I opened an issue on it in https://github.com/contiamo/operational-ui/issues/705 and feel free to add counterarguments, but my feeling right now is that we are blindly following design screens with implicit spacings intents, and as a result setting us up for a brittle system where we fight styles the same way we did with good old pre-emotion CSS.

TejasQ: @peterszerzo what's the solution you propose?

src/Message/Message.tsx
```diff
@@ -0,0 +1,56 @@
...
+    margin-bottom: ${props => props.theme.space.small}px;
+    color: ${props => (props.type === "error" ? props.theme.color.error : props.theme.color.text.default)};
+  }
+
+  p {
+    font-size: ${props => props.theme.font.size.small}px;
+    margin: 0;
+    color: ${props => props.theme.color.text.default};
+    word-break: break-word;
+    line-height: 1.46;
```
TejasQ: Where does this come from?

fabien0102: zeplin ;)

src/Message/Message.tsx
peterszerzo: Refrain from setting bottom margins. This is something we've discussed already offline and at that point we seemed to have an agreement an avoiding them (as in pushing margin settings to parents), and although it would make it inconsistent with other components at this point, we should at least not make new components with the same pattern.

TejasQ: I'd be okay if this component didn't manipulate margins.

fabien0102:
From an offline discussion with @micha-f -> let's merge this because we need it, and refactor later depending of the "spacing stratregy" discussion output

PR #878, opened 2018-12-18
**Feature:** Add sortable feature to Table

fabien0102:
# Summary

Add sortable feature to Table. Simply add `onSortClick` and `sortBy` in each column to handle any custom sorting logic.

b69a2ecb-4823-4354-9a17-9dea39667a64

# Related issue

https://trello.com/c/H6uFjzZN

# To be tested

Me

Tester 1

- [ ] I can sort a table
  

Tester 2

- [ ] I can sort a table

### DIFF
--- a/src/Table/Table.tsx
+++ b/src/Table/Table.tsx
@@ -28,7 +28,7 @@ export interface TableProps<T> extends DefaultProps {
 export interface Column<T> {
   heading: React.ReactNode
   cell: (dataEntry: T, index: number) => React.ReactNode
-  sortBy?: "asc" | "desc"
+  sortOrder?: "asc" | "desc"
   onSortClick?: (order: "asc" | "desc") => void
 }
 
@@ -157,10 +157,12 @@ function Table<T>({
               <Th
                 key={columnIndex}
                 sortable={Boolean(column.onSortClick)}
-                onClick={() => column.onSortClick && column.onSortClick(column.sortBy === "desc" ? "asc" : "desc")}
+                onClick={() => column.onSortClick && column.onSortClick(column.sortOrder === "desc" ? "asc" : "desc")}
               >
                 {column.heading}
-                {column.sortBy && <SortIcon size={16} name={column.sortBy === "desc" ? "CaretUp" : "CaretDown"} />}
+                {column.sortOrder && (
+                  <SortIcon size={16} name={column.sortOrder === "desc" ? "CaretUp" : "CaretDown"} />
+                )}
               </Th>
             ))}
             {Boolean(rowActions || (onRowClick && rowActionName)) && <Th key="infinity" />}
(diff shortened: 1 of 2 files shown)


fabien0102:
[Make tables sortable](https://trello.com/c/H6uFjzZN/657-make-tables-sortable)

src/Table/Table.tsx
```diff
@@ -28,6 +28,8 @@ export interface TableProps<T> extends DefaultProps {
 export interface Column<T> {
   heading: React.ReactNode
   cell: (dataEntry: T, index: number) => React.ReactNode
+  sortBy?: "asc" | "desc"
```
TejasQ: How about `sortOrder`?

peterszerzo:
Some design feedback:
* not having a caret on display when the table isn't sorted (`sortOrder === undefined`) hides the sortable functionality so I wouldn't know it's even possible. I would include a downward pointing caret with an opacity for this case and then switch to fully opaque once a sorting in specified.
* the caret belongs right next to the label I'd say, otherwise it looks like it belongs to the label to the right. This could work if there were a lot of padding between cells, but it's really not the case.
* carets could be a bit smaller.
* heading cells that are intended to support sorting should have a hover effect. We might want to have a `sortable` flag to toggle that behavior.

What do you think @micha-f?

TejasQ:
IMHO a design would be nice to have @micha-f

fabien0102:
Regarding the hover effect and caret displayed or not, I have benchmark some tables (as http://www.jquery-bootgrid.com/examples) it's quite standard to not have the caret displayed when it's not filtered, but agree, we can enhance the hover effect on sortable headers (it's just the cursor for now)

peterszerzo:
What other examples have you looked at?

TejasQ:
> https://datatables.net/ -> but I don't really like the style

looks a lot like @micha-f's mockup

micha-f:
image

@peterszerzo @fabien0102 what about this?

peterszerzo:
Looks very nice @micha-f - I think carets do a better job than chevrons.

micha-f:
@fabien0102 if you're ok with this, the specs are as follows:

- carets unsorted: #d8d8d8
- caret sorted: #primary
- caret sorted (up / down) needs to be aligned to vertical center of column titel 
- font of column title that is sorted: #747474
- table can only be sorted by one column (not sure we need to implement that in the table component though)

TejasQ:
`#primary` is not a valid hex code

troll

fabien0102:
@micha-f Any hover style?

micha-f:
@fabien0102 let's use #747474 for hover.

fabien0102:
@micha-f https://deploy-preview-878--operational-ui.netlify.com/#!/Table/15

PR #879, opened 2018-12-19
**Fix**: Push the confirm in front of modal

fabien0102:
# Summary

We have some case where we need to have a confirm modal inside a modal.

For example:

image

This PR take the shorter way to fix this specific issue: always put confirm in front of modal. It should be enought to fill our current needs, let's keep it simple for now.

# Related issue

# To be tested

Me

Tester 1

Tester 2


fabien0102:
[Support multi modals](https://trello.com/c/cDtJxbVQ/664-support-multi-modals)

src/Page/README.md
```diff
@@ -690,3 +690,63 @@ Modals are a more flexible alternative to the confirm box, that may contain anyt
...
+        }}
+      >
+        I want some bacon
+      </Button>
+    </Card>
+  )}
+</Page>
+```
+
+```
```
TejasQ: No!

PR #888, opened 2019-01-07
**Enhance:** foldable card sections

peterszerzo:
# Summary

This PR takes care of foldable behavior for card sections with a simple `onToggle` handler. Fold behavior is fully controlled in this implementation, anticipating the need to expand/collapse card sections when not clicking (e.g. dragging something into a different card section).

sections

# To be tested

Me

Tester 1

Tester 2

### DIFF
--- a/src/CardSection/CardSection.tsx
+++ b/src/CardSection/CardSection.tsx
@@ -27,6 +27,12 @@ export interface CardSectionProps extends DefaultProps, DragProps {
   collapsed?: boolean
   /** Toggle collapsed state */
   onToggle?: () => void
+  /** Fires when the toggle area is hovered */
+  onToggleMouseEnter?: () => void
+  /** Fires when the mouse leaves the hover area */
+  onToggleMouseLeave?: () => void
+  /** Force the toggle area to have hover styles whether they are hovered or not */
+  forceToggleHoverStyles?: boolean
 }
 
 export type DragAndDropFeedback = "validTarget" | "invalidTarget" | "dropping"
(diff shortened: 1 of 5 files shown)


peterszerzo:
@micha-f wondering about the toggle behavior here - if there are actions on the right side of the card section, we can't have the entire title bar be clickable, and I'm not sure if it's a good idea to have it be clickable in some cases and not in others. That's why I set it up with the little button initially. Thoughts on how to proceed here?

micha-f:
Do we have the case where we use actions on the right side and toggling in the same place?

peterszerzo:
We don't - I take it you're suggesting that the entire title should be clickable (+hover effect) when there is folding behavior?

TejasQ:
@peterszerzo how would the button look in [this example](http://operational-ui.netlify.com/#!/CardSection/5)?

peterszerzo:
@TejasQ since there isn't a specific use-case and it seems like a risky UX move, we can disallow/discourage both actions and toggling behavior active at the same time.

micha-f:
That makes sense. I don't think we want both in one card section.

TejasQ:
> we can disallow/discourage both actions and toggling behavior active at the same time

That could work. Alternatively, we could append the button and shift the actions to the left. Whatever you think makes the most sense would be great here.

micha-f:
Let's go with either or for now to keep things simple.

Can we reflect the design correctly here?

image

This screen is also in Zeplin

peterszerzo:
Absolutely, I'll make the tweaks!

micha-f:
Looks good @peterszerzo

Issue #892, opened 2019-01-16
Remove useless margin in Buttons


TejasQ:
```jsx
<Card action={<Button condensed>My Button yo</Button>} />
```

This thing has a right margin for a reason I cannot comprehend. Let's improve the `Button` style to _not_ have a right margin if there are no juxtaposed elements. 🙏

fabien0102:
Just to add some context

```jsx
<Card action={<>
  <Button condensed>I should have a right margin</Button>
  <Button condensed>I should not have a right margin</Button>
</>} />
```

TejasQ:
To add even more context,

```jsx
<Card action={<>
  <Button condensed>I should never have a right margin</Button>
</>} />
```

PR #893, opened 2019-01-18
**Fix**: Remove unnecessary button margin

TejasQ:
# Summary
This PR removes an unnecessary right margin on `Button`. (#892)

# To be tested

Me

Tester 1

Tester 2

### DIFF
--- a/src/Button/Button.tsx
+++ b/src/Button/Button.tsx
@@ -98,7 +98,10 @@ const containerStyles: Interpolation<
           },
         }
       : {}),
-    marginRight: theme.space.small,
+
+    "* + &": {
+      marginLeft: theme.space.small,
+    },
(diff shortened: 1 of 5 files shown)


src/Button/Button.tsx
```diff
@@ -98,7 +98,10 @@ const containerStyles: Interpolation<
           },
         }
       : {}),
-    marginRight: theme.space.small,
+
+    "* + &": {
```
stereobooster: Depends on how you separate responsibilities. Should element know what is outside of it and react according or should parent element control its behaviour?  I guess in this case it is not critical, but we should be careful with cascade.

TejasQ: Agree.

PR #890, opened 2019-01-08
**Feature**: Full width column in modal

fabien0102:
# Summary

Improve how we deal with spacing between cardColumn and Confirm to allow more complex setup.

It's published as `@operational/components@12.4.0-43-gd7a2df8e` and tested in labs-ui and pantheon-ui to be sure that everything follow as expected.

It's breaking for custom modal implementation, the migration to `ControlledModalContent` and `Actions` fix everything.

# Related issue

# To be tested

Me
- [x] No error or warning in the console on `localhost:6060`
- [x] This looks amazing in pantheon-ui
- [x] This looks amazing in labs-ui (PR incoming for the migration)

Tester 1

- [ ] Things look good on the demo.
  

Tester 2

- [ ] Things look good on the demo.

PR #894, opened 2019-01-18
**Feature:** Full Size Pages

TejasQ:
# Summary
This PR updates `Page` to allow truly FULL SIZE options (for Pantheon query editors).

# To be tested

Me

Tester 1

Tester 2

### DIFF
--- a/src/PageArea/PageArea.tsx
+++ b/src/PageArea/PageArea.tsx
@@ -13,16 +13,22 @@ import styled from "../utils/styled"
 export interface PageAreaProps extends DefaultProps {
   /** Name of the area */
   name?: "main" | "side"
-  fill_?: boolean
+  fill?: boolean
 }
 
-export const PageArea = styled("div")<PageAreaProps>(({ name, fill_ }) => ({
+const Container = styled("div")<{ name: PageAreaProps["name"]; fill_: boolean }>(({ name, fill_ }) => ({
   gridArea: name,
   height: fill_ ? "100%" : "auto",
 }))
 
+export const PageArea: React.SFC<PageAreaProps> = ({ name, fill, children }) => (
+  <Container name={name} fill_={Boolean(fill)}>
+    {children}
+  </Container>
+)
+
 PageArea.defaultProps = {
-  fill_: false,
+  fill: false,
 }
 
 export default PageArea
--- a/src/PageArea/PageArea.tsx
+++ b/src/PageArea/PageArea.tsx
@@ -13,10 +13,16 @@ import styled from "../utils/styled"
 export interface PageAreaProps extends DefaultProps {
   /** Name of the area */
   name?: "main" | "side"
+  fill_?: boolean
 }
 
-export const PageArea = styled("div")<PageAreaProps>(({ name }) => ({
+export const PageArea = styled("div")<PageAreaProps>(({ name, fill_ }) => ({
   gridArea: name,
+  height: fill_ ? "100%" : "auto",
 }))
+
+PageArea.defaultProps = {
+  fill_: false,
+}
 
 export default PageArea
--- a/src/PageContent/PageContent.tsx
+++ b/src/PageContent/PageContent.tsx
@@ -16,51 +16,59 @@ export interface PageContentProps extends DefaultProps {
   areas?: "main" | "main side" | "side main"
   /** Fill the entire width */
   fill?: boolean
+  /** Don't pad */
+  noPadding?: boolean
 }
 
 const sideSize = 280
 
-const StyledPageContent = styled("div")<{ areas?: PageContentProps["areas"]; fill_?: boolean }>(props => {
-  const gridTemplateColumns = {
-    main: props.fill_ ? "100%" : "auto",
-    "main side": `auto ${sideSize}px`,
-    "side main": `${sideSize}px auto`,
-  }[props.areas || "main"]
+const StyledPageContent = styled("div")<{ areas?: PageContentProps["areas"]; fill_: boolean; noPadding: boolean }>(
+  props => {
+    const gridTemplateColumns = {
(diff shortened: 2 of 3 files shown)


src/PageArea/PageArea.tsx
```diff
@@ -13,10 +13,16 @@ import styled from "../utils/styled"
 export interface PageAreaProps extends DefaultProps {
   /** Name of the area */
   name?: "main" | "side"
+  fill_?: boolean
 }
 
-export const PageArea = styled("div")<PageAreaProps>(({ name }) => ({
+export const PageArea = styled("div")<PageAreaProps>(({ name, fill_ }) => ({
   gridArea: name,
+  height: fill_ ? "100%" : "auto",
```
stereobooster: `fill_` does it mean non-public API? Traditionally people put underscore in the front, but I guess this is the matter of taste.

stereobooster: Also, this is kind of implementation detail leaking. I wonder if we should try to remove/hide it? For example, we can use `Context` and read `fill` property from parent container

TejasQ: I think `Context` is a bit overkill here, since we have a level of nesting that is not too complex: 

```
Page
|– PageContent
|–– PageArea
```

If this gets more complex, I think `Context` makes more sense.

`fill_` does not mean non-public API _only_. This convention exists because `fill` is a [_valid DOM element property_ that describes the `fill` property of a SVG](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/fill).

In this case, we are using it to set a property of a component whether it should _fill_ the screen or not. 

We do not want to cloud up the DOM with incorrect attributes, so this convention hints to React that this is not the actual DOM attribute but an internal property.

**The public API has no _ but is simply `fill`.**

Does this make sense?

stereobooster: Yes I got the idea, but to be fair I see this convention of using underscore to not collide with valid html properties for first time? What is the danger to collide? 

Developers will get confused by underscore - is it non-public API, is it deprecated, what is wrong with it, why it has underscore in the end, is it signal that there is something special about it, should I worry about it... People tend to judge things based on names, hence naming is hard

TejasQ: @stereobooster what would be your proposed solution?

Colliding with HTML properties triggers warnings in the console. We would like the library to be warning free.

IMO this convention is not a problem and we can explain if needed. I can even leave a comment if you like.

Do you have a better idea?

stereobooster: I don't get it, but `<PageContent fill={fill}>` doesn't trigger anything?

TejasQ: No, because [`PageContent` accepts `fill` as a prop but does not propagate it to an underlying DOM element](https://github.com/contiamo/operational-ui/blob/72e047382d3394347e68956a4506b5ac134b0f28/src/PageContent/PageContent.tsx#L88).

stereobooster: So this is issue of implementation. We can do something like this

```ts
export const PageArea = ({fill, children, ...rest}: PageAreaProps) => 
  (<div style={{gridArea: name, height: fill ? "100%" : "auto"}} {...rest}>children</div>);
```

fabien0102: I prefer the `_` prefix, it's simple 😉 (we can also use the emotion API to tell which props should not be forward, but it's painfullllllll as hell ^^)

stereobooster: It's just public API formation driven by implementation details. But ok, I don't think this is so big deal

src/PageContent/PageContent.tsx
```diff
@@ -70,14 +78,14 @@ const Container = styled("div")({
...
-const PageContent: React.SFC<PageContentProps> = ({ fill, children, ...props }) => {
+const PageContent: React.SFC<PageContentProps> = ({ fill, children, noPadding, ...props }) => {
   return (
     <Modal>
       {modal => (
         <Confirm>
           {confirm => (
             <Container>
-              <StyledPageContent {...props} fill_={fill}>
+              <StyledPageContent {...props} noPadding={Boolean(noPadding)} fill_={Boolean(fill)}>
```
stereobooster: Something disturbing me about `noPadding`, but I can't clearly understand what. 

When I think about positioning and abstractions around it I remember this talk https://www.deconstructconf.com/2017/kevin-lynagh-choosing-features.

src/PageArea/PageArea.tsx
fabien0102: True, I haven't seen that it was in the public API, let's clean this as here https://github.com/contiamo/operational-ui/blob/72e047382d3394347e68956a4506b5ac134b0f28/src/Page/Page.tsx#L34

Good catch @stereobooster

src/PageContent/PageContent.tsx
TejasQ: @stereobooster I'm following the convention introduced by @fabien0102 in #890.

https://github.com/contiamo/operational-ui/blob/ad9a58f39c9342e19e2354f863f5022483574b6c/src/CardColumn/CardColumn.tsx#L15

src/PageContent/PageContent.tsx
```diff
@@ -16,51 +16,59 @@ export interface PageContentProps extends DefaultProps {
...
-    gridTemplateAreas: `"${props.areas || "main"}"`,
-    gridGap: props.theme.space.element,
-    width: "100%",
-    height: "100%",
-    minWidth: 800,
-    maxWidth: props.fill_ ? "none" : 1150,
-    padding: `${props.theme.space.element}px`,
+    return {
+      gridTemplateColumns,
+      display: props.fill_ ? "block" : "grid",
```
fabien0102: NO!

PR #902, opened 2019-01-28
**Feature:** Add `width` to table columns

TejasQ:
# Summary
This PR adds an optional `width` property to Table's `Column` type, which will allow us to specify widths for columns and fix the IdP weird ugly issue.

It also adds support for fixed table layouts in case our users want perf benefits.

# To be tested

Me

Tester 1

Tester 2

### DIFF
--- a/src/Table/Table.tsx
+++ b/src/Table/Table.tsx
@@ -23,6 +23,8 @@ export interface TableProps<T> extends DefaultProps {
   iconColor?: (dataEntry: T) => string
   /** Remove the header? */
   headless?: boolean
+  /** Fixed Layout for performance */
+  fixedLayout?: boolean
 }
 
 export interface Column<T> {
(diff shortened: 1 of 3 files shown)


src/Table/Table.tsx
```diff
@@ -88,13 +92,17 @@ const ThContent = styled("span")<{ sorted?: boolean }>`
   ${props => props.sorted && `color: ${props.theme.color.text.light};`};
 `
 
-const Td = styled("td")(({ theme }) => ({
+const Td = styled("td")<{ cellWidth?: Column<any>["width"] }>(({ theme, cellWidth }) => ({
   verticalAlign: "middle",
   borderBottom: `1px solid ${theme.color.separators.default}`,
   color: theme.color.text.default,
+  wordBreak: "break-all",
```
stereobooster: should we do this only when there is `cellWidth`?

Issue #910, opened 2019-02-04
[Umbrella] Migrate to Hooks


TejasQ:
[React Hooks has been merged](https://github.com/facebook/react/pull/14679) and so it's time to rewrite our entire library in hooks to:

1. Be cool, but also
1. To more effectively reuse stateful logic between our components, and most importantly, to
1. Be **extra `1337`**.

There are 18 components that use class components and need to be migrated. Let this issue serve as an umbrella to port them over. **Open separate issues if needed** and update the checklist to include a link to the issue, instead of just plain text.

## Components to Migrate

- [ ] [Autocomplete](https://github.com/contiamo/operational-ui/blob/master/src/Autocomplete/Autocomplete.tsx)
- [ ] [ContextMenu](https://github.com/contiamo/operational-ui/blob/master/src/ContextMenu/ContextMenu.tsx)
- [ ] [DatePicker](https://github.com/contiamo/operational-ui/blob/master/src/DatePicker/DatePicker.tsx)
- [ ] [Debug](https://github.com/contiamo/operational-ui/blob/master/src/Debug/Debug.tsx)
- [ ] [Foldable](https://github.com/contiamo/operational-ui/blob/master/src/Foldable/Foldable.tsx)
- [ ] [OperationalUI](https://github.com/contiamo/operational-ui/blob/master/src/OperationalUI/OperationalUI.tsx)
- [ ] [Page](https://github.com/contiamo/operational-ui/blob/master/src/Page/Page.tsx)
- [ ] [Select](https://github.com/contiamo/operational-ui/blob/master/src/Select/Select.tsx)
- [ ] [Splash](https://github.com/contiamo/operational-ui/blob/master/src/Splash/Splash.tsx)
- [ ] [Textarea](https://github.com/contiamo/operational-ui/blob/master/src/Textarea/Textarea.tsx)
- [ ] [Tooltip](https://github.com/contiamo/operational-ui/blob/master/src/Tooltip/Tooltip.tsx)
- [ ] [TopbarSelect](https://github.com/contiamo/operational-ui/blob/master/src/TopbarSelect/TopbarSelect.tsx)
- [ ] [Tree](https://github.com/contiamo/operational-ui/blob/master/src/Tree/Tree.tsx)

### Internal Components
- [ ] [Confirm](https://github.com/contiamo/operational-ui/blob/master/src/Internals/Confirm.tsx)
[...]

Issue #908, opened 2019-02-04
Any plans on moving to hooks?


adeelibr:
With hooks coming out of alpha version soon, is there a road map in place to move/migrate all Class components to hooks. With hooks everything is a function.

fabien0102:
We will probably move to hooks here 😉 We already use them in production 👼 

But regarding the strategy, I don't really see the point to rewrite everything in hooks style (since the library API is the same) However it can be nice to have a hook version of our context component (to avoid nestead render props 😄 )

So let's begin by `OperationalContext`, rest are just for library confort. Also I will prefer to keep render props and hooks style, so don't break everything 🤠 

@adeelibr Are you interest to take this feature?
@TejasQ What do you think about this strategy?

TejasQ:
@fabien0102 I like your idea. Let's queue this for v14 as it will be a great breaking change. Step 1 is to update the `peerDependencies` entry for React to be `^16.8`. 😄

TejasQ:
@adeelibr, to add to @fabien0102's great answer to your question, here's a nice todo list if you're interested in contributing: https://github.com/contiamo/operational-ui/issues/908

I'd suggest discussing further there. Please close this issue if you agree, or feel free to carry on the conversation here.

adeelibr:
> @adeelibr Are you interest to take this feature?

I would love to 👍 @fabien0102 

> @adeelibr, to add to @fabien0102's great answer to your question, here's a nice todo list if you're interested in contributing: #908 I'd suggest discussing further there.

This is issue #908 😅 @TejasQ

TejasQ:
Oops, I meant #910 😂

adeelibr:
Great! I am closing this issue in favor of #910, I'd like to start of working with `OperationalContext` initially.

PR #911, opened 2019-02-04
 **Feature:** OperationalContext hook useOperationalContext

adeelibr:
# Summary

Keeping the current `render props` implementation in tact, add another function called `useOperationalContext` which returns the current `context` of `OperationalContext`.

Uses Hydra pattern https://americanexpress.io/hydra/

### usage

```js
import { useOperationalContext } from "../src/OperationalContext/OperationalContext"

const Foo = () => {
  const ctx = useOperationalContext();
  return <p>Window width is {ctx.windowSize.width}</p>
};
```

P.S: I have not commited `package.json` with the updated `react` & `react-dom` version, should I for this PR?

CC: @fabien0102 @TejasQ 

# Related issue

https://github.com/contiamo/operational-ui/issues/908
https://github.com/contiamo/operational-ui/issues/910

# To be tested

Me
- [x ] No error or warning in the console on `localhost:6060`


TejasQ:
> P.S: I have not commited package.json with the updated react & react-dom version, should I for this PR?

@adeelibr 💯% yes

src/OperationalContext/OperationalContext.init.tsx
```diff
@@ -39,8 +39,12 @@ const defaultContext: Context = {
   setLoading: (_: boolean) => void 0,
 }
 
-const { Provider, Consumer: OperationalContext } = React.createContext(defaultContext)
+const OperationalContext = React.createContext(defaultContext)
 
-export default OperationalContext
+const { Provider, Consumer } = OperationalContext
+
+export const useOperationalContext = () => React.useContext(OperationalContext)
```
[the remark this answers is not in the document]
TejasQ: @adeelibr that's weird. Let's take a look... 👀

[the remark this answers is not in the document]
adeelibr: Great! regenerating lock file helped.

TejasQ:
@adeelibr let's add an example to the README so we see it in the playground. 😉

adeelibr:
Can you kindly review again. @TejasQ

src/OperationalContext/README.md
```diff
@@ -14,6 +14,19 @@ Child components can get access to window dimensions as follows:
...
 
+### OperationalContext Hook
+
+You can get the value of `OperationalContext` in the shape of context.
+
+```jsx
+const { useOperationalContext } = require("./OperationalContext")
+const ILookPretty = () => {
+  const ctx = useOperationalContext()
+  return <p>{`The viewport is ${ctx.windowSize.width} pixels wide and ${ctx.windowSize.height} tall.`}</p>
```
TejasQ:
```suggestion
  return <p>The viewport is {ctx.windowSize.width} pixels wide and {ctx.windowSize.height} tall.</p>
```

adeelibr: I have made the changes 👍

adeelibr:
I made one small change in `OperationalContext.tsx`

```diff
+  const OperationalContext: React.SFC<Props> = props => {
+    const ctx = useOperationalContext()
+    return <React.Fragment>{props.children({ ...ctx })}</React.Fragment>
+ }
-  const OperationalContext: React.SFC<Props> = props => (
-     <OperationalContextOriginal>{props.children}</OperationalContextOriginal>
-  );
```

This was based on https://americanexpress.io/hydra/ Which uses context under the hood, to support render-props. This way we don't have to manage both of them at the same time.

It is working fine at my end, on the local dev-server. I don't know why the deployment is failing here.

I tried running this in the `dev-server`, it works fine.

```jsx
<OperationalContext>
            {operationalContext => (
              <p>{`The viewport is ${operationalContext.windowSize.width} pixels wide and ${
                operationalContext.windowSize.height
                } tall.`}</p>
            )}
</OperationalContext>
```

TejasQ:
@adeelibr you got it. @fabien0102 let's review this. 🚀

src/OperationalContext/OperationalContext.tsx
```diff
@@ -16,10 +10,11 @@ export interface Props {
  * This component simply wraps OperationalContext in order to allow styleguidist to pick up on
  * it and display it in the documentation page.
  */
-const OperationalContext: React.SFC<Props> = props => (
-  <OperationalContextOriginal>{props.children}</OperationalContextOriginal>
-)
+const OperationalContext: React.SFC<Props> = props => {
+  const ctx = useOperationalContext()
+  return <React.Fragment>{props.children({ ...ctx })}</React.Fragment>
```
TejasQ: Do we need a fragment here? 🤔

adeelibr: Yes, It gives a typing error otherwise for directly returning children.

TejasQ: Have you tried 

```suggestion
  return <>{props.children({ ...ctx })}</>
```

🤔

dev-server/index.tsx
```diff
@@ -6,7 +6,7 @@
  */
```
TejasQ: ? We probably can delete this file no?

adeelibr: Sorry did that by mistake, reverting it.

adeelibr: Can you kindly review again. 😅 @TejasQ

TejasQ:
@adeelibr @fabien0102 one final thing that concerns me, is that if this component is used in a class component, the hook will throw for **no fault of the user** and be a very surprising breakage iiuc. Can we protect against this somehow?

fabien0102:
@TejasQ Can you add an example for context? I don't really when what throw 😄

adeelibr:
The new linters rule will specify that hooks should not be used in class components. I don't see a work around for this I am afraid.

TejasQ:
@fabien0102 @adeelibr 

here's an example

```jsx
// Legacy code

class MySomething extends React.Component {
  render() {
    <OperationalContext>
      {({ pushState }) => <a onClick={() => pushState('/my-chicken-loves-me')}>HELLO 🐔❤️</a>}
    </OperationalContext>
  }
}
```

The new `OperationalContext` will throw

```Hooks can only be used in React function components```

and the whole app will revert to an error boundary or unmount.

adeelibr:
> The new `OperationalContext` will throw

Works fine at my end, doesn't throw an error & it won't throw an error. Because we are using that hook inside our render prop component. The way people access it using `render-props` it won't cause an error.

TejasQ:
@adeelibr I love you let's add more hooks WOOOOOO

PR #917, opened 2019-02-11
**Feature: ** Add custom padding prop to PageContent

fabien0102:
# Summary

Add `padding` props to `PageContent` to handle custom configuration.

It's a bit more flexible than before but still restrictive by design -> we only allow theme value as `padding`.

# Related issue

# To be tested

Me

Tester 1

Tester 2

### DIFF
--- a/src/PageContent/PageContent.tsx
+++ b/src/PageContent/PageContent.tsx
@@ -10,7 +10,7 @@ export interface ModalConfirmContext {
   confirm: <T = undefined>(confirmOptions: ConfirmOptions<T>) => void
 }
 
-const isRenderChild = (
+const isChildFunction = (
   children: PageContentProps["children"],
 ): children is ((modalConfirmContext: ModalConfirmContext) => React.ReactNode) => typeof children === "function"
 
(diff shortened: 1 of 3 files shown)


src/PageContent/PageContent.tsx
```diff
@@ -77,16 +96,15 @@ const Container = styled("div")({
...
   return (
     <Modal>
       {modal => (
         <Confirm>
           {confirm => (
             <Container>
-              <StyledPageContent {...props} noPadding={Boolean(noPadding)} fill_={Boolean(fill)}>
-                {typeof children === "function" ? children({ confirm, modal }) : children}
+              <StyledPageContent {...props}>
+                {isRenderChild(children) ? children({ confirm, modal }) : children}
```
TejasQ: I suggest this

```suggestion
                {isChildFunction(children) ? children({ confirm, modal }) : children}
```

src/PageContent/PageContent.tsx
```diff
@@ -2,73 +2,92 @@ import * as React from "react"
...
   areas?: "main" | "main side" | "side main"
   /** Fill the entire width */
   fill?: boolean
+}
+
+export interface PageContentNoPaddingProps extends BasePageContentProps {
   /** Don't pad */
   noPadding?: boolean
+  /** Add a padding */
+  padding?: never
```
stereobooster: Note: this is done for backward compatibility, otherwise it can be one param

PR #918, opened 2019-02-11
**Feature:** useSyncToPathState hook

fabien0102:
# Summary

Hooks are official, it's time to introduce some utility hooks in our library!

This `useSyncToPathState` hook give you a state that is connected to the url. It works almost as `useState` expect that it take a "name" (to reference the search param) and an "encoder" (to validate/encode the value from the url)

[Edit q5yvm3r24](https://codesandbox.io/s/q5yvm3r24)

# To be tested

Me

Tester 1

Tester 2


src/hooks/useSyncToPathState/README.md
```diff
@@ -0,0 +1,34 @@
+It's quite conveniant to be able to copy/past a link to somebody to show him a particular page of your application.
```
TejasQ:
```suggestion
It's quite convenient to be able to copy and paste a link to somebody to show them a particular page of your application.
```

src/hooks/useSyncToPathState/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+It's even more conveniant if this link have all the state of the current page! This is what this hook is made for.
```
TejasQ:
```suggestion
It's even more convenient if this link has the state of the current page! This is what this hook is made for.
```

src/hooks/useSyncToPathState/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+Simply use `useSyncToPathState` instead of the classic `useState` and your props is automatically sync the url.
```
TejasQ:
```suggestion
Simply use `useSyncToPathState` instead of the classic `useState` and your props are automatically synchronized to the URL.
```

src/hooks/useSyncToPathState/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+Because the url is not really the safest way to provide a state to an application, you need to provide an `encoder` function that validate and parse the input from the url to your state. You need to return `undefined` if the input is not valid.
```
TejasQ:
```suggestion
Because the URL is not really the safest way to provide state to an application, you need to provide an `encoder` function that validate and parse the input from the URL to your state. You need to return `undefined` if the input is not valid.
```

> You need to return `undefined` if the input is not valid.

This is confusing. Why?

src/hooks/useURLState/useURLState.ts
```diff
@@ -0,0 +1,41 @@
...
+ * Create a state that is sync with url search param.
+ *
+ * @param name Name of your state
+ * @param decoder Validate and decode the value from the url (you must return undefined if the value is not valid)
+ * @param search Search string from the url
+ * @param replaceState Replace state
+ * @param getPathname Get the current location pathname
+ * @param getHash Get the current location hash
+ */
+export const useURLState = <T>(
```
stereobooster: More than two params is hard to remember (at least for me). Maybe we can do this

```ts
export type StringValue =
    string
  | StringObject
  | StringArray;
export interface StringObject extends Record<string, StringValue> {}
export interface StringArray extends Array<StringValue> {}

const defaultOptions = {
  encode: (urlParam: StringValue | undefined) => urlParam,
  decode: (val: StringValue | undefined) => val,
  getSearchParams = () => qs.parse(window.location.search.replace("?", "")) || {},
  replaceState: History["replaceState"] = window.history.replaceState.bind(window.history),
  getPathname = () => window.location.pathname,
  getHash = () => window.location.hash,
};

const useURLState = <T extends StringValue>(name: string, initialValue: t, options: Options<T> = {}): [T, Dispatch<SetStateAction<T>>] => {
  options = {defaultOptions, ...options};
}
```

By default, if you do not provide options, you can use only `StringValue`, if you want `number` or `Date` you need to provide `encode`, `decode` functions which will convert from(to) JS format to(from) serialised string

src/hooks/useURLState/README.md
```diff
@@ -0,0 +1,34 @@
+It's quite convenient to be able to copy and paste a link to somebody to show them a particular page of your application.
+
+It's even more convenient if this link has the state of the current page! This is what this hook is made for.
+
+Simply use `useURLState` instead of the classic `useState` and your props are automatically synchronized to the URL.
+
+Because the URL is not really the safest way to provide state to an application, you need to provide an `decoder` function that validate and parse the input from the URL to your state. You need to return `undefined` if the input is not valid.
```
stereobooster: You need to return `undefined` if the input is not valid and `useURLState` will use default value instead.

src/hooks/useURLState/README.md
```diff
@@ -0,0 +1,34 @@
+It's quite convenient to be able to copy and paste a link to somebody to show them a particular page of your application.
+
+It's even more convenient if this link has the state of the current page! This is what this hook is made for.
+
+Simply use `useURLState` instead of the classic `useState` and your props are automatically synchronized to the URL.
```
stereobooster:
```suggestion
Disclaimer: when we say synchronized, we don't mean two way binding. It will read initial state from URL on component mount, but after internal state will be source of truth. Which is ok most of the time, because we use `replaceState` and there is no way from user point of view to change URL without triggering page reload. The only edge case to be aware here is when you use `useURLState` with `pushState`, in this case you will need to make sure that component will re-mount on push state, for example, if you use `react-router` use component with `useURLState` inside router. 
```

@TejasQ please check wording ^

src/hooks/useURLState/useURLState.ts
fabien0102: The `StringValue` doesn't work because the `initialValue` always override the `T`

src/hooks/useURLState/README.md
```diff
@@ -0,0 +1,34 @@
...
+It's even more convenient if this link has the state of the current page! This is what this hook is made for.
+
+Disclaimer: when we say synchronized, we don't mean two way binding. It will read initial state from URL on component mount, but after internal state will be source of truth. Which is ok most of the time, because we use `replaceState` and there is no way from user point of view to change URL without triggering page reload. The only edge case to be aware here is when you use `useURLState` with `pushState`, in this case you will need to make sure that component will re-mount on push state, for example, if you use `react-router` use component with `useURLState` inside router. 
+
+Because the URL is not really the safest way to provide state to an application, you need to provide an `decoder` function that validate and parse the input from the URL to your state. You need to return `undefined` if the input is not valid and `useURLState` will use default value instead.
+
+## Usage
+
+```jsx
+const countDecoder = i => {
```
stereobooster: `const decode = i => Number.isNaN(+i) ? undefined : +i;`?

src/hooks/useURLState/useURLState.test.tsx
```diff
@@ -0,0 +1,268 @@
...
+
+  it("should not erase previous search params", async () => {
+    const decoder = () => undefined
+    const getSearchParams = () => ({ please: "keepme" })
+    const replaceState = jest.fn()
+    const getPathname = () => "http://myapp.com"
+    const getHash = () => ""
+
+    const Counter = () => {
+      const [count, setCount] = useURLState<number>("count", 0, decoder, {
```
stereobooster: Do we need `<number>` here? Can't TS infer type from default value?

src/hooks/useURLState/useURLState.ts
```diff
@@ -0,0 +1,47 @@
+import qs from "qs"
+import { Dispatch, SetStateAction, useEffect, useState } from "react"
+
+/**
+ * Bunch of method that depends on `window`
+ *
+ * This is mostly for testing purpose but can also be used for SSR.
+ */
+const options = {
+  getSearchParams: () => qs.parse(window.location.search.replace("?", "")) || {},
```
stereobooster: `getSearchParams` injectable, but you can't really switch logic entirely, because this code assumes TS anyway.

```ts
  useEffect(() => {
    const params = getSearchParams()
    params[name] = value
    const search = `?${qs.stringify(params)}`
```

Maybe use `getSearch` or `getSearchString` instead?

src/hooks/useURLState/useURLState.test.tsx
fabien0102: Styleguidist complain… he infer `DispatchState<0>` instead of `DispatchState<number>` 🤦‍♂️

src/hooks/useURLState/README.md
TejasQ:
```
Disclaimer: when we say synchronized, we don't mean two way binding. It will read initial state from URL on component mount, but from that point on, the component's **internal state** will be the source of truth. 

This is ok most of the time, because we use `replaceState` and there is no way from a user's point of view to change the URL without triggering page reload. 

The only edge case to be aware of here is when you use `useURLState` with `pushState`. In this case, you will need to make sure that your component will re-mount on push state, for example, if you use `react-router` use component with `useURLState` inside router.
```

> The only edge case to be aware of here is when you use `useURLState` with `pushState`. In this case, you will need to make sure that your component will re-mount on push state, for example, if you use `react-router` use component with `useURLState` inside router.

This makes no sense.

src/hooks/useURLState/README.md
```diff
@@ -0,0 +1,29 @@
+It's quite convenient to be able to copy and paste a link to somebody to show them a particular page of your application.
+
+It's even more convenient if this link has the state of the current page! This is what this hook is made for.
+
+Disclaimer: when we say synchronized, we don't mean two way binding. It will read initial state from URL on component mount, but after internal state will be source of truth. Which is ok most of the time, because we use `replaceState` and there is no way from user point of view to change URL without triggering page reload. The only edge case to be aware here is when you use `useURLState` with `pushState`, in this case you will need to make sure that component will re-mount on push state, for example, if you use `react-router` use component with `useURLState` inside router.
```
TejasQ: > in this case you will need to make sure that component will re-mount on push state, for example, if you use `react-router` use component with `useURLState` inside router.

This needs to be reworded. You need to make sure your component will remount?

src/hooks/useURLState/README.md
```diff
@@ -0,0 +1,29 @@
+It's quite convenient to be able to copy and paste a link to somebody to show them a particular page of your application.
+
+It's even more convenient if this link has the state of the current page! This is what this hook is made for.
+
+Disclaimer: when we say synchronized, we don't mean two way binding. It will read initial state from URL on component mount, but after internal state will be source of truth. Which is ok most of the time, because we use `replaceState` and there is no way from user point of view to change URL without triggering page reload. The only edge case to be aware here is when you use `useURLState` with `pushState`, in this case you will need to make sure that component will re-mount on push state, for example, if you use `react-router` use component with `useURLState` inside router. 
+
+Because the URL is not really the safest way to provide state to an application, you need to provide an `decoder` function that validate and parse the input from the URL to your state. You need to return `undefined` if the input is not valid and `useURLState` will use default value instead.
```
TejasQ: > You need to return `undefined` if the input is not valid and `useURLState` will use default value instead.

Why? This is confusing.

fabien0102: Because it's easy -> return no value if no value match.

What do you propose?

src/hooks/useURLState/README.md
fabien0102: Only if you want two way binding, but it's not the purpose of this component 😉 The idea is to warn  about this bad usage, that it.

TejasQ:
```suggestion
Disclaimer: when we say synchronized, we don't mean two way binding. It will read initial state from URL on component mount, but after internal state will be source of truth. Which is ok most of the time, because we use `replaceState` and there is no way from user point of view to change URL without triggering page reload. The only edge case to be aware here is when you use `useURLState` with `pushState`, in this case you will need to make sure that component lives inside a valid router context.
```

src/hooks/useURLState/README.md
TejasQ: I propose removing the whole section about `undefined` and letting the users return fallbacks themselves.

```suggestion
Because the URL is not really the safest way to provide state to an application, you are encouraged to provide a `decoder` function that will validate and parse the input from the URL to your state. If the input from the URL is invalid, simply return a reasonable fallback.
```

src/hooks/useURLState/useURLState.ts
```diff
@@ -0,0 +1,54 @@
...
+ *
+ * @param search
+ */
+const getSearchParams = (search: string) => qs.parse(search.replace("?", "")) || {}
+
+/**
+ * Create a state that is sync with url search param.
+ *
+ * @param name Name of your state
+ * @param decoder Validate and decode the value from the url (you must return undefined if the value is not valid)
```
stereobooster:
```suggestion
 * @param decoder Validate and decode the value from the url 
```

src/hooks/useURLState/README.md
```diff
@@ -0,0 +1,29 @@
+It's quite convenient to be able to copy and paste a link to somebody to show them a particular page of your application.
+
+It's even more convenient if this link has the state of the current page! This is what this hook is made for.
+
+Disclaimer: when we say synchronized, we don't mean two way binding. It will read initial state from URL on component mount, but after internal state will be source of truth. Which is ok most of the time, because we use `replaceState` and there is no way from user point of view to change URL without triggering page reload. The only edge case to be aware here is when you use `useURLState` with `pushState`, in this case you will need to make sure that component lives inside a valid router context.
```
TejasQ:
```suggestion
Disclaimer: when we say synchronized, we don't mean two way binding. It will read the initial state from the URL on component mount, but afterwards, this hook's internal state will be source of truth. 

This is ok most of the time because we use `replaceState` and there is no way from user's point of view to change the URL without triggering a page reload. 

The only edge case to be aware of here is when you use `useURLState` with `pushState`. In this case, you will need to make sure that this hook lives inside a valid router context.
```

src/hooks/useURLState/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+Because the URL is not really the safest way to provide state to an application, you are encouraged to provide a `decoder` function that will validate and parse the input from the URL to your state. If the input from the URL is invalid, simply return a reasonable fallback.
```
TejasQ:
```suggestion
Because the URL is not really the safest way to provide state to an application, you are encouraged to provide a `decoder` function that will validate and parse the input from the URL to your state. If the input from the URL is invalid, simply return a reasonable fallback or `undefined` to fall back to the default value.
```

PR #920, opened 2019-02-13
**Fix:** Foldable hook

TejasQ:
This PR fixes a hook with `useEffect` that keeps adding and removing event listeners on each render.

screenshot

To fix, we only really care about one value: `isTogglerHovered`. If that updates, we need to update the event listener.

cc @adeelibr


adeelibr:
Understood your approach 💯 👍 @TejasQ

PR #921, opened 2019-02-14
**Breaking:** Fix operational context

fabien0102:
# Summary

image

- Make operational context component class compliante
- Remove windowSize on context (huge performance leak)
- Add some todo for the future (we need to refactor this tooltip component)

# Related issue

# To be tested

Me

Tester 1

Tester 2


src/OperationalContext/README.md
```diff
@@ -1,19 +1,5 @@
 The `OperationalContext` component provides utility methods and data that can be used inside `OperationalUI`:
 
-### Window size
```
stereobooster: We removed size from context, because it triggers re-render root component (`<OperationalUI>`) on every resize

src/Tooltip/Tooltip.tsx
```diff
@@ -146,7 +145,10 @@ class Tooltip extends React.Component<TooltipProps, State> {
     return position
   }
 
-  public getDisplayPosition(windowSize: { width: number; height: number }) {
+  public getDisplayPosition() {
+    // TODO: Refactor with `useWindowSize` hook
```
stereobooster: Let's pair and finish implementation of this component

src/Tooltip/Tooltip.tsx
```diff
@@ -84,131 +84,114 @@ export interface State {
...
+      const bbRect = containerNode.current.getBoundingClientRect()
+      setState({
+        bbTop: bbRect.top,
+        bbBottom: bbRect.bottom,
+        bbLeft: bbRect.left,
+        bbRight: bbRect.right,
+        singleLineTextWidth: bbOffScreen.width,
+      })
+    },
+    [containerNode.current, offScreenWidthTestNode.current],
```
stereobooster: can I use refs here 🤔

stereobooster:
Current problem: `Warning: src/Tooltip/Tooltip.tsx matches a pattern defined in “components” or “sections” options in your style guide config but doesn’t export a component.`

src/hooks/useWindowSize/useWindowSize.ts
```diff
@@ -0,0 +1,23 @@
+import { useEffect, useState } from "react"
+
+const useWindowSize = () => {
```
stereobooster: We don't have readme and tests for this hook

src/OperationalContext/OperationalContext.tsx
```diff
@@ -1,20 +1,38 @@
...
 
-export default OperationalContext
+const OperationalContext = React.createContext(defaultContext)
+
+export const { Provider, Consumer } = OperationalContext
+
+export const useOperationalContext = () => React.useContext(OperationalContext)
 
-export { Context, WindowSize, IMessage, MessageType, useOperationalContext }
+export default Consumer
```
TejasQ: Maybe we should still export the types?

src/hooks/useWindowSize/useWindowSize.ts
TejasQ: Add them.

src/hooks/useWindowSize/useWindowSize.ts
```diff
@@ -0,0 +1,23 @@
...
+  useEffect(() => {
+    const handler = () => {
+      setWindowSize({
+        width: window.innerWidth,
+        height: window.innerHeight,
+      })
+    }
+    window.addEventListener("resize", handler)
+    return () => window.removeEventListener("resize", handler)
+  }, [])
```
TejasQ: Are you sure about the empty array? Shouldn’t we put window.innerWidth and height in here?

stereobooster: The idea is to subscribe on mount and unsubscribe on unmount, that is why empty array. We want to trigger effect only once

TejasQ: In that case, `useWindowSize` is a lie because window size changes, but the change is not reflected in the return value of the hook dynamically.

TejasQ: In that case, `useWindowSize` is a lie because window size changes, but the change is not reflected in the return value of the hook dynamically.

micha-f: Looks correct to me (but I have no deep understanding of hooks). This effect doesn't need to re-run as it doesn't depend on props or state. Thus `[]` makes sense to me.

@stereobooster You can think about returning 
```
{ 
  width: window.innerWidth,
  height: window.innerHeight
}
```

from a function instead of repeating it.

TejasQ:
image

Do either of you test locally?

src/hooks/useWindowSize/useWindowSize.ts
TejasQ: You're right. My mistake.

PR #922, opened 2019-02-14
**Feature:** GO HARD on a11y

TejasQ:
# Summary
This PR fixes a number of issues with Inputs:
- Odd behavior when used without/with labels
- Fix spacing when an Input has an error AND a label
- Fix cursors on Input buttons
- Refactor legacy `styled` import
- Tooltips now have `id`s for screen readers
- Make all `Input`s accessible: with `aria` properties and correct label handling
- Update all examples
- Refactor `Autocomplete` to use hooks and support IDs

# To Do
- [x] Make the same changes to `Textarea`

# Open separate PRs for
- [ ] Make the same changes to `Select`
- [ ] Make the same changes to `Checkbox`
- [ ] Make the same changes to `Switch`
- [ ] Make the same changes to `Toggle`

# To be tested

Me

Tester 1

Tester 2


src/Autocomplete/Autocomplete.tsx
```diff
@@ -70,52 +76,42 @@ const Container = styled(ContextMenu)<{ fullWidth: boolean }>`
...
+  resultIcon,
+  loading,
+  noResultsMessage = "No Results Found",
+  onResultClick,
+  selectedResult,
+  children,
+  ...inputProps
+}: AutocompleteProps<T>) {
+  const [isContextMenuOpen, setIsContextMenuOpen] = React.useState(false)
+  const autocompleteId = React.useRef(id || nanoid())
```
stereobooster:
```ts
const useUniqueId = () => {
  const id = React.useRef<string>(null);
  if (!id.current) id.current = nanoid();
  return id.current;
}
```

TejasQ: UFUFUFFUFUFUFU

src/Tooltip/Tooltip.tsx
```diff
@@ -179,13 +180,14 @@ const Tooltip: React.SFC<TooltipProps> = props => {
         <p>{props.children}</p>
       </Container>
       <Container
+        aria-role="tooltip"
```
stereobooster:
```suggestion
        role="tooltip"
```

src/useUniqueId/index.ts
```diff
@@ -0,0 +1,11 @@
+import nanoid from "nanoid"
+import { useRef } from "react"
+
+export const useUniqueId = (defaultId: string | null = null) => {
```
stereobooster:
```ts
export const useUniqueId = (defaultId: string | null = null) => React.useState(nanoid)[0] as string;
```

source https://dev.to/alutom/comment/984j

stereobooster:
Does snapshot tests fail because of this https://github.com/facebook/jest/issues/7094?

src/Input/Input.Button.tsx
```diff
@@ -35,6 +35,9 @@ const Button = styled("div")`
   align-items: center;
   justify-content: center;
   cursor: pointer;
+  * {
```
stereobooster: why?

src/Input/Input.Field.tsx
```diff
@@ -128,39 +127,41 @@ const InputField: React.SFC<InputProps> = ({
...
-        {clear && value && (
-          <ClearButton onClick={clear}>
-            <Icon color="color.text.lightest" name="No" />
-          </ClearButton>
-        )}
-      </Container>
+    <Container fullWidth={fullWidth} withLabel={Boolean(label)}>
+      {shouldShowIconButton && renderButton()}
+      <Field
+        aria-labelledby={label ? `input-label-${id}` : undefined}
```
stereobooster: `ariaLabelledby`?

src/utils/mixins.ts
```diff
@@ -55,12 +55,15 @@ export const FormFieldControl = styled("div")(({ theme }) => ({
   ":hover": {
     color: theme.color.text.default,
   },
+  " *": {
```
stereobooster: Is it the same as `"& *": {`?

src/Tooltip/Tooltip.tsx
```diff
@@ -16,6 +16,7 @@ export interface BaseProps extends DefaultProps {
   children?: React.ReactNode
   /** Smart-positioned tooltip, with positioning reversed so it doesn't flow out of the window's bounding box. Currently works for left and top-positioned tooltips. */
   smart?: boolean
+  textId?: string
```
stereobooster:
```suggestion
id?: string
```
?

src/Textarea/Textarea.tsx
```diff
@@ -1,18 +1,21 @@
...
+import { useUniqueId } from "../useUniqueId"
+import { isCmdEnter, lighten } from "../utils"
 import { FormFieldControls, FormFieldError, inputFocus, Label } from "../utils/mixins"
+import styled from "../utils/styled"
 
 type ResizeOptions = "none" | "both" | "vertical" | "horizontal"
 
 export interface TextareaProps extends DefaultProps {
+  /** What is the identifier of this textarea? */
+  id?: string
```
stereobooster: `tabIndex`? Maybe introduce `type AccessibleInput = { ... }` and then reuse it across input, textarea, autocomplete 🤔

stereobooster:
lock icon is not focusable
Screenshot 2019-03-11 at 15 41 13

I'm not sure how this suppose to work. Copy button is not focusable.
Screenshot 2019-03-11 at 15 41 05

src/Input/Input.Field.tsx
```diff
@@ -128,39 +127,41 @@ const InputField: React.SFC<InputProps> = ({
...
+        aria-describedby={hint ? `input-hint-${id}` : undefined}
+        aria-label={label}
```
stereobooster: do we need both `aria-labelledby` and `aria-label`?

src/Input/Input.Field.tsx
stereobooster: Also move it to `Input/Input.tsx` so 

```tsx
aria-describedby={hint ? `input-hint-${id}` : undefined}
```

would be in the same file as 

```tsx
{hint && <Hint textId={`input-hint-${uniqueId}`}>{hint}</Hint>}
```

src/utils/mixins.ts
TejasQ: No. It is equivalent to `"> *"` instead. 🙂

src/Tooltip/Tooltip.tsx
TejasQ: Naaaaa

Issue #923, opened 2019-02-15
[Umbrella] Tree component features


stereobooster:
List of features we may want to add to `Tree` component which were discussed in @TejasQ PR https://github.com/contiamo/operational-ui/pull/919

- [ ] controlled version of tree component
- ~~`meta` field - to be able to attach additional data to node, to use it later in `onNodeClick` handler~~
- ~~`node` render prop - to override how leaf rendered~~
- [ ] support reorder for non-leaf nodes

fabien0102:
We took another direction for this component ;)

PR #924, opened 2019-02-15
**Feature:** Add meta field to the Tree nodes

stereobooster:
# Summary

Add `meta` field to the tree and `onClick` handler

# Related issue

https://github.com/contiamo/operational-ui/issues/923

# To be tested

Me

Tester 1

Tester 2

### DIFF
--- a/src/Tree/Tree.tsx
+++ b/src/Tree/Tree.tsx
@@ -1,24 +1,30 @@
 import * as React from "react"
 import { Draggable, Droppable, DropResult, ResponderProvided } from "react-beautiful-dnd"
-
 import styled from "../utils/styled"
 import ChildTree from "./ChildTree"
 
-export interface Tree {
+export interface TreeData<Meta> {
   label: string
   highlight?: boolean
-  childNodes?: Tree[]
+  childNodes?: Array<TreeData<Meta>>
   initiallyOpen?: boolean
   tag?: string
   disabled?: boolean
   color?: string
-  onClick?: (node: Tree) => void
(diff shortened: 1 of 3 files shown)


src/Tree/Tree.tsx
```diff
@@ -4,21 +4,22 @@ import { Draggable, Droppable, DropResult, ResponderProvided } from "react-beaut
...
   label: string
   highlight?: boolean
-  childNodes?: Tree[]
+  childNodes?: Array<Tree<Meta>>
   initiallyOpen?: boolean
   tag?: string
   disabled?: boolean
   color?: string
-  onClick?: (node: Tree) => void
+  meta?: Meta
```
stereobooster: cc @fabien0102

src/Tree/Tree.tsx
```diff
@@ -1,24 +1,30 @@
...
+export interface TreeData<Meta> {
   label: string
   highlight?: boolean
-  childNodes?: Tree[]
+  childNodes?: Array<TreeData<Meta>>
   initiallyOpen?: boolean
   tag?: string
   disabled?: boolean
   color?: string
-  onClick?: (node: Tree) => void
```
stereobooster: I moved it to different place. `onClick` didn't work in previous implementation, so it should not be breaking change.

Also I believe we need to move `onRemove` the same way and make it accept `node: TreeData<Meta>`

fabien0102:
We will go for a `onClick` handler on node instead :wink:

PR #931, opened 2019-02-21
**Feature:** Add support for SSR

TejasQ:
# Summary
This PR removes tight coupling with `window` and adds support for SSR. The only thing left to do here is to fork `react-json-view` and adjust it because it _also_ has a hard dependency on `window`.

# Related issue

# To be tested

Me

Tester 1

Tester 2

### DIFF
--- a/src/Debug/Debug.tsx
+++ b/src/Debug/Debug.tsx
@@ -94,12 +94,14 @@ class Debug<T> extends React.Component<DebugProps<T>, DebugState> {
 
   public dragStart = (event: React.DragEvent) => {
     event.persist()
-    const style = window.getComputedStyle(event.currentTarget, null)
+    if (window) {
+      const style = window.getComputedStyle(event.currentTarget, null)
 
-    this.setState(() => ({
-      contextLeft: parseInt(style.getPropertyValue("left"), 10) - event.clientX,
-      contextTop: parseInt(style.getPropertyValue("top"), 10) - event.clientY,
-    }))
+      this.setState(() => ({
+        contextLeft: parseInt(style.getPropertyValue("left"), 10) - event.clientX,
+        contextTop: parseInt(style.getPropertyValue("top"), 10) - event.clientY,
+      }))
+    }
   }
 
   public drop = (event: React.DragEvent) => {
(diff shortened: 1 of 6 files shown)


src/Debug/Debug.tsx
```diff
@@ -94,12 +94,14 @@ class Debug<T> extends React.Component<DebugProps<T>, DebugState> {
 
   public dragStart = (event: React.DragEvent) => {
     event.persist()
-    const style = window.getComputedStyle(event.currentTarget, null)
+    if (window) {
```
stereobooster:
```suggestion
    if (typeof window !== 'undefined') {
```
?

TejasQ: Accessing window on node throws because you're accessing something that doesn't exist.

TejasQ: image

TejasQ: Welcome to JS

fabien0102: I don't get it, if calling `window` cause a ReferenceError, we should use `typeof window` everywhere, isn't it? 🤔

fabien0102: I also propose to introduce a small utils `isClient` or `isSSR` to encapsulate this logic in proper way 😃

src/HeaderMenu/HeaderMenu.tsx
```diff
@@ -94,11 +94,15 @@ class HeaderMenu extends React.PureComponent<HeaderMenuProps, Readonly<HeaderMen
 
   public componentDidMount() {
     this.updateRenderedWidth()
-    window.addEventListener("resize", this.handleResize)
+    if (window) {
```
stereobooster: `componentDidMount` doesn't trigger on server as far as I remember

src/Debug/Debug.tsx
stereobooster: Update. I can't see how you gonna reach onDragStart function on server, so most likely we don't need this check at all

src/HeaderMenu/HeaderMenu.tsx
fabien0102: Sure, but to avoid this kind of question, a little `if (isClient)` is really not a huge cost ^^

TejasQ:
Thanks for the comments!

src/Debug/Debug.tsx
fabien0102: I still don't get it! ^^

src/useWindowSize/useWindowSize.test.tsx
```diff
@@ -51,9 +51,9 @@ describe("useURLState", () => {
 <div
   data-testid="result"
 >
-  200
+  -1
```
fabien0102: nop, the test set explicitly a window height and width

src/Debug/Debug.tsx
TejasQ: I will remove it

PR #940, opened 2019-02-27
**Feature:** Splash hooks & useInterval

JoshRosenstein:
# Summary
Convert Splash component to use hooks and optimized a bit.
- Removed State from main  [Splash Component](https://github.com/contiamo/operational-ui/blob/master/src/Splash/Splash.tsx)( window state moved to Splash.Animation)

- Added hook useInterval (forked from [use-interval](https://github.com/Hermanya/use-interval))
- Updated [Splash.Animation](https://github.com/contiamo/operational-ui/blob/master/src/Splash/Splash.Animation.tsx) to use hooks useWindowSize & useInterval
-useInterval has option for immediate, therefore animation is triggered immediately when loaded.
- Updated small logic where dynamic styles (top & left) do not pass through styled to eliminate hectic class generation
- Added fullscreen prop inorder to trigger useWindowSize 
- use type React.FC  ( SFC is depreciated)
- Fixed type warning on intialState array creator `Array.apply(null, { length: boxes })	
      .map(Number.call, Number)` to `Array.apply(null, Array(boxes)).map((_: any, i: number) => i)`

# Related issue
[#910](https://github.com/contiamo/operational-ui/issues/910) 

# To be tested

Me

### DIFF
--- a/src/Splash/Splash.Animation.tsx
+++ b/src/Splash/Splash.Animation.tsx
@@ -97,31 +97,24 @@ const Animation: React.FC<Props> = ({ isFullscreen, size = 600 }) => {
     5000,
     true,
   )
-  // Only will change if isFullscreen or size changes, a workaround from not having to set outter container width and height to max(window.height,window.width)
-  const Wrap = useCallback(
-    ({ children }) =>
-      isFullscreen ? (
-        <FullScreenWrap>
-          <Container>{children}</Container>
-        </FullScreenWrap>
-      ) : (
-        <Container style={{ width: size, height: size }}>{children}</Container>
-      ),
-    [isFullscreen, size],
-  )
 
-  return (
-    <Wrap>
-      {state.coordinates.map((coord: { x: number; y: number }, index: number) => (
-        <Box
-          key={index}
-          style={{
-            top: `${(coord.x / (squares - 1)) * 100}%`,
-            left: `${(coord.y / (squares - 1)) * 100}%`,
-          }}
-        />
-      ))}
-    </Wrap>
+  const children = state.coordinates.map((coord: { x: number; y: number }, index: number) => (
+    <Box
+      key={index}
+      style={{
+        top: `${(coord.x / (squares - 1)) * 100}%`,
+        left: `${(coord.y / (squares - 1)) * 100}%`,
+      }}
+    />
+  ))
+
+  // Only will change if isFullscreen or size changes, a workaround from not having to set outter container width and height to max(window.height,window.width)
+  return isFullscreen ? (
+    <FullScreenWrap>
+      <Container>{children}</Container>
+    </FullScreenWrap>
+  ) : (
+    <Container style={{ width: size, height: size }}>{children}</Container>
   )
 }
 
--- a/src/Splash/Splash.Animation.tsx
+++ b/src/Splash/Splash.Animation.tsx
@@ -1,8 +1,10 @@
-import * as React from "react"
-
+import React, { useState } from "react"
+import useInterval from "../useInterval"
+import useWindowSize from "../useWindowSize"
 import styled from "../utils/styled"
 
 export interface Props {
+  fullscreen?: boolean
   size?: number
 }
 
@@ -34,73 +36,75 @@ const bounce = (coord: number): number => {
   return coord
 }
 
-const Container = styled("div")<{ size: number }>(({ size }: { size: number }) => ({
-  width: size,
-  height: size,
+const Container = styled("div")({
   position: "absolute",
   top: "50%",
   left: "50%",
   transform: "translate3d(-50%, -50%, 0)",
-}))
+})
 
-const Box = styled("div")<{ x: number; y: number }>(({ x, y }) => ({
+/// Move highly highly dynamic style out of css-js to prevent uneeded classname generation
+
+const Box = styled("div")({
   position: "absolute",
   transition: "all 0.5s ease-in-out",
-  top: `calc(${(x / (squares - 1)) * 100}% + 4px)`,
-  left: `calc(${(y / (squares - 1)) * 100}% + 4px)`,
+
   borderRadius: 6,
   width: `calc(${100 / (squares - 1)}% - 8px)`,
   height: `calc(${100 / (squares - 1)}% - 8px)`,
   backgroundColor: "rgba(255, 255, 255, 0.06)",
-}))
+})
 
-class Animation extends React.Component<Props, State> {
-  public state = {
-    animationStep: 0,
-    coordinates: Array.apply(null, { length: boxes })
-      .map(Number.call, Number)
-      .map(() => ({ x: integerRandom(squares), y: integerRandom(squares) })),
+const initialState = {
+  animationStep: 0,
+  coordinates: Array.apply(null, Array(boxes))
+    .map((_: any, i: number) => i) // Fixes typescript Error
+    .map(() => ({ x: integerRandom(squares), y: integerRandom(squares) })),
+}
+
+const Animation: React.FC<Props> = ({ fullscreen, size }) => {
+  let containerSize = size || 600
+
+  if (fullscreen) {
+    const windowSize = useWindowSize()
+    containerSize = Math.max(windowSize.width, windowSize.height) - 20
   }
 
-  public animationInterval?: number
-
-  // Shift the coordinate of every third tile in a random direction.
-  // Each animation shifts a different set of tiles.
-  public shiftSomeTiles() {
-    this.setState(prevState => ({
-      animationStep: prevState.animationStep + 1,
-      coordinates: prevState.coordinates.map((coord: { x: number; y: number }, index: number) => {
-        if (index % 3 === prevState.animationStep % 3) {
-          const dx = integerRandom(3) - 1
-          const dy = integerRandom(3) - 1
-          return {
-            x: bounce(coord.x + dx),
-            y: bounce(coord.y - dy),
+  const [state, updateAnimation] = useState<State>(initialState)
+  useInterval(
+    () => {
+      updateAnimation({
+        animationStep: state.animationStep + 1,
+        coordinates: state.coordinates.map((coord: { x: number; y: number }, index: number) => {
+          if (index % 3 === state.animationStep % 3) {
+            const dx = integerRandom(3) - 1
+            const dy = integerRandom(3) - 1
+            return {
+              x: bounce(coord.x + dx),
+              y: bounce(coord.y - dy),
+            }
           }
-        }
-        return coord
-      }),
-    }))
-  }
+          return coord
+        }),
+      })
+    },
+    5000,
+    true,
+  )
 
-  public componentDidMount() {
-    this.animationInterval = window.setInterval(this.shiftSomeTiles.bind(this), 5000)
-  }
-
-  public componentWillUnmount() {
-    clearInterval(this.animationInterval)
-  }
-
-  public render() {
-    const size = this.props.size || 600
-    return (
-      <Container size={size}>
-        {this.state.coordinates.map((coord: { x: number; y: number }, index: number) => (
-          <Box key={index} x={coord.x} y={coord.y} />
-        ))}
-      </Container>
-    )
-  }
+  return (
+    <Container style={{ width: containerSize, height: containerSize }}>
+      {state.coordinates.map((coord: { x: number; y: number }, index: number) => (
+        <Box
+          key={index}
+          style={{
+            top: `${(coord.x / (squares - 1)) * 100}%`,
+            left: `${(coord.y / (squares - 1)) * 100}%`,
+          }}
+        />
+      ))}
+    </Container>
+  )
 }
 
 export default Animation
--- a/src/Splash/Splash.Animation.tsx
+++ b/src/Splash/Splash.Animation.tsx
@@ -1,10 +1,9 @@
-import React, { useState } from "react"
+import React, { useCallback, useState } from "react"
 import useInterval from "../useInterval"
-import useWindowSize from "../useWindowSize"
 import styled from "../utils/styled"
 
 export interface Props {
-  fullscreen?: boolean
+  isFullscreen?: boolean
   size?: number
 }
 
@@ -36,19 +35,30 @@ const bounce = (coord: number): number => {
   return coord
 }
 
+// css Hack so we dont need to worry about max(window.height,window.width)
+const FullScreenWrap = styled("div")({
+  position: "absolute",
+  width: "100%",
+  ":after": {
+    content: "''",
+    display: "block",
+    paddingBottom: "100%",
+  },
+})
+
 const Container = styled("div")({
   position: "absolute",
   top: "50%",
   left: "50%",
   transform: "translate3d(-50%, -50%, 0)",
+  width: "100%",
+  height: "100%",
 })
 
 /// Move highly highly dynamic style out of css-js to prevent uneeded classname generation
-
 const Box = styled("div")({
   position: "absolute",
   transition: "all 0.5s ease-in-out",
-
   borderRadius: 6,
   width: `calc(${100 / (squares - 1)}% - 8px)`,
   height: `calc(${100 / (squares - 1)}% - 8px)`,
@@ -57,20 +67,15 @@ const Box = styled("div")({
 
 const initialState = {
   animationStep: 0,
-  coordinates: Array.apply(null, Array(boxes))
-    .map((_: any, i: number) => i) // Fixes typescript Error
-    .map(() => ({ x: integerRandom(squares), y: integerRandom(squares) })),
+  coordinates: Array.from(Array(boxes), (_, index) => index).map(() => ({
+    x: integerRandom(squares),
+    y: integerRandom(squares),
+  })),
 }
 
-const Animation: React.FC<Props> = ({ fullscreen, size }) => {
-  let containerSize = size || 600
-
-  if (fullscreen) {
-    const windowSize = useWindowSize()
-    containerSize = Math.max(windowSize.width, windowSize.height) - 20
-  }
-
+const Animation: React.FC<Props> = ({ isFullscreen, size = 600 }) => {
   const [state, updateAnimation] = useState<State>(initialState)
+
   useInterval(
     () => {
       updateAnimation({
@@ -92,8 +97,20 @@ const Animation: React.FC<Props> = ({ fullscreen, size }) => {
     true,
   )
 
+  const Wrap = useCallback(
+    ({ children }) =>
+      isFullscreen ? (
+        <FullScreenWrap>
+          <Container>{children}</Container>
+        </FullScreenWrap>
+      ) : (
+        <Container style={{ width: size, height: size }}>{children}</Container>
+      ),
+    [isFullscreen, size],
+  )
+
   return (
-    <Container style={{ width: containerSize, height: containerSize }}>
+    <Wrap>
       {state.coordinates.map((coord: { x: number; y: number }, index: number) => (
         <Box
           key={index}
@@ -103,7 +120,7 @@ const Animation: React.FC<Props> = ({ fullscreen, size }) => {
           }}
         />
       ))}
-    </Container>
+    </Wrap>
   )
 }
 
--- a/src/Splash/Splash.Animation.tsx
+++ b/src/Splash/Splash.Animation.tsx
@@ -108,7 +108,7 @@ const Animation: React.FC<Props> = ({ isFullscreen, size = 600 }) => {
     />
   ))
 
-  // Only will change if isFullscreen or size changes, a workaround from not having to set outter container width and height to max(window.height,window.width)
+  // Only will change if isFullscreen or size changes, a workaround from not having to set outer container width and height to max(window.height, window.width)
   return isFullscreen ? (
     <FullScreenWrap>
       <Container>{children}</Container>
--- a/src/Splash/Splash.Animation.tsx
+++ b/src/Splash/Splash.Animation.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useState } from "react"
+import React, { useState } from "react"
 import useInterval from "../useInterval"
 import styled from "../utils/styled"
 
--- a/src/Splash/Splash.Animation.tsx
+++ b/src/Splash/Splash.Animation.tsx
@@ -35,7 +35,8 @@ const bounce = (coord: number): number => {
   return coord
 }
 
-// css Hack so we dont need to worry about max(window.height,window.width)
+// css Hack so we dont need to worry about max(window.height,window.width)- Only needed when fullscreen is enabled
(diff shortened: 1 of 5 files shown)


src/useInterval/index.ts
```diff
@@ -0,0 +1,37 @@
+import { useEffect, useRef } from "react"
```
stereobooster: The simplest implementation would be

```js
import { useEffect } from 'react';

const useInterval = (fn: () => void, delay: number) => {
  useEffect(() => {
    const id = setInterval(fn, delay)
    return () => clearInterval(id)
  })
};

export default useInterval;
```

src/Splash/Splash.Animation.tsx
```diff
@@ -34,73 +36,75 @@ const bounce = (coord: number): number => {
...
-  height: size,
+const Container = styled("div")({
   position: "absolute",
   top: "50%",
   left: "50%",
   transform: "translate3d(-50%, -50%, 0)",
-}))
+})
+
+/// Move highly highly dynamic style out of css-js to prevent uneeded classname generation
```
TejasQ: 👍

src/Splash/Splash.Animation.tsx
```diff
@@ -34,73 +36,75 @@ const bounce = (coord: number): number => {
...
-          }
-        }
-        return coord
-      }),
-    }))
-  }
+const initialState = {
+  animationStep: 0,
+  coordinates: Array.apply(null, Array(boxes))
+    .map((_: any, i: number) => i) // Fixes typescript Error
```
TejasQ: Let's remove this `any`.

src/Splash/Splash.Animation.tsx
```diff
@@ -34,73 +36,75 @@ const bounce = (coord: number): number => {
...
+    .map(() => ({ x: integerRandom(squares), y: integerRandom(squares) })),
+}
 
-  public componentDidMount() {
-    this.animationInterval = window.setInterval(this.shiftSomeTiles.bind(this), 5000)
-  }
+const Animation: React.FC<Props> = ({ fullscreen, size }) => {
+  let containerSize = size || 600
```
TejasQ: Render is supposed to be pure => all mutable values `useRef`.

src/Splash/Splash.Animation.tsx
```diff
@@ -34,73 +36,75 @@ const bounce = (coord: number): number => {
...
 
-  public componentWillUnmount() {
-    clearInterval(this.animationInterval)
+  if (fullscreen) {
```
TejasQ: It's a nicer naming convention to prefix booleans with `is`:

```suggestion
  if (isFullscreen) {
```

src/Splash/Splash.tsx
```diff
@@ -2,6 +2,7 @@ import * as React from "react"
 
 import { readableTextColor } from "../utils"
 import { expandColor, OperationalStyleConstants } from "../utils/constants"
+
```
TejasQ: Why the new line? 🤔

src/useInterval/index.ts
```diff
@@ -0,0 +1,37 @@
...
+
+// Ref: https://github.com/Hermanya/use-interval/blob/master/src/index.tsx
+
+// tslint:disable-next-line
```
TejasQ:
```suggestion
// tslint:disable-next-line always say why even if it's obvious please
```

src/useInterval/index.ts
```diff
@@ -0,0 +1,37 @@
...
+const noop = () => {}
+
+/**
+ * useInterval
+ */
+export function useInterval(callback: () => void, delay: number | null, immediate?: boolean) {
+  const savedCallback = useRef(noop)
```
TejasQ: How about `useCallback`?

src/useInterval/useInterval.test.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+beforeEach(cleanup)
+
+const MyComponent: React.FC<{ fn: () => void; delay: number | null; immediate?: boolean }> = ({
+  fn,
+  delay,
+  immediate,
+}) => {
+  useInterval(fn, delay, immediate)
+  return <span>A Interval is running </span>
```
TejasQ:
```suggestion
  return <span>An Interval is running </span>
```

src/useInterval/index.ts
stereobooster: We don't need neither `useCallback` neither `useRef`

src/useInterval/index.ts
JoshRosenstein: @stereobooster ,Dan wrote a [nice piece on this](https://overreacted.io/making-setinterval-declarative-with-react-hooks/).

Simple Implementation works for [simple cases](https://codesandbox.io/s/7wlxk1k87j) ( and how its currently being used in Splash)

But this version is more scalable/reusable ...
>"But setInterval() does not “forget”. It will forever reference the old props and state until you replace it — which you can’t do without resetting the time."


Where useRef comes into play...
>"The problem boils down to this: 
> - We do setInterval(callback1, delay) with callback1 from first render.
> -  We have callback2 from next render that closes over fresh props and state.
>  -   But we can’t replace an already existing interval without resetting the time!

>So what if we didn’t replace the interval at all, and instead introduced a mutable savedCallback variable pointing to the latest interval callback?"

>"This mutable savedCallback needs to “persist” across the re-renders. So it can’t be a regular variable. We want something more like an instance field."

For [Example](https://codesandbox.io/s/l240mp2pm7), this allows us to change the delay or callback between renders without resetting the time 


But if you want be to change to simple version, i will do that!

stereobooster: My concern is that the code should be clear, so people after us would be able to read it easily and change it if they want to. It takes quite some iterations to understand what happens here. After reading Dan's article this all start to make sense, but before it seemed over-engineered. So adding link to the Dan's article and maybe some more comments would help

src/Splash/Splash.Animation.tsx
JoshRosenstein: 👍 - gonna change this again to `Array.from(Array(5), (_, index) => index )`

src/useInterval/index.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  // Remember the latest callback.
```
stereobooster:
```suggestion
  // Remember the latest callback.
  // useEffect has no second argument so it will be executed after each render
  // but we don't want to change this value directly in the body of the render function, 
  // because render should be pure function
```

src/useInterval/index.ts
```diff
@@ -0,0 +1,37 @@
...
+
+// Ref: https://github.com/Hermanya/use-interval/blob/master/src/index.tsx
+
+// tslint:disable-next-line
+const noop = () => {}
+
+/**
+ * useInterval
+ */
+export function useInterval(callback: () => void, delay: number | null, immediate?: boolean) {
```
stereobooster: As far as I can tell we don't use `immediate`, right? Maybe let's add it when we will need it

src/useInterval/index.ts
```diff
@@ -0,0 +1,37 @@
...
+  const savedCallback = useRef(noop)
+
+  // Remember the latest callback.
+  useEffect(() => {
+    savedCallback.current = callback
+  })
+
+  // Execute callback if immediate is set.
+  useEffect(() => {
```
stereobooster: this will be removed if we will drop support for `immediate`

src/useInterval/index.ts
```diff
@@ -0,0 +1,37 @@
...
+    }
+  }, [immediate])
+
+  // Set up the interval.
+  useEffect(() => {
+    if (delay === null) {
+      return undefined
+    }
+    const tick = () => savedCallback.current()
+    const id = setInterval(tick, delay)
```
stereobooster:
```suggestion
  const id = setInterval(() => savedCallback.current(), delay)
```

Matter of taste I guess

src/useInterval/index.ts
```diff
@@ -0,0 +1,37 @@
+import { useEffect, useRef } from "react"
+
+// Ref: https://github.com/Hermanya/use-interval/blob/master/src/index.tsx
+
+// tslint:disable-next-line
+const noop = () => {}
+
+/**
+ * useInterval
```
stereobooster:
```suggestion
 * Hook version of setInterval. Pass null to the delay to cancel an execution.
```

src/useInterval/index.ts
```diff
@@ -0,0 +1,37 @@
+import { useEffect, useRef } from "react"
+
+// Ref: https://github.com/Hermanya/use-interval/blob/master/src/index.tsx
+
+// tslint:disable-next-line
+const noop = () => {}
```
stereobooster:
```suggestion
const noop = () => undefined
```

I guess this will work without `tslint:disable-next-line`

TejasQ: We can also import `noop` from `_`.

src/Splash/Splash.Animation.tsx
JoshRosenstein: @TejasQ - will change name, however-
This is a big mistake on my part since Hooks should not be conditional. I think i will have to initiate `useWindowSize`, and simply create a handler `const getWidth=()=> isFullscreen? Math.max(windowSize.width, windowSize.height): (size | 500)`

We should add a tslint rules, tslint-react currently has open issue [Add linting for React Hooks](https://github.com/palantir/tslint-react/issues/186) with a [PR](https://github.com/palantir/tslint-react/pull/204) ready to go.

src/Splash/Splash.Animation.tsx
JoshRosenstein: Eliminating this

src/Splash/Splash.tsx
JoshRosenstein: 😖

src/useInterval/index.ts
JoshRosenstein: `const noop = () => undefined` will throw type errors, will import noop from lodash

src/useInterval/index.ts
JoshRosenstein: Added that so when the splash screen loads, the animation immediately begins- rather then waiting 5 seconds before the first movement.

src/Splash/Splash.Animation.tsx
```diff
@@ -34,73 +35,93 @@ const bounce = (coord: number): number => {
...
-            x: bounce(coord.x + dx),
-            y: bounce(coord.y - dy),
+const Animation: React.FC<Props> = ({ isFullscreen, size = 600 }) => {
+  const [state, updateAnimation] = useState<State>(initialState)
+
+  useInterval(
+    () => {
+      updateAnimation({
+        animationStep: state.animationStep + 1,
+        coordinates: state.coordinates.map((coord: { x: number; y: number }, index: number) => {
```
stereobooster: Side question. Why we can't use css animation for this? From my POV we need to do animation with JS only if it is interruptible animation, or we need to sync animations (like in game)

JoshRosenstein: I wouldnt even know where to begin to imitate with pure css. Then again. im no expert, but some potential obstacles that come to mind
-how to not make it obvious the boxes will be moving around in an repeating loop. 
-How to get the boxes to stay within the window screen, since we wont be able to bounce back in.
- synchronizing the timing of only 1/3 moving every x seconds
- currently every 50 individual box moves in a random spot- we wouldn't be able to add 50 separate keyframes - we would have to make maybe a handful of random movement loops and assign them randomly.
- also not quite sure how performance is with 16 elements transitioning at once in css

OR we use 0 css and all JS with canvas. love stuff like [this](https://codepen.io/soulwire/full/Ffvlo)

src/useInterval/index.ts
JoshRosenstein: @stereobooster - you good with this or would you like me to take that out?
 thing is though on the splash screen like docs intro- when page loads and if it doesnt freeze like it currently does, people will click a button within 5 seconds, so i thought either speed up transition delay, or what this does- start the first transition immediately

src/Splash/Splash.Animation.tsx
TejasQ: Let's just use JS for now. This is a refactor not a rewrite.

src/Splash/Splash.Animation.tsx
```diff
@@ -34,73 +35,93 @@ const bounce = (coord: number): number => {
...
-  width: size,
-  height: size,
+// css Hack so we dont need to worry about max(window.height,window.width)
+const FullScreenWrap = styled("div")({
+  position: "absolute",
+  width: "100%",
+  ":after": {
+    content: "''",
+    display: "block",
+    paddingBottom: "100%",
```
TejasQ: What is this? 🤔 What does it do?

src/useInterval/index.ts
TejasQ: Agree. @JoshRosenstein please add some documentation if you wish to keep the component the way it is. ❤️

src/useInterval/index.ts
TejasQ: Leave it in @JoshRosenstein

src/Splash/Splash.Animation.tsx
JoshRosenstein: Its a css trick (only used if fullscreen is enabled) that way the outer component doesnt need to change for every window resize therefore doesnt have to use useWindowSize. Before, needed to take max of window height and width to maintain a maximum square container- which all the inner moving boxes then maintain squared shapes.

[Where i got the idea](https://spin.atomicobject.com/2015/07/14/css-responsive-square/)

mpotomin: Interesting. Would [this solution](https://codepen.io/anon/pen/xBOmPW) work in this case? If I understand correctly, we're trying to have a square container relative to the window width, right?

src/Splash/Splash.Animation.tsx
```diff
@@ -34,73 +35,94 @@ const bounce = (coord: number): number => {
...
+  },
+})
+
+const Container = styled("div")({
   position: "absolute",
   top: "50%",
   left: "50%",
   transform: "translate3d(-50%, -50%, 0)",
-}))
+  width: "100%",
```
mpotomin: Can we use `vh` here?

src/Splash/Splash.Animation.tsx
```diff
@@ -34,73 +35,94 @@ const bounce = (coord: number): number => {
...
-  }
+          return coord
+        }),
+      })
+    },
+    5000,
+    true,
+  )
+  // Only will change if isFullscreen or size changes, a workaround from not having to set outter container width and height to max(window.height,window.width)
+  const Wrap = useCallback(
```
mpotomin: This is a React component. We probably don't need to `useCallback` on it because it's a function of React to rerender components when their props/state change anyway. 🤔

src/useInterval/README.md
```diff
@@ -0,0 +1,112 @@
+Hook version of setInterval.
```
mpotomin:
```suggestion
An implementation of `setInterval` with Hooks.
```

src/useInterval/README.md
```diff
@@ -0,0 +1,112 @@
...
+
+### Referece
```
mpotomin:
```suggestion
### Reference
```

src/useInterval/README.md
```diff
@@ -0,0 +1,112 @@
...
+    </>
+  )
+}
+
+;<Counter />
+```
+
+#### Immediate argument
+
+If you need your callback to be delayed for an extended period of time, but you want it to trigger imediatley upon mounting, then pass in immediate.
```
mpotomin:
```suggestion
If you need your callback to be delayed for an extended period of time, but you want it to trigger immediately upon mounting, then pass in immediate as the third argument.
```

src/useInterval/README.md
```diff
@@ -0,0 +1,112 @@
...
+
+```jsx
+function MyComp() {
+  const [inverted, setInvert] = useState(false)
+
```
mpotomin:
```suggestion
```

src/useInterval/README.md
```diff
@@ -0,0 +1,112 @@
...
+    10000,
+    true,
+  )
+
+  // Invert colors
+  useInterval(() => {
+    setInvert2(!inverted2)
+  }, 10000)
+
+  // Make it faster every second!
```
mpotomin: What does this mean/do?

src/useInterval/README.md
```diff
@@ -0,0 +1,112 @@
...
+
+  return (
+    <div>
```
mpotomin: This can just be a fragment: 
```suggestion
    <>
```

src/useInterval/README.md
```diff
@@ -0,0 +1,112 @@
...
+      <div
+        style={{
+          background: inverted ? "navy" : "white",
+          color: inverted ? "white" : "navy",
+        }}
+      >
+        <h3> This immediatly gets inverted with navy BG </h3>
```
mpotomin:
```suggestion
        <h3>This immediately gets inverted with a navy background</h3>
```

src/useInterval/README.md
```diff
@@ -0,0 +1,112 @@
...
+      </div>
+
+      <div
+        style={{
+          background: inverted2 ? "navy" : "white",
+          color: inverted2 ? "white" : "navy",
+        }}
+      >
+        <h3> This needs to wait for delay to pass to make Navy </h3>
```
mpotomin:
```suggestion
        <h3>This needs to wait for the delay to pass to fire</h3>
```

src/useInterval/index.ts
mpotomin: @JoshRosenstein have you seen these comments?

JoshRosenstein: @mpotomin - It was my understanding the "simplest implementation" version breaks, and should keep the current useInterval as is, and i added additional comments since, Did i miss something ? Or just forgot to resolve? 

[Edit useInterval Test](https://codesandbox.io/s/2w2qyyp5pp?fontsize=14)

TejasQ: You just forgot to acknowledge. Thanks for responding, @JoshRosenstein! @stereobooster is that okay with you?

src/Splash/Splash.Animation.tsx
JoshRosenstein: That will work in a way but does not produce the same result as the original code intended. Here you are setting the boxes to be sized based on only the screens width, and the original code- the size of the boxes are a function of the # total boxes, and the max(window.height, window.width). if css still had max function your box size would be max(40vh,40vw). Therefore what the FullScreenWrap does, is creates the container that that has a height=width=max(window.height, window.width), and the underlying box elements have a size of `calc(${100 / (squares - 1)}% - 8px)`, where the percentages would then be referencing max(window.height, window.width). I dont know why the the original code did this, and dont know where the component is used, so i wanted to ensure the commit had the same expected results as before and was not a breaking change.

src/Splash/Splash.Animation.tsx
TejasQ: You can also declare it _outside_ of this currently component.

JoshRosenstein: During testing the entire Animation component was re rendering, rather than just the internal boxes , wrapping it stopped that, however a better way to do it is not define the Wrap Component beforehand, define the children beforehand.

```jsx
const children= state.coordinates.map((coord: { x: number; y: number }, index: number) => (
    <Box
      key={index}
      style={{
        top: `${(coord.x / (squares - 1)) * 100}%`,
        left: `${(coord.y / (squares - 1)) * 100}%`,
      }}
    />
  ))

  // Only will change if isFullscreen or size changes, a workaround from not having to set outter container width and height to max(window.height,window.width)
  return (
    isFullscreen ? (
      <FullScreenWrap>
        <Container>{children}</Container>
      </FullScreenWrap>
    ) : (
      <Container style={{ width: size, height: size }}>{children}</Container>
    )
  )
```

src/Splash/Splash.Animation.tsx
JoshRosenstein: Not if we need the container to match the "FullScreenWrap" size which = max(window.height,window.width). see above.

src/Splash/Splash.Animation.tsx
JoshRosenstein: Forgot to mention, setting box sizes to vw unit, also may not render the expected result when isFullScreen= false, and a size is given.

src/useInterval/README.md
JoshRosenstein: " // Make it faster every second!"- needs to be deleted, was a copy and paste from example above it.

src/Splash/Splash.Animation.tsx
TejasQ: Great. Do it.

src/useInterval/index.ts
```diff
@@ -0,0 +1,42 @@
+import noop from "lodash/noop"
+import { useEffect, useRef } from "react"
+
+// From: https://overreacted.io/making-setinterval-declarative-with-react-hooks/
+
+/**
+ * Hook version of setInterval. Pass null to the delay to cancel an execution.
+ */
```
TejasQ: Can you add JSDoc style documentation to this?

I feel `@param {delay} Pass null to the delay to cancel an execution.` or similar would be of great benefit here.

src/useInterval/index.ts
```diff
@@ -0,0 +1,44 @@
...
+ * @see https://overreacted.io/making-setinterval-declarative-with-react-hooks/
+ */
+export function useInterval(callback: () => void, delay: number | null, immediate?: boolean) {
+  const savedCallback = useRef(noop)
+
+  // Remember the latest callback.
+  // useEffect has no second argument so it will be executed after each render
+  // but we don't want to change this value directly in the body of the render function,
+  // because render should be pure function
+  // useEffect has no second argument so it will be executed after each render
```
TejasQ: Bad copy/paste. See line 15.

src/Splash/Splash.Animation.tsx
```diff
@@ -34,73 +35,87 @@ const bounce = (coord: number): number => {
...
+    <Box
+      key={index}
+      style={{
+        top: `${(coord.x / (squares - 1)) * 100}%`,
+        left: `${(coord.y / (squares - 1)) * 100}%`,
+      }}
+    />
+  ))
+
+  // Only will change if isFullscreen or size changes, a workaround from not having to set outter container width and height to max(window.height,window.width)
```
TejasQ:
```suggestion
  // Only will change if isFullscreen or size changes, a workaround from not having to set outer container width and height to max(window.height, window.width)
```

src/useInterval/useInterval.test.tsx
```diff
@@ -0,0 +1,75 @@
+import React from "react"
+import { cleanup, render } from "react-testing-library"
+
+import useInterval, { useInterval as namedHook } from "."
+
+// ref https://github.com/Hermanya/use-interval/blob/master/src/test.tsx
+
```
TejasQ:
```suggestion
```

PR #941, opened 2019-02-28
Tabs hooks

JoshRosenstein:
# Summary
Refactor Tabs to use hooks

# Related issue

# To be tested

Me

Tester 1

Tester 2

### DIFF
--- a/src/Internals/Tabs.tsx
+++ b/src/Internals/Tabs.tsx
@@ -1,4 +1,4 @@
-import * as React from "react"
+import React, { useCallback, useState } from "react"
 import Icon, { IconName } from "../Icon/Icon"
 import Spinner from "../Spinner/Spinner"
 import styled from "../utils/styled"
@@ -17,7 +17,7 @@ export interface Props {
   activeTabName?: string
   onTabChange?: (newTabName: string) => void
   condensed?: boolean
-  children: (childrenConfig: { tabsBar: React.ReactNode; activeChildren: React.ReactNode }) => React.ReactNode
+  children: (childrenConfig: { tabsBar: React.ReactNode; activeChildren: React.ReactNode }) => JSX.Element
 }
 
 export interface State {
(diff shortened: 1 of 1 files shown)


src/Internals/Tabs.tsx
```diff
@@ -54,56 +54,44 @@ const Tab = styled("div")<{ active?: boolean; condensed?: boolean }>(({ theme, a
   },
 }))
 
-class Tabs extends React.Component<Props, State> {
-  public state = {
-    activeTab: 0,
+const Tabs = ({ onTabChange, tabs, activeTabName, condensed, children }: Props) => {
+  let initTab = 0
```
TejasQ: You should `useRef` for mutable values inside components since `render` is always supposed to be pure.

src/Internals/Tabs.tsx
```diff
@@ -17,7 +17,7 @@ export interface Props {
   activeTabName?: string
   onTabChange?: (newTabName: string) => void
   condensed?: boolean
-  children: (childrenConfig: { tabsBar: React.ReactNode; activeChildren: React.ReactNode }) => React.ReactNode
+  children: (childrenConfig: { tabsBar: React.ReactNode; activeChildren: React.ReactNode }) => JSX.Element
```
TejasQ: I'm not sure this is correct? It's expected to return a ReactNode and not some arbitrary JSX element. 🤔

src/Internals/Tabs.tsx
JoshRosenstein: Thanks! This is exact reason why I'm doing this, to get feedback in order to learn! (...btw these commits are my first time using hooks api, and i am also a newbie at TS)- just a heads up so you dont get too relaxed reviewing my commits 😆

src/Internals/Tabs.tsx
JoshRosenstein: dont remember why i did that- maybe a type error when i first began constructing the function and never changed it back. will revert

JoshRosenstein: k just saw why i changed it after running start. Under Cards and Pages- I get typescript error
'JSX element type 'ReactNode' is not a constructor function for JSX elements.'

I'll try to figure it out, and then will have to go back to the other ones i have changed as well.

JoshRosenstein: [Issue](https://github.com/Microsoft/TypeScript/issues/21699)

> "Yeah, unfortunately right now when returning children you need to cast them to as JSX.Element | null "

JoshRosenstein: as stated on [Modal PR](https://github.com/contiamo/operational-ui/pull/942#discussion_r261436897), When its wrapped in a fragment, no type error is thrown

src/Internals/Tabs.tsx
```diff
@@ -54,56 +54,50 @@ const Tab = styled("div")<{ active?: boolean; condensed?: boolean }>(({ theme, a
...
-    return activeTab
-  }
+  const onTabClick = useCallback(
+    (index: number) => {
+      setActiveTab(index)
+      if (onTabChange) {
+        onTabChange(tabs[index].name)
+      }
+    },
+    [onTabChange, tabs],
```
stereobooster: if `onTabChange` is function constructed on the fly (arrow function), like

```tsx
<Tabs onTabChange={() => { /* do something */ }} />
```

then it will trigger recalculation of `useCallback` on every render. You can fix it like this

```tsx
const Tabs = (props: Props) => {
  const { tabs, activeTabName, condensed, children } = props
  //..
  const onTabClick = useCallback(
    (index: number) => {
      setActiveTab(index)
      if (props.onTabChange) {
        props.onTabChange(tabs[index].name)
      }
    },
    [tabs],
)
```

stereobooster: > Pass an inline callback and an array of inputs. useCallback will return a memoized version of the callback that only changes if one of the inputs has changed. This is useful when passing callbacks to optimized child components that rely on reference equality to prevent unnecessary renders (e.g. shouldComponentUpdate).

But we use it inside arrow function 

```tsx
onClick={() => onTabClick(index)}
```

so `useCallback` doesn't add any value here. We can construct function directly without `useCallback`

JoshRosenstein: Yep makes since. Thanks!

src/Internals/Tabs.tsx
TejasQ: Great! Thanks Josh! That clears that up!

TejasQ: Can you add a ref to this in the code?

src/Internals/Tabs.tsx
JoshRosenstein: wanted to leave this here though [Edit 7kym96lyp0](https://codesandbox.io/s/7kym96lyp0?fontsize=14).

--- Rather using the usecallback for issues below the example above shows how they replace it with useEventCallback, the only difference between that example and this snip, is that we are having to reference the tabs props because the api is requiring us to return the name rather than index. 
Stemmed from [react #14099](https://github.com/facebook/react/issues/14099#issuecomment-457885333)

src/Internals/Tabs.tsx
```diff
@@ -54,56 +54,48 @@ const Tab = styled("div")<{ active?: boolean; condensed?: boolean }>(({ theme, a
...
-  public state = {
-    activeTab: 0,
-  }
-
-  private onTabClick(index: number) {
-    this.setState(() => ({ activeTab: index }))
-    if (this.props.onTabChange) {
-      this.props.onTabChange(this.props.tabs[index].name)
+const Tabs = ({ onTabChange, tabs, activeTabName, condensed, children }: Props) => {
+  // Need to check if we need activeTabName can change after first render
```
mpotomin: What does this mean? Could you please reword it?

JoshRosenstein: Was meant to be placeholder for discussion piece. but looking at the other components that use Tabs, by only using 'activeTabName'- as an intializer and not a controller, this is fine and i will remove comment.

PR #942, opened 2019-02-28
Modal hooks

JoshRosenstein:
# Summary
Updated [Internals/Modal](https://github.com/contiamo/operational-ui/blob/master/src/Internals/Modal.tsx)  to use hooks.
Also,made small change on its type and the state model since options was the only thing stored.

# Related issue

# To be tested

Me

Tester 1

Tester 2

### DIFF
--- a/src/Internals/Modal.tsx
+++ b/src/Internals/Modal.tsx
@@ -10,7 +10,7 @@ export interface ModalOptions {
 export type State = Partial<ModalOptions>
 
 export interface Props {
-  children: (modal: (options: ModalOptions) => void) => JSX.Element
+  children: (modal: (options: ModalOptions) => void) => React.ReactNode
 }
 
 export const Modal = ({ children }: Props) => {
--- a/src/Internals/Modal.tsx
+++ b/src/Internals/Modal.tsx
@@ -1,5 +1,4 @@
-import * as React from "react"
-
+import React, { useCallback, useState } from "react"
 import ControlledModal from "./ControlledModal"
 
 export interface ModalOptions {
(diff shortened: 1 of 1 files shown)


src/Internals/Modal.tsx
```diff
@@ -8,47 +7,33 @@ export interface ModalOptions {
...
 }
 
-export interface State {
-  options: Partial<ModalOptions>
-}
+export type State = Partial<ModalOptions>
 
 export interface Props {
-  children: (modal: (options: ModalOptions) => void) => React.ReactNode
+  children: (modal: (options: ModalOptions) => void) => JSX.Element
```
TejasQ: I'm not sure this is correct? We want `children` to return a ReactNode...

JoshRosenstein:
@TejasQ - i was also thinking of that- but couple things
 -wanted to "upgrade" this component before potentially making a breaking changes
But there are also several different options to tackle this. 
Heres some reads samples
[kentcdodds](https://kentcdodds.com/blog/react-hooks-whats-going-to-happen-to-render-props)
[the hydra pattern](https://americanexpress.io/hydra/)

src/Internals/Modal.tsx
JoshRosenstein: same issue as Tabs - will have to keep as JSX.Element | null , unless there's another solution im missed

JoshRosenstein: So i opened this back up and ran start- It didnt throw type error when using ReactNode.
So I think for the other ones that throw a type error  - When I wrap the output in a fragment like Tabs is, it wont throw. Do you prefer changing the type, or wrapping in a <></>

JoshRosenstein:
Some current methods in the wild
[react-modal-hook](https://github.com/mpontus/react-modal-hook)-[Example](https://codesandbox.io/s/v8qy4w1j77)
[ProtonMail](https://github.com/ProtonMail/react-components/tree/master/components/modal)
[react-use-modal](https://github.com/wowlusitong/react-use-modal/tree/master)- [Example](https://react-use-modal.netlify.com/)
[neinteractiveliterature/intercode](https://github.com/neinteractiveliterature/intercode/tree/master/app/javascript/ModalDialogs)

src/Internals/Modal.tsx
```diff
@@ -8,47 +7,33 @@ export interface ModalOptions {
...
-              : this.state.options.body}
-          </ControlledModal>
-        )}
-      </>
-    )
-  }
+export const Modal = ({ children }: Props) => {
+  const [options, setOptions] = useState<State>({})
+
+  const openModal = useCallback((newoptions: ModalOptions) => {
```
stereobooster: `newOptions`

stereobooster:
@JoshRosenstein one small fix of variable name otherwise clear and precise 👍

src/Internals/Modal.tsx
JoshRosenstein: 🤦‍♂️ 😆

JoshRosenstein:
Should open an issue after this merges to continue discussion of potential options for better implementation.

stereobooster:
@TejasQ answering `useModal` question, but on the other side "modal" is a component (I know it is possible to return component from hook or create component outside of the current react tree) I prefer components for anything that have visual representation.

My question - this is not really a feature. I meant this PR doesn't add feature. Do we have (or want to add) `**Chore**` tag? For refactoring and other non project related things?

PR #943, opened 2019-02-28
**Feature:** TopbarSelect hooks

JoshRosenstein:
# Summary
- Added useEventListener
- Added useDebouncedCallback (replaced lodash/debounce)
- Added useDebounce (not used in module, but maybe good for any autocomplete or fetching state  )

# Related issue

# To be tested

Me

Tester 1

Tester 2


src/TopbarSelect/TopbarSelect.tsx
```diff
@@ -65,73 +61,48 @@ const TopbarSelectLabel = styled("p")`
...
-              <TopbarSelectValueSpan active={Boolean(selected)}>{selected}</TopbarSelectValueSpan>
-              <Icon color="color.text.lightest" name={isActive ? "CaretUp" : "CaretDown"} size={12} />
-            </TopbarSelectValue>
-          </TopbarSelectContainer>
-        )}
-      </ContextMenu>
-    )
-  }
+  const debouncedUpdateRenderedWidth = useDebouncedCallback(updateContainerWidth, 100, [
+    // containerRef.current,
```
TejasQ: ?

src/useDebouncedCallback/index.ts
```diff
@@ -0,0 +1,25 @@
+import { useCallback, useEffect, useRef } from "react"
+
+/**
+ * debouncedCB hook.
+ */
+export function useDebouncedCallback(callback: (...args: any[]) => void, delay: any, deps: React.DependencyList) {
```
TejasQ: You could destructure the import from `"react"` here:

```suggestion
export function useDebouncedCallback(callback: (...args: any[]) => void, delay: any, deps: DependencyList) {
```

since you're already using other named imports on the top.

src/useDebouncedCallback/index.ts
```diff
@@ -0,0 +1,25 @@
...
+  const debouncedFn = useCallback(callback, deps)
+
+  useEffect(
+    () => () => {
+      clearTimeout(fnTimeoutHandler.current!)
+    },
+    [],
+  )
+
+  return (...args: any[]) => {
```
TejasQ: We probably should type these hooks stricter.

src/useDebouncedCallback/index.ts
```diff
@@ -0,0 +1,25 @@
...
+    clearTimeout(fnTimeoutHandler.current!)
```
TejasQ: We probably should embrace the type system and avoid escape hatches (`!`)

src/useEventListener/index.ts
```diff
@@ -0,0 +1,18 @@
+import { useEffect } from "react"
+
+/**
+ * useEventListener
```
stereobooster: This comment doesn't help hear. Either add description or remove it

src/useEventListener/index.ts
```diff
@@ -0,0 +1,18 @@
...
+ */
+function useEventListener<K extends keyof GlobalEventHandlersEventMap>(
```
stereobooster: It is more like `useWindowEventListener`. And there is already `useWindowSize` hook, which is seems your use case

JoshRosenstein: Will change name. However, cant use useWindowSize because need to get the width of TopbarSelectContainer to pass to ContextMenu width.

src/useEventListener/index.ts
JoshRosenstein: Sorry- will add details! i quickly added these because that seemed to stop getting warning messages from styleguidist!

src/useDebouncedCallback/index.ts
JoshRosenstein: So im fairly new to typescript, but heres what i got after fiddling with it:
from useCallback:`  function useCallback<T extends (...args: any[]) => any>(callback: T, deps: DependencyList): T;`

```tsx
type AnyFunc=(...args: any[]) => any

 type ArgumentTypes<F extends AnyFunc> = F extends (...args: infer A) => any ? A : never;

export function useDebouncedCallback <T extends AnyFunc>(callback: T, delay: number, deps: React.DependencyList) {
  const fnTimeoutHandler = useRef<any>(null)
  const debouncedFn = useCallback(callback, deps)

  useEffect(
    () => () => {
      clearTimeout(fnTimeoutHandler.current)
    },
    [],
  )

  return (...args:ArgumentTypes<T>) => {
    clearTimeout(fnTimeoutHandler.current)
    fnTimeoutHandler.current = setTimeout(() => {
      debouncedFn(...args)
    }, delay)
  }
}
```

src/useDebouncedCallback/index.ts
JoshRosenstein: 👍

src/useDebouncedCallback/index.ts
JoshRosenstein: 👍

src/TopbarSelect/TopbarSelect.tsx
JoshRosenstein: ?

stereobooster:
```suggestion
```

src/useDebouncedCallback/index.ts
```diff
@@ -0,0 +1,30 @@
+import { DependencyList, useCallback, useEffect, useRef } from "react"
```
stereobooster: This code needs comments.

Trivial example of debounce (which is not useDebouncedCallback), but I wonder if all of this complexity is really needed is it essential?

```ts
import { useEffect } from 'react';

const useDebounce = (fn: () => any, ms: number = 0, args: Array<any> = []) => {
  useEffect(() => {
    let handle = setTimeout(fn.bind(null, args), ms);

    return () => {
      // if args change then clear timeout
      clearTimeout(handle);
    }
  }, args);
};

export default useDebounce;
```

src/TopbarSelect/TopbarSelect.tsx
TejasQ: 👆

src/useDebouncedCallback/index.ts
TejasQ: @stereobooster please advise here about if/how the types can be improved. Also @mpotomin.

src/useDebounce/README.md
```diff
@@ -0,0 +1,21 @@
+## Usage
+
+useDebounce- debounces values
```
TejasQ: Please add a little more explanation here: this is a `README` of an open source library that we'd like to be inclusive and approachable, even to beginners. 😄

src/useDebounce/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+```jsx
+//import useDebounce from "@operational/ui/src/useDebounce"
```
TejasQ: Please remove this if it isn't used, or export it from `index.ts` if it is.

JoshRosenstein: Did not know if you had to import it for the new styleguidist api.
[Here says i dont](https://react-styleguidist.js.org/docs/documenting.html#writing-code-examples)
. will remove.

src/useDebounce/README.md
JoshRosenstein: 🤣 now your really asking me to work lol. But yes that is very bad, i cant stop laughing, imagining when i started learning and i went to someones docs and saw that 🤣

src/TopbarSelect/TopbarSelect.tsx
JoshRosenstein: 🔎 😮 😞

src/useDebouncedCallback/index.ts
JoshRosenstein: Heres Example of all three, I cant get that one 🔼  to act like the other two, But this does show the differences. 

[Edit useDebounce Example](https://codesandbox.io/s/zwnw82mrpx?fontsize=14)

src/useDebouncedCallback/index.ts
mpotomin: I'd go with smth like this for the useDebouncedCallback signature:

```tsx
export function useDebouncedCallback<T extends [], R>(callback: (...args: T) => R, delay: number, deps: DependencyList,) {}
```

`useCallback` forces us to cast the `callback` to `(...args: any[]) => any` though 😠 

```tsx
const debouncedFn = useCallback(callback as (...args: any[]) => any, deps)
```

mpotomin: Also, how about making `delay: number` optional and using `window.requestAnimationFrame` instead of `setTimeout` for debouncing if no `delay` value has been passed? Will give us the snappiest animation and eliminate the need for thinking of the optimal delay every time.

src/useDebounce/README.md
```diff
@@ -0,0 +1,22 @@
+Creates a debounced [state,setState] method that delays setting the new state until after desired milliseconds have elapsed since the last time an attempt to update the state was made.
```
mpotomin: Let's call this `useDebouncedState` because `useDebounce` sounds like it debounces a function. 🤔 That way the `useDebouncedCallback` can be renamed to `useDebounce`, which I would expect.

src/useDebouncedCallback/index.ts
mpotomin: I don't quite understand: from the CodeSandbox, they look and work identically unless I missed something? What is the intended behavior here?

src/useWindowEventListener/README.md
```diff
@@ -0,0 +1,11 @@
+## Usage
+
+TODO
```
mpotomin: Let's add some documentation here.

src/useDebouncedCallback/index.ts
JoshRosenstein: Yes i was trying to show how they can all work identically, but could not get @stereobooster - example to be like the other two. By making them work identically, you can see how their api is different and how they may differ on how you would use the two. For example in order for useDebouncedCallback to act like useDebounced, the CB to debounce would just be a simple 
```js
const debouncedFunction = useDebouncedCallback(
    value => {
      setValue(value);
    },
    2000,
    []
  )
```

But in reality, if the CB is that simple, just use `useDebounce`, where in any other situation, where you need a CB to do some heavy calculation or checking prior to state changes, then you would need `useDebouncedCallback`. OR using `useDebouncedCallback` as an HOC for other hooks build-in methods that you need debounced. Heres an example where we have a useHistory hook, and for some reason, we want to debounce the log method:
[Edit useDebouncedCallback Ex 2](https://codesandbox.io/s/34nl020l41?fontsize=14)

In addition, since the  log method requires an input,   @stereobooster useDebounce example above will not work correctly.

src/useDebounce/README.md
JoshRosenstein: yeah thats what I originally was thinking, but from the hooks naming conventions and inputs,  `useDebouncedState`- IMO, sounds like it should return everything `useState` would return. Hence, would be more similar to: [THIS](https://github.com/tolotrasmile/use-debounce-state/blob/master/index.js)

src/useWindowEventListener/README.md
JoshRosenstein: whoops missed that one! thanks

JoshRosenstein:
@TejasQ  @mpotomin @stereobooster  - After testing TopbarSelect- using the example in TopBar readme, on dev-server, I caught a bug where the width was not getting passed to the ContextMenu, unless the window got resized. After playing around with it, I came to the conclusion that its unnecessary, to check the width on window resize. So i removed the hooks. If you agree that these methods are not needed, please let me know if I should remove all 3 of the new hooks from this PR, or not.

src/useWindowEventListener/index.ts
```diff
@@ -3,11 +3,12 @@ import { useEffect } from "react"
...
  * Hook version of window.addEventListener.
  */
-function useWindowEventListener<K extends keyof GlobalEventHandlersEventMap>(
+
+function useWindowEventListener<K extends keyof WindowEventMap>(
   type: K,
-  listener: EventListenerOrEventListenerObject,
-  addOptions?: AddEventListenerOptions,
-  removeOptions?: EventListenerOptions,
+  listener: (ev: WindowEventMap[K]) => any,
```
TejasQ: Does

`(ev: WindowEventMap[K]) => void` instead of `=> any` work in this case?

src/useWindowEventListener/index.ts
```diff
@@ -3,11 +3,12 @@ import { useEffect } from "react"
...
+  addOptions?: boolean | AddEventListenerOptions,
+  removeOptions?: boolean | EventListenerOptions,
```
TejasQ: Why the booleans?

JoshRosenstein: just used the types from [lib.dom.ts](https://github.com/Microsoft/TypeScript/blob/a5f93ef7a4fb1ce1e1e19faea27f57c749abe2f3/lib/lib.dom.d.ts#L18255)

src/useDebounce/README.md
TejasQ:
```suggestion
Creates a debounced [state, setState] method that delays setting the new state until after desired milliseconds have elapsed since the last time an attempt to update the state was made.
```

TejasQ: And why can't we do that?

TejasQ: Is ^ even correct? Because it's a different usage [here](https://github.com/contiamo/operational-ui/pull/943/files#diff-6869b14f7b5b80edf49baaf18137eb34R10)...

src/useDebouncedCallback/README.md
```diff
@@ -0,0 +1,26 @@
+## Usage
+
+Similar to UseDebounce but a debounced Callback is returned and you can do prelim calculations prior to setting the Value.
+
+```jsx
+//import useDebouncedCallbacks from '@operational/ui/src/useDebouncedCallbacks';
```
TejasQ: ? Why is this commented out?

src/useDebouncedCallback/README.md
```diff
@@ -0,0 +1,26 @@
+## Usage
+
+Similar to UseDebounce but a debounced Callback is returned and you can do prelim calculations prior to setting the Value.
```
TejasQ:
```suggestion
Similar to UseDebounce but a debounced callback is returned and you can do preliminary calculations prior to setting the value.
```

src/useDebouncedCallback/index.ts
```diff
@@ -0,0 +1,30 @@
...
+
+type AnyFunc = (...args: any[]) => any
+
+type ArgumentTypes<F extends AnyFunc> = F extends (...args: infer A) => any ? A : never
+
+/**
+ * useDebounced Callback wraps a callback within useCallback and returns
```
TejasQ: `and returns` what?

src/useWindowEventListener/index.ts
TejasQ: @JoshRosenstein

TejasQ:
@JoshRosenstein 

> please let me know if I should remove all 3 of the new hooks from this PR, or not.

It's tricky because:

1. We would rather your work not go to waste, so let's keep them, but
1. Less is _always_ more/better in code.

Let's remove them. We still have [the](https://github.com/contiamo/operational-ui/pull/943/commits/42f329a6f9d03f15afd6abbed0f9c97a5f8f8355) [individual](https://github.com/contiamo/operational-ui/pull/943/commits/f776e597260da0bb609ed6d874be65f5c323f32f) [commits](https://github.com/contiamo/operational-ui/pull/943/commits/d9c72594e4ee5de488a31730d533cd6078df52e0) that add them in case we need them later. 😄

src/TopbarSelect/TopbarSelect.tsx
```diff
@@ -65,73 +59,39 @@ const TopbarSelectLabel = styled("p")`
...
-
-  public componentWillUnmount() {
-    window.removeEventListener("resize", this.handleResize)
-  }
-
-  public componentDidUpdate() {
-    this.updateRenderedWidth()
-  }
+const TopbarSelect = ({ label, selected, items, onChange, ...props }: TopbarSelectProps) => {
+  const [containerWidth, setContainerWidth] = useState(0)
```
stereobooster: consider `useSize` hook API instead. For inspiration:

- https://github.com/infodusha/react-hook-size
- https://github.com/rehooks/component-size

TejasQ: I agree with this, but it is not a show stopper.

PR #945, opened 2019-03-01
**Fix:** width auto, for some reason, collapses width of textarea inside of it

stereobooster:
# Summary

before:
screenshot 2019-03-01 at 14 43 54

after:
screenshot 2019-03-01 at 14 44 13

# Related issue

N/A

# To be tested

Me

Tester 1

Tester 2

### DIFF
--- a/src/utils/mixins.ts
+++ b/src/utils/mixins.ts
@@ -16,7 +16,7 @@ export const Label = styled("label")<{ fullWidth?: boolean; left?: boolean }>(({
   display: "inline-block",
   position: "relative",
   maxWidth: fullWidth ? "none" : "360px",
-  width: fullWidth ? "100%" : "auto",
+  width: "100%",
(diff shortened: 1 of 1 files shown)


mpotomin:
There are some unexpected regressions/changes to `Input`s and other side effects by this PR. @TejasQ should have already fixed this in #922? 

screenshot 2019-03-01 at 16 13 07 2

stereobooster:
I wouldn't call it unwanted behaviour. Initial problem which I tried to solve is that textareas and inputs have different width. See screenshot before. And after all inputs and textareas have the same width (except size modifiers, like fullwidth)

TejasQ:
@stereobooster let's base this pull request on #922

TejasQ:
@stereobooster would you like to:

a) rebase this on #922 
b) fix this directly in #922 

?

TejasQ:
I suggest we close this in favor of #922

Issue #936, opened 2019-02-23
Broken link on documentation site


brentbaum:
Visiting the project home page and clicking "Docs" doesn't work (the browser remains on the home page), so I can't see your beautiful component library.

TejasQ:
Hey Brent! Thanks for the report! We're currently working on a comprehensive overhaul of our docs. We expect to ship it this Tuesday. 😄

JoshRosenstein:
@TejasQ - will you be bumping styleguidist to v9? Wonder since you need to now explicitly import the components within the examples, if that help out performance or not. I think the main issue with current setup is that everything is loaded in at once and navigation is only anchors. where as [storybook](https://baseui.design/components/input/) or [doz](https://rosenstein.io/emotion-icons/usage-props)- 
can separate out by pages. Not too familiar with styleguidist.

after typing that ^ i dug into their api and found [pagepersection](https://github.com/styleguidist/react-styleguidist/blob/master/docs/Configuration.md#pagepersection)- so that might help. The only issue now would having to manually list all files and sections in the config

TejasQ:
@JoshRosenstein - that's really cool. I'm actually chipping away at a styleguidist clone that is more suited to our needs. It's coming along nicely but still has some rough edges (read: it's ready but doesn't quite get hot module replacement correctly). 

Great finding the pagepersection though – that could be a nice interim solution!

JoshRosenstein:
@TejasQ - I was just playing around with it. I stripped away the custom components besides the wrapper- just to test out the pagepersection. It is so much faster! haha.
Also found out you dont need to type all components out, each component within the section is its own page. Heres my [branch ](https://github.com/JoshRosenstein/operational-ui/tree/docs-withPages) i was testing with

![Alt Text](https://i.imgur.com/xrkpRnd.gif)

TejasQ:
@JoshRosenstein that's incredible! Let's merge it, but please keep the markdown files untouched and let's change only the styleguide since we're gonna be moving it over soon.

JoshRosenstein:
Merge into your update-docs branch? well theres really not much I did but comment out a few lines in the config, and upgraded deps. I wouldnt have been able to test that out if you didnt prep those READMEs 😉

TejasQ:
Ah, thanks. Maybe do a fork of my branch with the READMEs + your changes and let's merge to `master`?

JoshRosenstein:
[...]

Issue #944, opened 2019-03-01
OperationalContext - seperate logic from presentaion


JoshRosenstein:
After realizing i couldn't use some components without having them wrapped in OperationalUI and then reading over [useOperationalContext Feature](https://github.com/contiamo/operational-ui/pull/911),

Whenever you choose to refactor some of the components like (Button, Breadcrumb,Logo etc), I think it would be best to separate out the logic from UI components. I dont need to preach the reasons why, but i thought after seeing that PR, that would be a good time to do it when those get refactored.

Also you can even take it a step further and breakdown and export everything like [Uber/baseui](https://github.com/uber-web/baseui/tree/master/src/button-group) 😄

TejasQ:
Great idea! Do you have a little bit more concrete information about which components would be touched?

JoshRosenstein:
- [Breadcrumb](https://github.com/contiamo/operational-ui/blob/master/src/Breadcrumb/Breadcrumb.tsx#L41)
- [Logo](https://github.com/contiamo/operational-ui/blob/master/src/Logo/Logo.tsx#L68)
- [Input.Button](https://github.com/contiamo/operational-ui/blob/master/src/Input/Input.Button.tsx)
- [SimpleLink](https://github.com/contiamo/operational-ui/blob/master/src/SimpleLink/SimpleLink.tsx)
- [SidenavItem](https://github.com/contiamo/operational-ui/blob/master/src/SidenavItem/SidenavItem.tsx)
- [Button](https://github.com/contiamo/operational-ui/blob/master/src/Button/Button.tsx)
- [SidenavHeader](https://github.com/contiamo/operational-ui/blob/master/src/SidenavHeader/SidenavHeader.tsx)

- off topic but just 👀simpleLink, and wonder if there are more components like this where you should pull the container out from under and make use of the (as-prop)[https://emotion.sh/docs/styled#as-prop]

TejasQ:
This is a great idea! It's actually a continuation (IMO) of #911. Let's do it and HOOKS ALL THE THINGS! Wanna collab?

JoshRosenstein:
Yeah saw that and why i brought it up as it would be a perfect time to knock out 2 birds with one stone. I gotta get used to correctly linking my refs haha.

TejasQ:
I think this is closable since we merged #911, right, @JoshRosenstein?

PR #948, opened 2019-03-02
**Fix:** Docs

JoshRosenstein:
# Summary
- All current READMES in this branch render correctly
- Got all custom components & splash screen working correctly
- Docs configuration altered to use pages, which allows to remove some bulk from custom styleguidist components
- Big bug fix was disabling useURLState example from rendering- was effecting sites hash.
- react-styleguidist & react-docgen-typescript upgrades latest ( no commits show package.json)
- react-waypoint - removed

- Need to merge with master than do another test run/debug.

# Related issue

[#936](https://github.com/contiamo/operational-ui/issues/936)
# To be tested

Me

Tester 1

Tester 2


JoshRosenstein:
One small bug i noticed, the search input keeps getting focused when navigating.

 alt text

mpotomin:
@JoshRosenstein this is **amazing**! 

Let's add missing examples that we commented out (`useWindowSize`) and resolve the conflict with `Textarea` and this should be good to go!

JoshRosenstein:
> @JoshRosenstein this is **amazing**!
> 
> Let's add missing examples that we commented out (`useWindowSize`) and resolve the conflict with `Textarea` and this should be good to go!

You were actually referring to the commented out example 'useURLState' correct? We cant enable any examples for that one since it alters the URL

styleguide/TableOfContentsRenderer.tsx
```diff
@@ -1,29 +1,56 @@
...
+}))
+
+const Container = styled("div")(({ theme }) => ({
+  label: "sidenavheader",
+  textDecoration: "none",
+  width: "100%",
+  borderBottom: "1px solid",
+  borderBottomColor: theme.color.separators.default,
+}))
+
```
JoshRosenstein: Passing Search input into a render prop  causes entire parent to re-render and loses focus of the input. This is temporary  and relates to [#944](https://github.com/contiamo/operational-ui/issues/944)- where its a good idea to export all individual UI components.

TejasQ:
@JoshRosenstein 

> You were actually referring to the commented out example 'useURLState' correct? We cant enable any examples for that one since it alters the URL

I don't understand though – it only alters the URL by query param. This doesn't navigate away or anything, right? So why can't we use it?

JoshRosenstein:
@TejasQ - your right- i think when i was fiddling with the routing api, there was an issue with setting the sideNav item to isActive- so i turned it off- but never tested it again after finishing it up. Just tested now, all good.

TejasQ:
Nice! @JoshRosenstein [go here](https://deploy-preview-948--operational-ui.netlify.com/), click "Docs" and see that the default route is empty. 🤔 

Can we fix?

JoshRosenstein:
Did a temporarily fix, but a better solution would have a intro MD that styleguidist can use as the root .

styleguide/StyleGuideRenderer.tsx
```diff
@@ -38,7 +38,15 @@ const SplashScreen = ({ setSplash }: { setSplash: React.Dispatch<React.SetStateA
     title="Operational UI"
     actions={
       <>
-        <Button onClick={() => setSplash(false)}>Docs</Button>
+        <Button
+          onClick={() => {
+            setSplash(false)
+            // @ts-ignore
```
TejasQ: Please _always_ add an explanation to any statements like this.

`// @ts-ignore because window can't be extended` or similar.

Also, you should be reassigning `window.location.href`, which likely won't make TypeScript complain.

TejasQ:
@JoshRosenstein you win the internet.

TejasQ:
One more rebase and CI will be green 💚  @JoshRosenstein

styleguide/TableOfContentsRenderer.tsx
```diff
@@ -1,30 +1,56 @@
...
+import React from "react"
+import { Input, Sidenav, styled } from "../src"
 
 export interface TableOfContentsRendererProps {
   children: React.ReactElement<any>
   searchTerm: string
   onSearchTermChange: (newVal: string) => void
 }
 
+// Dont pass Input in RenderProps-(<SideNavHeader label={<Input/>} />) makes Compnent rerender and loses input focus
```
stereobooster:
```suggestion
// Don't pass Input in RenderProps-(<SideNavHeader label={<Input/>} />) makes Component rerender and lose input focus
```

JoshRosenstein:
Tried fixing Layout Readme for third time- Only needs to remove trailing comma from imports- for some reason when i commit the changes- it doesnt show up as a file change, i think its due to lint-staged maybe? and trailingComma:all in prettier.

JoshRosenstein:
With the last commit [Add getExampleFilename](https://github.com/contiamo/operational-ui/pull/948/commits/f411991ba0ee8168383505384d5569d6a5893335), I can revert [Move Typography Components to their own folders](https://github.com/contiamo/operational-ui/pull/948/commits/aba1f7ba2323c590742e626d95ceafb1fb644062) back to its original structure if need be.

This also fixes Components like ChildTree, and Tree having the same README file loaded in.
Inorder to do this we need  `skipComponentsWithoutExample: true`. That way a component page in the docs is only created if:
- The filename === Folder name- defaults to README.md
- Filename != folderName but a MD file exists with the same name.

fabien0102:
Wooow amazing PR!😍 Just a little comment about useURLState, normally you can give fake windows dependencies, so we can handle the side effect clean for the example 😉 like the "Page tab with routing" example (if my memory is correct ^^)

Note: Since I'm on my phone, my comment is maybe totally out of scope 😄 but can also help!

JoshRosenstein:
@fabien0102 - Yeah in order to do something like that would be something like...

```jsx
import {useState} from 'react'
import { Button, useURLState } from "@operational/components"
const countDecoder = i => (isNaN(+i) ? undefined : +i)

const Counter = () => {
const [fakeWindowLocation,setFakeWindowLocation]=useState({...this.window.location,urlChange:''})
  const getSearch= () => fakeWindowLocation.search
  const getHash= () => fakeWindowLocation.hash
  const getPathname= () => fakeWindowLocation.pathname
const replaceState= (data,title,urlChange) => setFakeWindowLocation({...fakeWindowLocation,urlChange}) 
  // this is sync with `?count={number}`
  // you can try to reload the page to see that the state is retain
  const [count, setCount] = useURLState("count", 0, countDecoder,{
        getSearch,
        replaceState,
        getPathname,
        getHash,
      })

  return (
    <>
      <h3>{count}</h3>
      <Button onClick={() => setCount(count + 1)} children="up" />
      <Button onClick={() => setCount(count - 1)} children="down" />
      <pre>{JSON.stringify(fakeWindowLocation,null,2)}</pre>
    </>
  )
}
;<Counter />
```


However to get it perfect- not using 'urlChange' variable, and to actually replace the search parameter will take a bit more work to parsing the third arg passed through `replaceState`.

But i think we should wait on that and get this pushed through- before any other commits that change examples gets pushed on to master, and causing more merging conflicts.

fabien0102:
@JoshRosenstein let's keep it like this for now, and maybe introduce a fakeBrowser component just for the doc 😀

src/ContextMenu/README.md
```diff
@@ -2,7 +2,20 @@ Context menus are nested menus that can expand from anywhere on a page. Their us
 
 ### Basic Usage
 
+```ts
```
TejasQ: Let's remove this snippet.

JoshRosenstein:
> Great. Couple fixes:
> 
> 1. There's a duplicate snippet for `ContextMenu`'s readme
> 2. The welcome page is broken:
> 
> image
> 
> To fix, please wrap it in a card with a `title`.

Welcome Page is not broken, its just that its getting loaded as a content and not a component in the style guide config, and so it doesn't pass through the `ReactComponentRenderer`, i couldnt find in the docs where raw MD files can be wrapped like this. I'll try tinkering a bit more

JoshRosenstein:
All working [here ](https://deploy-preview-948--operational-ui.netlify.com/#/Introduction).

Will merge in changes from master and start fixing all the MD's that need fixin from old stylguide api.

fabien0102:
image

You can remove the greenkeeper badge from the REAME.md, we are not using it anymore (it was too annoying for us 😅 )

fabien0102:
Looks really good! 💯 Just a little question/remark, when I click to `useWindowSize`, the page totally reload 😕 It's the only example with this behaviour, so I'm curious 😅

TejasQ:
@fabien0102 it's not when you navigate _to_ `useWindowSize`, but when you navigate _away_ from `useURLState` – he cleans up the query params. This is fine in my opinion.

JoshRosenstein:
@fabien0102 - might not be `useWindowSize`, its probably due to `useUrlState`- you might have navigated to `useUrlState` then to `useWindowSize`- and since useUrlState changes the url

JoshRosenstein:
😭  lol

PR #949, opened 2019-03-05
**Breaking:** Add columns param to CardColumns

stereobooster:
# Summary

Add `columns` param to the CardColumns component, so you can specify how many columns there can be inside the CardColumns.
- if there are not enough columns, it will show empty space for missing columns
- it there more columns than specified it will wrap columns (create rows)
- fixes bug with wrong margin for case when there is only one column

# Related issue

N/A

# To be tested

Me

Tester 1

Tester 2

### DIFF
--- a/src/CardColumns/CardColumns.tsx
+++ b/src/CardColumns/CardColumns.tsx
@@ -1,13 +1,18 @@
 import * as React from "react"
 import styled from "../utils/styled"
 
-export const CardColumns = styled("div")<React.Props<{}>>(({ children, theme }) => ({
-  display: "flex",
-  flexWrap: "wrap",
-  margin: -(theme.space.element / 2),
-  "& > *": {
-    flexBasis: `${React.Children.count(children)}%`,
-  },
-}))
+export const CardColumns = styled("div")<React.Props<{}> & { columns?: number }>(({ children, theme, columns }) => {
+  columns = columns === undefined ? React.Children.count(children) : columns
+  return {
+    display: "flex",
+    flexWrap: "wrap",
+    margin: -(theme.space.element / 2),
+    "& > *": {
+      // !important required for flexColumn
+      flexBasis: `calc(100% / ${columns}) !important`,
+      maxWidth: `calc(100% / ${columns})`,
+    },
+  }
+})
 
 export default CardColumns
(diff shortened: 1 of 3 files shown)


src/CardColumns/CardColumns.tsx
```diff
@@ -1,13 +1,18 @@
...
-  },
-}))
+export const CardColumns = styled("div")<React.Props<{}> & { columns?: number }>(({ children, theme, columns }) => {
+  columns = columns === undefined ? React.Children.count(children) : columns
+  return {
+    display: "flex",
+    flexWrap: "wrap",
+    margin: -(theme.space.element / 2),
+    "& > *": {
+      // !important required for flexColumn
```
stereobooster: haven't found better solution for this so far

stereobooster:
@TejasQ fixed

TejasQ:
Awesome.

Let's add a note here about wrapping and how it _will wrap_ despite the `columns` prop number

image

src/CardColumns/README.md
```diff
@@ -94,34 +94,41 @@ The `CardColumns` component is used as a wrapper around groups of `CardColumn` c
...
+      <Chip>production</Chip>
     </CardColumn>
   </CardColumns>
-  <CardColumns>
-    <CardColumn>
-      <Button color="primary">Send Request</Button>
+</Card>
+```
+
+If there is not enough space to place required number of columns, columns will wrap
```
stereobooster: @TejasQ do you like wording here?

TejasQ:
```suggestion
Columns will always wrap and overflow if their containing element is not wide enough for them.
```

src/CardColumns/CardColumns.tsx
```diff
@@ -1,13 +1,22 @@
...
-export const CardColumns = styled("div")<React.Props<{}>>(({ children, theme }) => ({
-  display: "flex",
-  flexWrap: "wrap",
-  margin: -(theme.space.element / 2),
-  "& > *": {
-    flexBasis: `${React.Children.count(children)}%`,
-  },
-}))
+export type CardColumnsProps = React.Props<{}> & {
+  /** Force exact number of columns */
```
TejasQ:
```suggestion
  /** Render an exact number of columns */
```

src/CardColumns/README.md
```diff
@@ -94,34 +94,41 @@ The `CardColumns` component is used as a wrapper around groups of `CardColumn` c
 </Card>
 ```
 
-### With flexColumn
+### With columns param
```
TejasQ:
```suggestion
### With an exact number of columns
```

TejasQ:
The example looks a little broken. 😄 

uff

stereobooster:
@TejasQ yes, but it exactly demonstrates the edge case.

PR #950, opened 2019-03-06
  **Fix:** Improve tree component

stereobooster:
# Summary

Need to fix following styling issues:

- [x] cursor
- [x] margin

screenshot 2019-03-06 at 16 02 44

---

Drag cursor is not testable in styleguidist, because react-beautiful-dnd sets cursor on the body, but it's get overridden by some of styleguidist's classes

screenshot 2019-03-06 at 17 12 02

# Related issue

N/A

# To be tested

Me

Tester 1

Tester 2


src/Tree/TreeItem.tsx
```diff
@@ -0,0 +1,96 @@
...
+  :hover {
+    background-color: rgba(0, 0, 0, 0.1);
+  }
+`
+
+const TreeIcon = styled(Icon)`
+  margin-right: ${({ theme }) => theme.space.base}px;
+`
+
+interface TreeItemProps {
```
TejasQ: Can we move this to the top? It's convention (sort of) to have it directly under the imports.

PR #953, opened 2019-03-07
**Fix:** it was not possible to disable Tree component inside CardSection

stereobooster:
# Summary

screenshot 2019-03-07 at 13 45 45

# Related issue

N/A

# To be tested

Me

Tester 1

Tester 2

### DIFF
--- a/src/CardSection/CardSection.tsx
+++ b/src/CardSection/CardSection.tsx
@@ -43,10 +43,11 @@ export type OverlayType = "noOverlay" | "disabled" | DragAndDropFeedback
  * The flex rule only kicks in when the parent is flex-positioned in case
  * sections are stacked horizontally.
  */
-const Container = styled("div")`
+const Container = styled("div")<{ disabled?: boolean }>`
   flex: 1 1;
   display: block;
   position: relative;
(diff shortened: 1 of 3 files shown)


src/CardSection/CardSection.tsx
```diff
@@ -43,10 +43,11 @@ export type OverlayType = "noOverlay" | "disabled" | DragAndDropFeedback
  * The flex rule only kicks in when the parent is flex-positioned in case
  * sections are stacked horizontally.
  */
-const Container = styled("div")`
+const Container = styled("div")<{ disabled?: boolean }>`
```
stereobooster: Maybe we need to add `aria-disabled`

TejasQ: Yes, but not here.

src/CardSection/CardSection.tsx
```diff
@@ -161,7 +162,7 @@ const CardSection: React.SFC<CardSectionProps> = ({
   forceToggleHoverStyles,
   ...props
 }) => (
-  <Container {...props}>
+  <Container {...props} disabled={disabled}>
```
TejasQ:
```suggestion
  <Container {...props} disabled={disabled} aria-disabled={disabled}>
```

PR #972, opened 2019-04-01
**Breaking:** More flexible drag & drop on Tree

fabien0102:
# Summary

To be able to have the best drag and drop experience (at least a better one 😅) we need to expose `react-beautiful-dnd` props into our `Tree` component.

The idea is to provide flexibility here, so we can define `Droppable` and `Draggable` props as we use `react-beautiful-dnd` directly.

This PR was tested against my local usecase and it works as expected. 

## Example of usage

```tsx
<Tree
  droppableProps={{ droppableId: getDroppableId("aggregateMeasure") }}
  trees={state.aggregate.query.measures.map(m => ({
    label: m,
    key: m,
    color: PslColor.measure,
    tag: "M",
    draggableProps: { draggableId: getDraggableId("aggregateMeasure", m) },
  }))}
/>
```

# Related issue

# To be tested

Me

Tester 1

Tester 2

### DIFF
--- a/src/Tree/Tree.tsx
+++ b/src/Tree/Tree.tsx
@@ -1,6 +1,6 @@
 import { Omit } from "emotion-theming/types/helper"
 import * as React from "react"
-import { Draggable, DraggableProps, Droppable, DroppableProps } from "react-beautiful-dnd"
+import { Draggable, DraggableProps, Droppable, DroppableProps, DroppableStateSnapshot } from "react-beautiful-dnd"
 import styled from "../utils/styled"
 import ChildTree from "./ChildTree"
 
@@ -33,6 +33,7 @@ export type Tree = TreeWithChildren | TreeWithoutChildren
 export interface TreeProps {
   trees: Tree[]
   droppableProps?: Omit<DroppableProps, "children">
+  placeholder?: React.ComponentType<DroppableStateSnapshot>
 }
 
 const Container = styled("div")`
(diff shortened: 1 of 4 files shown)


src/Tree/Tree.tsx
```diff
@@ -16,11 +16,24 @@ export interface Tree {
...
+}
+
+export type Tree = TreeWithChildren | TreeWithoutChildren
+
 export interface TreeProps {
   trees: Tree[]
-  draggable?: boolean
-  onDrop?: (result: DropResult, provided: ResponderProvided) => void
-  id?: string
+  droppableProps?: Omit<DroppableProps, "children">
```
TejasQ: I would prefer `droppableOptions` because `props` too closely relates to React components. This is not a React component.

src/Tree/Tree.tsx
```diff
@@ -16,11 +16,24 @@ export interface Tree {
...
+  placeholder?: React.ComponentType<DroppableStateSnapshot>
```
TejasQ: What in the world is this? Why can't we just have a simple `React.ReactNode` here? 🤔

src/Tree/TreeItem.tsx
```diff
@@ -63,6 +64,7 @@ interface TreeItemProps {
   label: string
   tag?: string
   color?: string
+  cursor?: string
```
stereobooster:
```suggestion
  cursor?: CursorProperty
```

fabien0102: It's part of `csstype` package, so we need to add this dep in our package.json to avoid implicit dependency

PR #979, opened 2019-04-09
**Fix:** [SidenavHeader] Hide cursor pointer if no onClick

TejasQ:
Currently, `SidenavHeader`s have `cursor: pointer` even if no click handler is set. Let's remove it to improve UX.

### DIFF
--- a/src/SidenavHeader/SidenavHeader.tsx
+++ b/src/SidenavHeader/SidenavHeader.tsx
@@ -27,6 +27,7 @@ export interface SidenavHeaderProps extends DefaultProps {
   /** Close handler (via chevron button on the top right) */
   onClose?: () => void
   children?: React.ReactNode
+  /** Should the header be small? */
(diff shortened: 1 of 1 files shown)


src/SidenavHeader/SidenavHeader.tsx
```diff
@@ -39,22 +40,25 @@ const makeContainer = (type: "link" | "block") =>
...
-  }),
-)
+const Content = styled("div")<{
+  onClick: SidenavHeaderProps["onClick"]
+  isCondensed: boolean
+  isActive: boolean
+  compact: SidenavHeaderProps["compact"]
+}>(({ theme, onClick, isCondensed, compact, isActive }) => ({
+  textDecoration: "none",
+  cursor: Boolean(onClick) ? "pointer" : "initial",
```
TejasQ: Just changed this

PR #982, opened 2019-04-17
Move to Operational Styleguide

TejasQ:
This PR uses a new typescript-first styleguide.


stereobooster:
```
Cannot find module 'react-beautiful-dnd'
```

TejasQ:
@JoshRosenstein if you'd like a fun challenge, you could fix each broken example in this branch. 😍

JoshRosenstein:
@TejasQ  I cant even get it to start. 
First it couldnt find components,so had to build:package, now its giving me
```sh
operational-ui\node_modules\@operational\styleguide\lib\components\Wrapper.tsx:1:25: Cannot resolve dependency '..........styleguideWrapper.tsx'
```
Kinda hard to diagnose @operational\styleguide, I would usually do yarn:link on both repos but im guessing `@operational\styleguide` is private?

JoshRosenstein:
Since i dont have access to @operational\styleguide, I started fiddling with [this a bit ](https://rosenstein.io/operational-docz/). 😆

TejasQ:
@JoshRosenstein, it's not private? https://www.npmjs.com/package/@operational/styleguide

Did you `yarn` after you checked out the branch?

WaitNkill:
No I have it in my deps and in node_modules. I meant the source code is private. The error I’m getting is within that package, and it be easier to debug by downloading the source and yarn link the dep rather than compiled code from npm.

JoshRosenstein:
No I have it in my deps and in node_modules. I meant the source code is private. The error I’m getting is within that package, and it be easier to debug by downloading the source and yarn link the dep rather than compiled code from npm.

JoshRosenstein:
Fixed my first issue by changing how 
`operational-ui\node_modules\@operational\styleguide\lib\components\Wrapper.tsx`
gets generated.  
was returning `module.exports = require("..\..\..\..\..\styleguide/Wrapper.tsx")` instead of `module.exports = require("../../../../../styleguide/Wrapper.tsx")`
Then I got a node memory error during the building phase. So I edited the operational-styleguide.cmd to increase memory limit
```sh
@IF EXIST "%~dp0\node.exe" (
  "%~dp0\node.exe" "--max_old_space_size=8192" "%~dp0\..\@operational\styleguide\lib\cli.js" %*
) ELSE (
  @SETLOCAL
  @SET PATHEXT=%PATHEXT:;.JS;=;%
  node  "--max_old_space_size=8192" "%~dp0\..\@operational\styleguide\lib\cli.js" %*
)
```

Now Im able to start debugging these examples! 😃

TejasQ:
Seems I wrote a memory leak. I'm such a fool. Thanks for working around my giant hack.

JoshRosenstein:
Also trying to debug some other issues in  `@operational/styleguide`. Since im working in windows, the cli is not working as expected. For example in generateTypedefs:
```js
    const getTypedefs = () => {
        const components = glob_1.default
            .sync(path_1.join(parentPath, getTypesDir(), "**/*.d.ts"))
            .filter(path => path.includes(path_1.join(parentPath, getTypesDir())));
```

All components end up being filtered out because glob returns unix style paths and path.join returns windows style. So the generated typeMap does not include any @operational/components definitions. 

So I would either replace path with [Upath](https://www.npmjs.com/package/upath) or just wrap any path variables in a utility like [slash](https://github.com/sindresorhus/slash). Also any path manipulations should be done with path ie. 
```js
      const indexPath = indexFile // /Users/tejaskumar/Sites/project/whatsup/lib/src/index.d.ts
            .split("/")
            .slice(0, -1)
            .join("/"); // /Users/tejaskumar/Sites/project/whatsup/lib/src/

```
Safer:
```js
   const indexPath =path.join(indexFile,'..')
```

TejasQ:
Great catch! Let's collab on the source code!

JoshRosenstein:
> Great catch! Let's collab on the source code!

I dont have access to source code. Thats what i meant above when i was asking if it was private aka the github repo, because its a pain debugging the compiled source within node_modules. 🤕

TejasQ:
Fixed everything except the `"react-beautiful-dnd"` issue. This seems like it will be a bigger topic with dependency management. 🤦‍♂

TejasQ:
@contiamo/frontend do y'all think we can merge this without `"react-beautiful-dnd"` or should I spend more time there?

TejasQ:
This needs a _lot_ of optimizations. This is the status:

- [x] make it work
- [ ] make it right
- [ ] make it fast

There needs to be some adjustments to be more correct, and we currently have a 3.5MB bundle, which can be optimized and code split better. I have an abstract plan. 

My question is business value.

Does it make sense to invest time in this right now? Perfectly happy to focus on the product, close this and come back to it later because styleguidist already gives us great value.

Issue #984, opened 2019-04-18
Button Remounted on every render


JoshRosenstein:
**Problem:** Button currently creates the Container within itself which is not optimal as it creates a new component on every render.
https://github.com/contiamo/operational-ui/blob/402249ae501e468c5b3ad8ccd5fc581a70836b22/src/Button/Button.tsx#L86
**Suggestion:** extract container out, utilized emotions `as` prop to render a anchor if `to` prop is passed. 

**Origin:** #468 

**Example:**
current button vs button with suggested changes can be viewed here:
[![Edit Operational-ui Button Issue](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/8lqrj509oj?expanddevtools=1&fontsize=14)
(updated the example to also show how the Test in the comments below works)

PR on standby.

------
### Edit
 Extracting out Container also initializes a step for completing #944 

Searched for some components with this similar issue and found:
https://github.com/contiamo/operational-ui/blob/d72016c11638a4dd751b2b0396c13470951644a7/src/SidenavItem/SidenavItem.tsx#L120
https://github.com/contiamo/operational-ui/blob/f59c010633c3503c77787838fa6271756e4f3ace/src/Breadcrumb/Breadcrumb.tsx#L39
(memo does not work here)
https://github.com/contiamo/operational-ui/blob/c84c10aaa754649587a7c41aade0e155d9e2f06d/src/SimpleLink/SimpleLink.tsx#L31-L32
https://github.com/contiamo/operational-ui/blob/ecc692971b6f207c9d78080addb98d4b4ee7f522/src/SidenavHeader/SidenavHeader.tsx#L136

JoshRosenstein:
I Also wanted to write some sort of test for this. Below is a test for the test 😆 . Comparing the fixed button solution to the current button.

Below currently ✅ , so i will extract out this testing method and put it under some __testUtils__ folder that way it can be applied to this future PR, but also for any of the other components that have this issue.
```ts
import * as React from "react"
import { mount } from "enzyme"
import { Button as ThemelessButton } from "../../index"
import ThemelessButton2  from "../ButtonOld"
import wrapDefaultTheme from "../../utils/wrap-default-theme"

// The 'fixed' Button 
const Button = wrapDefaultTheme(ThemelessButton)
// The Old Button --- This is only here to make sure this test captures the underlying issue
const Button2 = wrapDefaultTheme(ThemelessButton2)

[...]

PR #985, opened 2019-04-19
**Fix:** Components Remounting on every render

JoshRosenstein:
# Summary
Originally this PR was going to be just for Button, but I went ahead and added the same fix to  SimpleLink, SidenavItem,SidenavHeader.
Note: Breadcrumb did not need the fix since `ContainerLink` and `Container` are already created. However we may want to apply the fix ( which will also eliminate the need to cast ContainerComponent as any ) or wrap in memo if any prop of Breadcrumb are expected to change. For now I left it as is, but added tests.
https://github.com/contiamo/operational-ui/blob/402249ae501e468c5b3ad8ccd5fc581a70836b22/src/Breadcrumb/Breadcrumb.tsx#L39

See #984 for additional details.

Side Nav example before and after:
OperationalUI_Render_ISSUE
OperationalUI_Render_Performance
# Related issue
#984


src/__testUtils__/itShouldOnlyMountOnce.tsx
```diff
@@ -0,0 +1,56 @@
+import { mount } from "enzyme"
+import * as React from "react"
+
+class Child extends React.Component<{ mockFN: jest.Mock }> {
+  public componentWillMount() {
```
TejasQ: This hook is slated for deprecation in React. Consider implementing the same component but as a function component with hooks.

src/__testUtils__/itShouldOnlyMountOnce.tsx
```diff
@@ -0,0 +1,56 @@
...
+    this.props.mockFN()
+  }
+
+  public render() {
+    return <div />
+  }
+}
+
+// tslint:disable-next-line: max-classes-per-file
+class TestWrapper extends React.Component<{ comp: React.Factory<{ count: number }> }, { count: number }> {
```
TejasQ: Consider implementing the same component but as a function component with hooks.

src/Button/__tests__/Button.test.tsx
```diff
@@ -9,4 +10,6 @@ describe("Button Component", () => {
   it("Should initialize properly", () => {
     expect(render(<Button>hi</Button>)).toMatchSnapshot()
```
TejasQ: While we're here, maybe we can remove the snapshot test. It was quite useless.

src/__testUtils__/itShouldOnlyMountOnce.tsx
```diff
@@ -1,37 +1,48 @@
...
-// tslint:disable-next-line: max-classes-per-file
-class TestWrapper extends React.Component<{ comp: React.Factory<{ count: number }> }, { count: number }> {
-  constructor(props: any) {
-    super(props)
-    this.state = { count: 1 }
+/**
+ * Must be a class in order to update state with enzyme.
+ * We use this re-render its child on every state change
+ */
+class TestWrapper extends React.Component<{ comp: React.FunctionComponent<{ count: number }> }, { count: number }> {
```
TejasQ: Let's maybe use a function component here instead of a class.

src/Breadcrumb/__tests__/Breadcrumb.test.tsx
```diff
@@ -0,0 +1,16 @@
...
+import { render } from "enzyme"
+import { itShouldOnlyMountOnce } from "../../__testUtils__/itShouldOnlyMountOnce"
+import { Breadcrumb as ThemelessBreadcrumb } from "../../index"
+import wrapDefaultTheme from "../../utils/wrap-default-theme"
+
+const Breadcrumb = wrapDefaultTheme(ThemelessBreadcrumb)
+
+describe("SidenavItem Component", () => {
+  it("Should initialize properly", () => {
+    expect(render(<Breadcrumb to="/some-url">Home</Breadcrumb>)).toMatchSnapshot()
```
TejasQ: Let's not add more snapshot tests willy nilly please.

src/SidenavHeader/__tests__/SidenavHeader.test.tsx
```diff
@@ -0,0 +1,19 @@
...
+import { render, shallow } from "enzyme"
+import { SidenavHeader as ThemelessSideNavHeader } from "../../"
+
+import wrapDefaultTheme from "../../utils/wrap-default-theme"
+
+const SidenavHeader = wrapDefaultTheme(ThemelessSideNavHeader)
+
+describe("SidenavHeader Component", () => {
+  it("Should initialize properly", () => {
+    expect(render(<SidenavHeader label="Chapter One" to="/one" />)).toMatchSnapshot()
```
TejasQ: 😢

src/SidenavItem/__tests__/SidenavItem.test.tsx
```diff
@@ -0,0 +1,13 @@
...
+
+import { render } from "enzyme"
+import { SidenavItem as ThemelessSidenavItem } from "../../index"
+import wrapDefaultTheme from "../../utils/wrap-default-theme"
+
+const SidenavItem = wrapDefaultTheme(ThemelessSidenavItem)
+
+describe("SidenavItem Component", () => {
+  it("Should initialize properly", () => {
+    expect(render(<SidenavItem label="My Account" icon="User" />)).toMatchSnapshot()
```
TejasQ: Never.

src/SimpleLink/__tests__/SimpleLink.test.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+describe("SimpleLink Component", () => {
+  it("Should initialize properly", () => {
+    expect(
+      render(
+        <SimpleLink left icon="Open" color="#000000">
+          Custom color
+        </SimpleLink>,
+      ),
+    ).toMatchSnapshot()
```
TejasQ: Snapshots are evil unless used carefully. This test doesn't add much value to us.

src/Button/Button.tsx
```diff
@@ -69,6 +69,53 @@ const ButtonSpinner = styled(Spinner)<{ containerColor?: ButtonProps["color"] }>
   color: makeColors(theme, containerColor || "").foreground,
 }))
 
+const BaseButton = styled<"button" | "a">("button")<{
```
fabien0102: I don't get it, it was a `a` or `button` before, regarding the `to` props, now it's only a `button` but we have still `button | a` as type 🤔

src/Button/Button.tsx
```diff
@@ -83,57 +130,12 @@ const Button: React.SFC<ButtonProps> = ({
...
-    }
-  })
-
   return (
     <OperationalContext>
       {ctx => (
-        <Container
+        <BaseButton
           {...props}
+          as={to ? "a" : undefined}
```
fabien0102: Don't really follow the logic here also, we have `to` and `as` as props, but we override to `undefined` 🤔

src/SidenavHeader/SidenavHeader.tsx
```diff
@@ -31,14 +31,15 @@ export interface SidenavHeaderProps extends DefaultProps {
   compact?: SidenavProps["compact"]
 }
 
-const makeContainer = (type: "link" | "block") =>
-  styled(type === "link" ? "a" : "div")<{ compact: SidenavHeaderProps["compact"] }>(({ theme, compact }) => ({
+const SidenavHeaderBase = styled<"div" | "a">("div")<{ as?: "div" | "a"; compact: SidenavHeaderProps["compact"] }>(
```
fabien0102: Same, I don't see how this pattern can result to a `a` in the DOM

src/__testUtils__/itShouldOnlyMountOnce.tsx
```diff
@@ -0,0 +1,67 @@
...
+  const mockFn = jest.fn()
+
+  const Inner = ({ count }: { count: number }) => {
+    return (
+      <React.Fragment>
+        <button>Count:{count}</button>
+        <Component>
+          <Child mockFN={mockFn} />
+        </Component>
+      </React.Fragment>
```
fabien0102: This file seams a bit overengineering, a simple jest fn can do the trick:

```jsx
const children = jest.fn();
children.mockReturnValue(<div />);

render(<MyComp>{children}</MyComp>);

expect(children).toBeCalledTimes(1);
```

And it's way more readable 😉

src/SidenavHeader/__tests__/__snapshots__/SidenavHeader.test.tsx.snap
```diff
@@ -0,0 +1,29 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`SidenavHeader Component Should initialize properly 1`] = `
+<div
+  class="css-ftf61z"
+>
+  <div
+    class="css-p29iif"
+  />
+  <a
```
fabien0102: Apparently I'm wrong, we have a `a` tag over there, still don't understand but nobody is perfect 😅 

We should maybe add a test for the `div` case, so at least we have the result. Also, we can specify a bit more the test, a simple snapshot is not really usefull without context.

So for example I would prefer `it("should have a 'a' tag if we provide a 'to' props")` and `it("should have a 'div' tag without 'to' props")`

src/__testUtils__/itShouldOnlyMountOnce.tsx
JoshRosenstein: Need to use a class to work with enzyme, as mentioned in the jsdoc

src/SimpleLink/__tests__/SimpleLink.test.tsx
JoshRosenstein: hmm.. adding a basic snapshot test vs having no tests and solely depending on no errors being thrown in styleguide? I think this would be a step in the right direction...
I actually forget to add jest-emotion in snapshotSerializers,
but for dumb independent representational components that have dynamic styles based on props, having a simple initialized snapshot confirms your css styles and you know what your going to get with the default props.

Again, this is just my thought process. If you'd rather have no tests then i'll revert them all.

src/Button/Button.tsx
JoshRosenstein: Yes because `as` prop is sets to `a` if to is passed. and if we dont tell styled that this can be a button or an anchor, we would get a type error for passing `href` to a button.

src/Button/Button.tsx
JoshRosenstein: BaseButton is a button, if `to` is passed tell styled to change the tag to an anchor, if not, it remains a button. I could replace undefined to 'button' for clarity, but also your telling styled to change the button tag to a button tag....

src/SidenavHeader/SidenavHeader.tsx
JoshRosenstein: https://github.com/emotion-js/emotion/blob/51e48563e22be9a2cae9bcf2cc9ac2f1deca78a0/packages/styled-base/src/index.js#L78

src/__testUtils__/itShouldOnlyMountOnce.tsx
JoshRosenstein: Yeah, I dont see how that is a equivalent test. Your rendering MyComp once, so children are going to mount and rendered once. Purpose of the test is to show the component does not unmount every re-render if none of its props change. so 
```js
const DisplayCurrentCount=({count})=
(<>
<div>{count}<div>
<MyComp > This shouldnt unmount when count changes </MyComp >
</>)

```
Thats why in this file we have a Stateful wrapper in order to update the count with enzyme and track if  MyComp ever re-mounts. Which is the underlying issue of this bug. 

Please see #984 for details of what this test is mimicking.

src/SimpleLink/__tests__/SimpleLink.test.tsx
JoshRosenstein: also, would you like me to delete
https://github.com/contiamo/operational-ui/blob/master/src/Sidenav/__tests__/Sidenav.test.tsx

JoshRosenstein:
bump @fabien0102  @TejasQ

TejasQ:
Dang bro, this has been 2 weeks? Major major

# Major apologies!

Will prioritize this today.

src/__testUtils__/itShouldOnlyMountOnce.tsx
TejasQ: Perhaps we can remove enzyme in favor [`react-testing-library`](https://github.com/testing-library/react-testing-library)? At this point, writing class components seems like an anti-pattern and regressive code style since we're moving things over to Hooks more cumulatively.

TejasQ:
No response from @JoshRosenstein => close.

JoshRosenstein:
@TejasQ - You responded, to one change request (after 2 weeks), and your asking for changes beyond the scope of the this PR. (changing testing library will require refactors to every test that uses enzyme ). Plus the objective of this PR fixes a big bug that effects performance of these components and any application using them, so you close this PR because the **extra test** I wrote doesn't use hooks?

TejasQ:
Hey @JoshRosenstein, I appreciate your response. 

First off, I’d like to be very clear: for me, pull requests are always more about **communication** moreso than code. When communication significantly slows down or stops, I think it best to close Pull Requests as stale rather than leaving them hanging around for a while. 

I think closing a PR is fine because:
- it can be opened at a later stage if/when the communication picks up
- the changes are not lost
- it allows the maintainer, or team if maintainers to focus on other PRs

We try to have as few PRs open as possible for over a week and it has proven quite valuable to our project. 

Secondly, I must apologize: I could have and should have pinged you a few more times before closing it. I could have also been a little more communicative. I apologize. 

Now — about the code. 

- I didn’t really ask to change testing libraries, but hoped to open up a discussion wherein we could dismiss it together as out of the scope of this PR. 

- I really appreciate the fixes this PR makes. My intention was never really to throw it or its changes away, but to close them while the team is busy with other tasks.  Also because communication with you seemed to have slowed down. We would, of course, come back to it when either we have more time, or you were able to respond. 

- As stated, I closed it not because of hooks, but because of communication and workload on our internal team. 

Shall we reopen it and work towards getting it merged? 😄

JoshRosenstein:
10-4, and i apologize, didn't see `react-testing-library` was already a dependency. Although `react-testing-library` doesn't have built-in functions to directly update the state. I was able to come up a way to mock it. (without classes 😉 )

Heres the sandbox with the `itShouldOnlyMountOnce` using `react-testing-library`, If you open it up, i will push the changes.

[Edit Operational-ui Button Issue](https://codesandbox.io/s/8lqrj509oj?fontsize=14&moduleview=1)

TejasQ:
Dude @JoshRosenstein, that's incredible. Thanks so much for your incredible contributions and communication. 😍  Let's get this bad boy merged. 🚀

TejasQ:
@JoshRosenstein I love you 😍 

Great work! Thanks so much!

PR #987, opened 2019-04-23
Some updates to fix examples

JoshRosenstein:
PR to merge into #982 

Realized I didnt push out my commits yesterday, and when i tried- I had all my commits were on upstream branch. So i had to make a new branch and push these 2 squashed commits. 

For Icons, i went ahead a made a enum for all icons and changed up the typing a bit. Im export `allIcons` so it can be used in the example, but just realized i can just use the `ICONS` enum to get the example to work, so i can removed that export in `index.ts`.


src/Icon/types.ts
```diff
@@ -0,0 +1,127 @@
+import { DefaultProps } from "../types"
+
+export enum ICONS {
```
fabien0102: I'm really not a fan of this enum, as a lazy developer, I prefer union instead of enum to write `Add` instead of `ICONS.Add`. It will also probably break the current API 😕.

After I need to try, if we can still write `<Icon name="Add" />` with the autocompletion as before, it's good to me 😉

src/Icon/types.ts
```diff
@@ -0,0 +1,127 @@
...
+  Undo = "Undo",
+  Users = "Users",
+  Yes = "Yes",
+}
+
+export type BRAND_ICON = ICONS.Contiamo | ICONS.OperationalUI | ICONS.Pantheon | ICONS.Labs
+export type CUSTOM_ICON = Exclude<ICONS, BRAND_ICON>
+
+// Remove keyof typeof if we require enums to be used
+export type IconName = keyof typeof ICONS
```
JoshRosenstein: @fabien0102 This here is whats preventing the api from changing and still gives auto-completion .

fabien0102: Should we update the examples so? to use the string instead of the enum? 😃

src/Icon/Icon.Brand.tsx
```diff
@@ -31,7 +24,7 @@ const OperationalUI: React.SFC<OperationalUIIconProps> = ({ rotation, size, colo
   )
 }
 
-const Contiamo: React.SFC<Props> = ({ size, color, ...props }) => {
+const Contiamo: BrandIconsData[ICONS.Contiamo] = ({ size, color, ...props }) => {
```
TejasQ: This is confusing. We're referencing:

- property `Contiamo` on 
- an `enum ICONS` which is 
- a computed property of `interface BrandIconsData`.

I think we can and should simplify this, perhaps with a simpler union type?

JoshRosenstein: Just did it that way to be safer, for instance, if you were change the name of an icon, you will get a typeerror reminding you to update the enum.

also in order to replaces the original generalized interface 
```js
export interface BrandIcons {
  [key: string]: React.SFC<Props> | React.SFC<OperationalUIIconProps> | React.SFC<PantheonIconProps>
}
```
I created the BrandIconsData, but i can delete this interface and just have
```js
const Contiamo: React.SFC<SVGProps> ...
const  OperationalUI: React.SFC<OmitName<OperationalUIIconProps>> ...

const BrandIcons = {
  OperationalUI,
  Contiamo,
  Labs,
  Pantheon,
}
export default BrandIcons
```

but if you add another icon
```js
const BrandIcons = {
  OperationalUI,
  Contiamo,
  Labs,
  Pantheon,
  somethingElse
}
```
You wont get error reminding you to update ICONS and BRAND_ICON.

Unless you want we to completely scrap this im fine with that, the only thing we would have to do is export the icon names in order to get the example to work.

fabien0102:
@TejasQ Is this PR still relevant? I know you are working with styleguist team to add typescript in their stack, so… Bref, I let you the lead of this PR 😉

TejasQ:
Closing this because it's stale.

Issue #983, opened 2019-04-18
Spread props onto `TextArea`


TejasQ:
There's no way to spread an `autoFocus` prop on a `Textarea`. Let's fix this.

mpotomin:
Implemented with [988](https://github.com/contiamo/operational-ui/pull/988)

PR #988, opened 2019-04-23
**Feature:** Allow autoFocus for the TextArea

mpotomin:
# Summary
Passes autoFocus prop to the TextArea
https://github.com/contiamo/operational-ui/issues/983

# To be tested

Me

Tester 1

### DIFF
--- a/src/Textarea/Textarea.tsx
+++ b/src/Textarea/Textarea.tsx
@@ -28,6 +28,8 @@ export interface TextareaProps extends DefaultProps, DefaultInputProps {
   height?: number
   /** Text for a hint */
   hint?: string
+  /** Automatically focus the input on page load */
+  autoFocus?: boolean
   /** Error text */
   error?: string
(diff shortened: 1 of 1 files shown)


mpotomin:
@TejasQ Could you please be more specific regarding which exactly props should be passed to the TextArea? Shall I pass all from TextArea's [MDN page](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea)?

TejasQ:
@mpotomin – according to the issue, I mean let's just spread all incoming props onto the underlying textarea, [like we do on `Input`](https://github.com/contiamo/operational-ui/blob/402249ae501e468c5b3ad8ccd5fc581a70836b22/src/Input/Input.Field.tsx#L153).

https://github.com/contiamo/operational-ui/blob/402249ae501e468c5b3ad8ccd5fc581a70836b22/src/Input/Input.Field.tsx#L153

PR #991, opened 2019-04-24
**Fix:** Chip a11y

### LINKED ISSUE #990: `Chip`s update on hover even if unclickable
Currently, the background color of `Chip`s becomes darker when hovered – even if no click handler is present. I suggest we don't react to a cursor if a Chip is meant to be unresponsive.

TejasQ:
Fixes #990 and adds a11y superpowers to `Chip`.

lalala

### DIFF
--- a/src/Chip/Chip.tsx
+++ b/src/Chip/Chip.tsx
@@ -43,7 +43,7 @@ const Container = styled("div")<{ color_?: string }>(({ theme, color_ }) => {
   }
 })
 
-const getHoverStyle = (
+const getInteractiveStyle = (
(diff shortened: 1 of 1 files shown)


src/Chip/Chip.tsx
```diff
@@ -17,58 +18,90 @@ export interface ChipProps extends DefaultProps {
   children: React.ReactNode
 }
 
+const height = 20
```
fabien0102: Why? The theme value was not good enough for you? 🤔

TejasQ: theme.space for height is an anti-pattern

Issue #993, opened 2019-05-03
Page: tabs - controlled activeTabName is broken


mpotomin:
Page section `With activeTabName controlled (classically with a router)` is broken
http://operational-ui.netlify.com/#/Components/Page

TejasQ:
so fast

PR #994, opened 2019-05-03
  **Fix:** Page controlled activeTabName

mpotomin:
# Summary
Fixes the bug with non-working controlled activeTabName for Tabs at the page. The value of the activeTabName renders correctly at the first component renders and stays stale, ignoring new value being passed through props later

# Related issue
[bug 993](https://github.com/contiamo/operational-ui/issues/993)

# To be tested

Me

Tester 1
- [ ] Tabs work as expected

Tester 2
- [ ] Tabs work as expected


TejasQ:
MISCHA HOW ARE YOU SO GOOD WHAT THE HECK

PR #998, opened 2019-05-07
**Feature:** DataTable

TejasQ:
This PR contains an initial implementation of this component:

image

- It is **fully controlled** and is completely agnostic of the data that its cells can contain. 
- It is **generic** and enforces rigidity in its row structure:
  - all rows have to be of a consistent length (no case where a cell is missing)
  - all cells have to be of a consistent type

## Proposed API

```ts
interface DataTableProps<T> {
  editMode: boolean
  onAddClick: () => void
  rows: Array<{
    isHeading?: boolean
    isDisabled?: boolean
    cells: T
  }>
}
```


micha-f:
Would be nice to implement sticky headers so that you can always see them when scrolling.

micha-f:
@TejasQ Not every table should support deleting columns. Do we really want to add this functionality here? Can't we just say that whenever someone needs column deletion, the developer has to add and wire up the icon in each cell?

TejasQ:
I'm not sure because the point of this library is to "plug and play", right? I think it would be more pleasant for @mpotomin if he feeds it data and it provides him pixel-perfect controls.

I think this component should be stateless and fully controlled by him, but should fully handle the concerns of UI visuals so that we don't duplicate icons/similar buttons in other parts of the product.

micha-f:
@TejasQ Ok! Adding / removing columns is an optional feature though - right?

TejasQ:
Yes! You only get a little delete icon if you specify `onRemoveColumn`.

micha-f:
@TejasQ makes sense

src/DataTable/DataTable.tsx
```diff
@@ -0,0 +1,176 @@
...
+`
+
+const HeaderRow = styled("div")<{ rowHeight: number; numHeaders: number }>`
+  left: 0;
+  width: 100%;
+  height: ${({ rowHeight, numHeaders }) => rowHeight * numHeaders}px;
+  position: sticky;
+  grid-row: 1;
+  top: 0;
+  z-index: 100;
```
mpotomin: Will it be isolated by emotion within this component for us or will it be global?

src/DataTable/DataTable.tsx
```diff
@@ -0,0 +1,176 @@
...
+  display: grid;
+  grid-template-columns: repeat(${({ numCells, cellWidth }) => `${numCells}, ${cellWidth}`});
+  background-color: ${({ theme, isEven }) => (isEven ? theme.color.background.almostWhite : theme.color.white)};
+`
+
+const Cell = styled("div", { shouldForwardProp: prop => prop !== "height" })<{ height: number; cell: number }>`
+  position: relative;
+  display: flex;
+  align-items: center;
+  box-shadow: 0 0 0 0.5px ${({ theme }) => theme.color.border.medium};
```
mpotomin: 0.5 might get blurry

src/DataTable/DataTable.tsx
```diff
@@ -0,0 +1,185 @@
...
+const Container = styled("div", { shouldForwardProp: prop => prop !== "width" })<{ width: string }>`
+  position: relative;
+  width: ${({ width }) => width};
+  border: 1px solid ${({ theme }) => theme.color.border.medium};
+`
+
+const TableContainer = styled("div")<{
+  numColumns: number
+  numHeaders: number
+  columnWidth: string
```
mpotomin: string only? Can it be `string | number` here?

TejasQ: I would rather leave it string, in case some people want to use `%`, or `fr`, or `em`.

Would you prefer assuming a `px` suffix in case of `number`?

src/DataTable/DataTable.tsx
TejasQ: This won't be isolated by emotion, but by the parent context => the `TableContainer`. 😄

src/DataTable/DataTable.tsx
mpotomin: Let's have is as string and let the consumer decide 👌

fabien0102:
image

It's a bit strange to have the scrolling bar covering the entire table since we have sticky header. I would expect the same UX as github:

image

fabien0102:
image

@TejasQ You are so lazy 😅 Can you use the operational-ui `Button` instead of the default one? 😇

src/DataTable/DataTable.tsx
```diff
@@ -0,0 +1,193 @@
+import * as React from "react"
+import { FixedSizeList, ListChildComponentProps } from "react-window"
+
+import Message from "../Internals/Message/Message"
+import styled from "../utils/styled"
+
+export interface DataTableProps<T, P> {
```
fabien0102: I think we should use `TRow` and `TColumn` here to have a better interface

TejasQ:
> Did you add any security about the maximal array lenght?

Yes. [I increment by 1 instead of exponentially when it gets too big.](https://github.com/contiamo/operational-ui/pull/998/files#diff-a4aa2f64a168ae40520af51414b2664eR6). 😅

kemal-contiamo:
> image
> 
> It's a bit strange to have the scrolling bar covering the entire table since we have sticky header. I would expect the same UX as github:
> 
> image

I'll go with Fabien's recommendation on the scrollbar. I think makes sense to have the scrollbar only covers the content area (and not the whole height including header).

TejasQ:
@kemal-contiamo what happens in case of horizontal scrolling then? Do the column headers scroll independently of the rows? It seems like this needs a bit more thought.

cypress/integration/DataTable.ts
```diff
@@ -0,0 +1,26 @@
+describe("DataTable", () => {
+  before(() => {
+    cy.visit("/#!/DataTable/1")
+  })
+  it("should exist", () => {
+    cy.get(".operational-ui__DataTable--virtual-scroller")
```
stereobooster: `cy.get("[role=table]")`?

src/DataTable/DataTable.tsx
```diff
@@ -0,0 +1,192 @@
...
+  rows: P[][]
+
+  /** How high is each row (in pixels)? */
+  rowHeight?: "regular" | "compact" | number
+
+  /** Shall we include a footer? */
+  footer?: React.ReactNode
+
+  /** How much shall we restrict the height? */
+  height?: number
```
stereobooster:
```suggestion
  height?: string
```

the same as width?

src/DataTable/DataTable.tsx
```diff
@@ -0,0 +1,192 @@
...
+  position: relative;
+  width: ${({ width }) => width};
+  border: 1px solid ${({ theme }) => theme.color.border.medium};
+`
+
+const HeadersContainer = styled("div")<{
+  numColumns: number
+  numHeaders: number
+  columnWidth: string
+  rowHeight: number
```
stereobooster:
```suggestion
  rowHeight: string
```

src/DataTable/DataTable.tsx
```diff
@@ -0,0 +1,192 @@
...
+  )
+
+  const numCells = React.useMemo(() => rows[0].length, [rows])
+  const VirtualRow: React.FC<ListChildComponentProps> = React.useMemo(
+    () =>
+      React.memo(({ style, index }) => (
+        <Row isEven={index % 2 === 0} style={{ ...style, height: rowHeight }} cellWidth={cellWidth} numCells={numCells}>
+          {rows[index] &&
+            rows[index].map((cell, cellIndex) => (
+              <Cell key={`op-row-${cellIndex}-${performance.now()}`} cell={cellIndex + 1} height={rowHeight}>
```
stereobooster: `performance.now()` why?

cypress/integration/DataTable.ts
stereobooster: `can't use data-cy or any other prop because of react-window `

src/DataTable/DataTable.tsx
TejasQ: `react-window`'s API requires a number. Numbers are also a lot more comfortable to use elsewhere. I had it as `string` initially but I did `parseInt(height, 10)` too many times so I made it a number.

src/DataTable/DataTable.tsx
TejasQ: https://github.com/contiamo/operational-ui/pull/998#discussion_r286358057

src/DataTable/DataTable.tsx
TejasQ: Ensure keys are 100% unique at render time in the case of multiple tables on the same page. Is there a better way to do this? 😅

src/utils/truncate.ts
```diff
@@ -0,0 +1,17 @@
+export const truncate = (length: number) => (thingToTruncate: any) => {
+  if (["string", "number"].includes(typeof thingToTruncate)) {
+    const stringifiedValue = String(thingToTruncate)
+    const truncatedValue = String(thingToTruncate).slice(0, length)
+    if (stringifiedValue.length === truncatedValue.length) {
+      return thingToTruncate
+    } else {
+      return truncatedValue + "..."
```
stereobooster:
```suggestion
      return truncatedValue + "…"
```

it is `&hellip;`

src/utils/truncate.ts
```diff
@@ -0,0 +1,17 @@
+export const truncate = (length: number) => (thingToTruncate: any) => {
+  if (["string", "number"].includes(typeof thingToTruncate)) {
```
stereobooster: maybe we can do this check in dev environment, but not in prod?

stereobooster:
Screenshot 2019-05-22 at 10 07 25

src/utils/truncate.ts
TejasQ:
```suggestion
  if (["string", "number"].includes(typeof thingToTruncate) && process.env.NODE_ENV !== "production") {
```

TejasQ: Like this?

src/DataTable/DataTable.tsx
stereobooster: yes I saw it later in the code. The confusing part is `<DataTable height={10} width={"10"} />`

src/DataTable/DataTable.tsx
```diff
@@ -0,0 +1,192 @@
...
+  /** How high is each row (in pixels)? */
+  rowHeight?: "regular" | "compact" | number
+
+  /** Shall we include a footer? */
+  footer?: React.ReactNode
+
+  /** How much shall we restrict the height? */
+  height?: number
+
+  /** How much shall we restrict the width? Default: unlimited */
```
stereobooster:
```suggestion
  /** How much shall we restrict the width? Default: `unlimited` */
```

src/DataTable/DataTable.tsx
stereobooster:
```suggestion
              <Cell key={`op-row-${cellIndex}-${index}`} cell={cellIndex + 1} height={rowHeight}>
```

src/utils/__tests__/truncate.test.ts
```diff
@@ -0,0 +1,22 @@
+import { truncate } from "../truncate"
+
+describe("Truncate", () => {
+  it("should return the same thing if no truncation required", () => {
+    expect(truncate(5)("Tejas")).toEqual("Tejas")
+  })
+
+  it("should return a truncated string if it can", () => {
+    expect(truncate(5)("Je m'appelle Claude")).toEqual("Je m'...")
```
TejasQ:
```suggestion
    expect(truncate(5)("Je m'appelle Claude")).toEqual("Je m'…")
```

src/DataTable/DataTable.tsx
TejasQ: I agree, but it's quite common on the web. Kind of like the [IntersectionObserver API](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API#Creating_an_intersection_observer) (`threshold` is `number` while `rootMargin` is `string`)

src/utils/truncate.ts
stereobooster:
```ts
/** Use `text-overflow: ellipsis` if you can */ 
export const truncate = (length: number) => (thingToTruncate: any) => {
  if (process.env.NODE_ENV !== "production")
    if (!["string", "number"].includes(typeof thingToTruncate)) {
      console.warn(
        "Attempting to truncate something that is not a JS primitive: this will fail for functions, objects and arrays. Please reconsider the item you are truncating:",
        thingToTruncate,
      )
    }
  }

  const stringifiedValue = String(thingToTruncate)
  if (stringifiedValue.length <= length) {
    return stringifiedValue
  } else {
    return stringifiedValue.slice(0, length) + "…"
  }
}
```

?

TejasQ:
@kemal-contiamo @micha-f are the scrollbars referenced in [this comment](https://github.com/contiamo/operational-ui/pull/998#issuecomment-494404123) super important or a showstopper? I ask because: it's quite complicated to add them that way, separating the headers from the content because the headers will be in a separate element, with separate overflow properties.

This means that in the case of a horizontal overflow (second example), there will be two horizontal scrollbars: one at the bottom of the "data", and another at the bottom of the "headers", which will scroll independently. The solution then is to spy on one scrolling element and programatically scroll the other one. I'm not sure this is the best approach, or needed right now.

I'm not sure if we want to spend more time there, blocking this component from being considered done? On Mac devices, scrollbars aren't even visible by default. Would appreciate your thoughts.

micha-f:
@TejasQ let's come back to that at a later time then 👍

src/utils/truncate.ts
TejasQ: We decided to remove the warning.

src/utils/truncate.ts
```diff
@@ -0,0 +1,23 @@
+let didWarn = false
+
+export const truncate = (length: number) => (thingToTruncate: any) => {
```
stereobooster: How about truncation in the middle, like `long long text…but I can read ned as well`?

TejasQ: Let's add this later.

TejasQ:
@kemal-contiamo @micha-f your thoughts?

https://deploy-preview-998--operational-ui.netlify.com/#!/DataTable

micha-f:
@TejasQ Looks great, merge it 👍

PR #1003, opened 2019-05-29
  **Feature:** Tabs component

stereobooster:
# Summary

Screenshot 2019-05-31 at 16 39 05

Drag-n-drop will be in separate PR

# To be tested

Me

Tester 1

Tester 2

### DIFF
--- a/cypress/integration/Accordion.ts
+++ b/cypress/integration/Accordion.ts
@@ -1,11 +1,6 @@
-describe("Accordion", () => {
-  const skipStyleguidistLinks = () =>
-    cy
-      .get("body")
-      .tab() // Show all components
-      .tab() // Props & methods
-      .tab() // link in the readme
+import { tabPastStyleguidistLinks } from "../support/helpers"
 
+describe("Accordion", () => {
   before(() => {
     cy.visit("/#!/Accordion")
   })
@@ -50,7 +45,7 @@ describe("Accordion", () => {
       .first()
       .should("not.have.attr", "hidden")
 
-    skipStyleguidistLinks()
+    tabPastStyleguidistLinks()
       .tab() // First header in Accordion
       .type(" ")
 
@@ -66,7 +61,7 @@ describe("Accordion", () => {
       .first()
       .should("have.attr", "hidden")
 
-    skipStyleguidistLinks()
+    tabPastStyleguidistLinks()
       .tab()
       .type(" ")
 
--- a/cypress/integration/Accordion.ts
+++ b/cypress/integration/Accordion.ts
@@ -1,10 +1,13 @@
 describe("Accordion", () => {
+  const skipStyleguidistLinks = () =>
+    cy
+      .get("body")
+      .tab() // Show all components
+      .tab() // Props & methods
+      .tab() // link in the readme
+
   before(() => {
     cy.visit("/#!/Accordion")
-  })
-
-  it("should exist", () => {
-    cy.get("[data-cy=operational-ui__Accordion]")
   })
 
   it("opens section if you click corresponding header", () => {
@@ -47,10 +50,7 @@ describe("Accordion", () => {
       .first()
       .should("not.have.attr", "hidden")
 
-    cy.get("body")
-      .tab() // Show all components
-      .tab() // Props & methods
-      .tab() // link
+    skipStyleguidistLinks()
       .tab() // First header in Accordion
       .type(" ")
 
@@ -66,10 +66,7 @@ describe("Accordion", () => {
       .first()
       .should("have.attr", "hidden")
 
-    cy.get("body")
-      .tab()
-      .tab()
-      .tab()
+    skipStyleguidistLinks()
       .tab()
       .type(" ")
 
--- a/cypress/integration/Tabs.ts
+++ b/cypress/integration/Tabs.ts
@@ -1,0 +1,114 @@
+describe("Tabs", () => {
+  const skipStyleguidistLinks = () =>
+    cy
+      .get("body")
+      .tab() // Show all components
+      .tab() // Props & methods
+      .tab() // link in the readme
+
+  before(() => {
+    cy.visit("/#!/Tabs")
+  })
+
+  it("opens section if you click corresponding header", () => {
+    cy.get("[data-cy=operational-ui__Tabs]")
+      .find("[role=tabpanel]")
+      .last()
+      .should("have.attr", "hidden")
+
+    cy.get("[data-cy=operational-ui__Tabs]")
+      .find("[role=tab]")
+      .last()
+      .click()
+
+    cy.get("[data-cy=operational-ui__Tabs]")
+      .find("[role=tabpanel]")
+      .last()
+      .should("not.have.attr", "hidden")
+  })
+
+  it("closes section if you click `x` icon in corresponding header", () => {
+    cy.get("[data-cy=operational-ui__Tabs]")
+      .find("[role=tabpanel]")
+      .should("have.length", 4)
+
+    cy.get("[data-cy=operational-ui__Tabs]")
+      .find("[role=tab]")
+      .find("svg")
+      .last()
+      .click()
+
+    cy.get("[data-cy=operational-ui__Tabs]")
+      .find("[role=tabpanel]")
+      .should("have.length", 3)
+    cy.focused().should("have.length", 0)
+  })
+
+  it("tab gets on currently selected tab", () => {
+    skipStyleguidistLinks().tab() // Selected header in Tabs
+
+    cy.focused().contains("tab 3")
+  })
+
+  it("home button moves focus to the first tab", () => {
+    cy.focused().contains("tab 3")
+
+    cy.focused().type("{home}")
+
+    cy.focused().contains("tab 1")
+    cy.get("[data-cy=operational-ui__Tabs]")
+      .find("[role=tabpanel]")
+      .first()
+      .should("not.have.attr", "hidden")
+  })
+
+  it("end button moves focus to the last tab", () => {
+    cy.focused().contains("tab 1")
+
+    cy.focused().type("{end}")
+
+    cy.focused().contains("tab 3")
+    cy.get("[data-cy=operational-ui__Tabs]")
+      .find("[role=tabpanel]")
+      .last()
+      .should("not.have.attr", "hidden")
+  })
+
+  it("right arrow moves focus to the next tab", () => {
+    cy.focused().contains("tab 3")
+
+    cy.focused().type("{rightarrow}")
+
+    cy.focused().contains("tab 1")
+    cy.get("[data-cy=operational-ui__Tabs]")
+      .find("[role=tabpanel]")
+      .first()
+      .should("not.have.attr", "hidden")
+  })
+
+  it("left arrow moves focus to the previous tab", () => {
+    cy.focused().contains("tab 1")
+
+    cy.focused().type("{leftarrow}")
+
+    cy.focused().contains("tab 3")
+    cy.get("[data-cy=operational-ui__Tabs]")
+      .find("[role=tabpanel]")
+      .last()
+      .should("not.have.attr", "hidden")
+  })
+
+  it("delete button coloses tab", () => {
+    cy.get("[data-cy=operational-ui__Tabs]")
+      .find("[role=tabpanel]")
+      .should("have.length", 3)
+    cy.focused().contains("tab 3")
+
+    cy.focused().type("{del}")
+
+    cy.get("[data-cy=operational-ui__Tabs]")
+      .find("[role=tabpanel]")
+      .should("have.length", 2)
+    cy.focused().contains("tab 2")
+  })
+})
--- a/cypress/integration/Tabs.ts
+++ b/cypress/integration/Tabs.ts
@@ -10,16 +10,13 @@ describe("Tabs", () => {
     cy.visit("/#!/Tabs")
   })
 
-  it("opens section if you click corresponding header", () => {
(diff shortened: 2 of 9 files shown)


src/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,243 @@
...
+
+// we need this one to show ellipsis if title is to long
+const TitleWrapper = styled("span")`
+  overflow: hidden;
+  white-space: nowrap;
+  text-overflow: ellipsis;
+`
+
+const LeftIcon = styled(Icon)`
+  color: #0b6285;
```
stereobooster: how to name this one?

micha-f: TabIcon?

stereobooster: I mean color. But nice idea for component name

TejasQ: @stereobooster is this not `darken(primary, 0.5)`?

cypress/integration/Accordion.ts
```diff
@@ -1,12 +1,15 @@
 describe("Accordion", () => {
+  const skipStyleguidistLinks = () =>
```
TejasQ: Maybe export this from one place instead of declaring it twice (here and in `Tabs`).

TejasQ: I also suggest calling it `tabPastStyleguidistLinks`.

cypress/integration/Tabs.ts
```diff
@@ -0,0 +1,137 @@
...
+    cy.focused().type("{leftarrow}")
+
+    cy.focused().contains("tab 3")
+    cy.get("[data-cy=operational-ui__Tabs]")
+      .find("[role=tabpanel]")
+      .last()
+      .should("not.have.attr", "hidden")
+  })
+
+  it("Delete key coloses tab", () => {
```
TejasQ:
```suggestion
  it("Delete key closes tab", () => {
```

src/Tabs/README.md
```diff
@@ -0,0 +1,100 @@
...
+Tabs component implemented according to [WAI-ARIA specification](https://www.w3.org/TR/wai-aria-practices/examples/tabs/tabs-1/tabs.html).
+
+### Simple usage
+
+```jsx
+import * as React from "react"
+import { Tabs } from "@operational/components"
+
+const MyComponent = () => {
+  const [tabs, setTabs] = React.useState([
```
TejasQ: Is there place for a `useTabState` hook? Or is this fine?

src/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,243 @@
...
+  :focus {
+    outline: none;
+    ${({ theme }) => `box-shadow: ${theme.shadows.insetFocus};`}
+  }
+  z-index: 1;
+`
+
+TabHeader.defaultProps = {
+  role: "tab",
+  // @ts-ignore styled components TS definitions doesn't like `as` prop, but it works just fine
```
TejasQ: 👍

src/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,243 @@
...
+  as: "button",
+}
+
+const TabContainer = styled("div")`
+  border: solid 1px ${({ theme }) => theme.color.separators.default};
```
TejasQ: Would using `box-shadow` or `outline` remove the need for the negative margin hack below?

src/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,243 @@
...
+  margin-right: 8px;
+`
+LeftIcon.defaultProps = {
+  size: 14,
+}
+
+const Tabs = ({ tabs, active, onClose, onActivate, onInsert, label, style }: TabsProps) => {
+  if (!Number.isInteger(active) || active < 0 || active >= tabs.length) {
+    active = active > 0 ? tabs.length - 1 : 0
+    console.warn("active tab is out of bound, fall-back to closest value")
```
TejasQ:
```suggestion
    if (process.env.NODE_ENV !== "production") console.warn("Active tab is out of bounds, falling back to closest value.")
```

src/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,243 @@
...
+      activeTab.current.focus()
+      userAction.current = false
+    }
+  }, [active, tabs])
+
+  return (
+    <Container data-cy="operational-ui__Tabs" style={style}>
+      <TabList
+        aria-label={label}
+        onKeyDown={e => {
```
TejasQ: Let's `useCallback` this util above the `return` for readability. 🙌  Maybe we add a unit test for it too.

src/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,243 @@
...
+          }
+        }}
+      >
+        {tabs.map(({ key, title, icon }, i) => (
+          <TabHeader
+            tabIndex={i === active ? 0 : -1}
+            first={i === 0}
+            aria-selected={i === active}
+            aria-controls={`TabPanel${key}`}
+            id={`TabHeader${key}`}
```
TejasQ: We will have duplicate `id`s if multiple `Tabs`s exist on a page. Do we want to `useUniqueId`?

src/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,243 @@
...
+            key={key}
+            onClick={() => {
+              userAction.current = true
+              onActivate(i)
+            }}
+            onFocus={() => {
```
TejasQ: We can maybe hoist, memoize and test this because it's a duplicate callback.

src/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,243 @@
...
+              />
+            )}
+          </TabHeader>
+        ))}
+        {onInsert && (
+          <TabHeader
+            tabIndex={-1}
+            first={false}
+            aria-selected={false}
+            aria-controls={``}
```
TejasQ: ?

src/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,243 @@
...
+            id={`TabAddHeader`}
```
TejasQ: Is having duplicate `id`s OK?

src/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,243 @@
...
+            addButton={true}
+            onMouseDown={e => {
+              e.preventDefault()
+              onInsert(tabs.length - 1)
+            }}
+          >
+            <Icon size={14} name="Add" onMouseDown={() => undefined} />
```
TejasQ:
```suggestion
            <Icon size={14} name="Add" onMouseDown={noop} />
```

src/Tabs/README.md
stereobooster: maybe after d-n-d feature

src/Tabs/Tabs.tsx
stereobooster: this is to make active tab and space below as one object

src/Tabs/Tabs.tsx
stereobooster: 🤔 I don't know need to check

src/Tabs/README.md
```diff
@@ -0,0 +1,100 @@
...
+        <div>
+          <p>Lorem ipsum 4</p>
+        </div>
+      ),
+      key: 3,
+    },
+  ])
+  const [active, setActive] = React.useState(0)
+
+  const onClose = index => {
```
TejasQ: React.useCallback so we don't recompute this on every render.

src/Tabs/Tabs.tsx
stereobooster: darken(primary, 0.5) === #1497ca
darken(primary, 50) === #000000

src/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,254 @@
+import * as React from "react"
+import Icon, { IconName } from "../Icon/Icon"
+import { SectionHeader } from "../Internals/SectionHeader"
+import { DefaultProps } from "../types"
+import styled from "../utils/styled"
+import { useUniqueId } from "../useUniqueId"
+
+const noop = () => undefined
```
TejasQ:
```suggestion
import noop from "lodash/noop"
```

src/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,254 @@
...
+      userAction.current = false
+    }
+  }, [active, tabs])
+
+  const onKeyDown = React.useCallback(
+    (event: React.KeyboardEvent) => {
+      userAction.current = true
+      switch (event.key) {
+        case "ArrowRight":
+          event.preventDefault()
```
TejasQ: This can go outside the `case` block.

stereobooster: nope, we want to prevent only specific keys not all of them

TejasQ: :angry-shaking-fist:

src/Tabs/Tabs.tsx
stereobooster: But I guess importing it will take more bytes than just inline. noop is alway noop, no real value importing it `¯\_(ツ)_/¯`.

src/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,253 @@
...
+  color: #0b6285;
+  margin-right: 8px;
+`
+LeftIcon.defaultProps = {
+  size: 14,
+}
+
+const Tabs = ({ tabs, active, onClose, onActivate, onInsert, label, style, id }: TabsProps) => {
+  if (!Number.isInteger(active) || active < 0 || active >= tabs.length) {
+    active = active > 0 ? tabs.length - 1 : 0
```
TejasQ: Are you mutating props?

src/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,253 @@
...
+      }
+    },
+    [active, onActivate, onInsert],
+  )
+
+  return (
+    <Container data-cy="operational-ui__Tabs" style={style}>
+      <TabList aria-label={label} onKeyDown={onKeyDown}>
+        {tabs.map(({ key, title, icon }, i) => {
+          const onClick = () => {
```
TejasQ: Can we hoist this above `return` and wrap it in `useCallback` since it's a callback?

src/Tabs/Tabs.tsx
stereobooster: it is not props as as is this is local variable which you get after destructuring props. No I'm not mutating props

src/Tabs/Tabs.tsx
stereobooster: It relies on `i` inside this closure

src/Tabs/Tabs.tsx
TejasQ: This local variable is now holding state though, right?

src/Tabs/Tabs.tsx
TejasQ: We can pass `i` as an argument though. 🤔 

In exchange we get memoization and readability in this block.

stereobooster: we don't get memoisation, because every time I will bind new i it will produce new closure

src/Tabs/Tabs.tsx
stereobooster: No I just reassigned variable as if I would use

```
const Tabs = (props: TabsProps) => {
const { tabs, onClose, onActivate, onInsert, label, style, id } = props;
let { active } = props;
active = active > 0 ? tabs.length - 1 : 0
```

src/Tabs/Tabs.tsx
TejasQ: Why do you need to bind a new `i` if you pass it as an argument?

Will this produce a new closure when you call it with different `i`s?

```js
const handleThing = useCallback(
  function(i) {
    alert(i)
  }, [])
```

stereobooster: As of now `TabHeader` is styled `div`. So I get this code:

```tsx
{tabs.map(({ key, title, icon }, i) => {
  return (<div onClick={() => handleThing(i)} />)
})}
```

I need to change TabHeader, so that it would preserve `i` and pass it down if I want to use memoised callback. 

```tsx
const TabHeader = ({ i, onClick, ...rest }) => (<div onClick={onClick(i)} />)
```

then I would be able to do

```tsx
{tabs.map(({ key, title, icon }, i) => {
  return (<TabHeader onClick={handleThing} />)
})}
```

src/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,249 @@
...
+  children?: never
+  label?: string
+  style?: React.CSSProperties
+  id?: string
+}
+
+const Container = styled("div")`
+  label: Tabs;
+  display: grid;
+  grid-template-rows: ${({ theme }) => `${theme.space.element * 2}px 1fr`};
```
stereobooster:
```suggestion
  grid-template-rows: 40px 1fr;
```

src/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,249 @@
...
+
+// we need this one to show ellipsis if title is to long
+const TitleWrapper = styled("span")`
+  overflow: hidden;
+  white-space: nowrap;
+  text-overflow: ellipsis;
+`
+
+const TabIcon = styled("span")`
+  margin-right: 8px;
```
stereobooster:
```suggestion
  margin-right: ${({theme}) => theme.space.small}px;
```

PR #1004, opened 2019-05-30
🧹Housekeeping 🧹

TejasQ:
This PR:
- upgrades TypeScript (and uses `incremental` 🔥)
  - Fixes a _lot_ of type issues
- removes TSLint
- moves type-checking to a separate thread (so faster startup)
- upgrades TS on every start (let's try it; we can remove it if it gets annoying or if TS is too unstable)
- Reformats `Status` because it was untouched by prettier somehow
- Only publishes `@next` when we merge to `master`. (This reduces spam on our slack at **#front-end-team**)


.tsbuildinfo
```diff
@@ -0,0 +1,5582 @@
+{
+  "program": {
+    "fileInfos": {
+      "/users/tejaskumar/sites/operational-ui/node_modules/typescript/lib/lib.es5.d.ts": {
```
stereobooster: add this file to the gitignore

src/Icon/Icon.Brand.tsx
```diff
@@ -128,11 +131,9 @@ const Pantheon: React.SFC<PantheonIconProps> = ({ size, colored, ...props }) =>
...
 
-const BrandIcons: BrandIcons = {
+export const BrandIcons: IBrandIcons = {
   OperationalUI,
   Contiamo,
   Labs,
   Pantheon,
 }
-
-export default BrandIcons
```
stereobooster: Is this a breaking change?

package.json
```diff
@@ -15,7 +15,9 @@
   ],
   "scripts": {
     "dev": "parcel dev-server/index.html",
-    "start": "styleguidist server",
+    "start": "run-s start:*",
+    "start:update-ts": "yarn add typescript@next -D || (echo \"You don't have yarn. Please install yarn to use this package from https://yarnpkg.com/\"; exit 1;)",
```
stereobooster: this is not friendly to open source contributors

.tsbuildinfo
TejasQ: It'll speed up CI builds though...

src/Icon/Icon.Brand.tsx
TejasQ: No. `BrandIcons` was an implementation detail used in `Icon`.

package.json
TejasQ: Let's remove this as soon as (if) we get the first issue complaining about it. 

image

.tsbuildinfo
TejasQ: Ah right. It won't because it has my path in it.

TejasQ: Will do. 👍

stereobooster: Does CI server have "/users/tejaskumar/sites/operational-ui" folder?

package.json
stereobooster: I know what this for. Yet this is not friendly for novice users. They want to tweak some small thing for example in readme. Run yarn start just out of curiosity and got updated files. It depends on people, but I would unroll those changes because I wanted to update readme only. If user is unfamiliar with git they will do it manually...

PR #1010, opened 2019-06-06
**Breaking:** Page title redesign.

ImogenF:
# Summary

Page titles redesigned/simplified as per the new specs. 

https://contiamo.atlassian.net/secure/RapidBoard.jspa?rapidView=43&projectKey=UI&modal=detail&selectedIssue=UI-50

Tabs will be updated separately, at which point the `condensedTitle` prop will also be considered.

# To be tested

Me

Tester 1

Tester 2

### DIFF
--- a/src/Page/Page.tsx
+++ b/src/Page/Page.tsx
@@ -6,8 +6,6 @@ import PageContent, { PageContentProps, isChildFunction } from "../PageContent/P
 import Progress from "../Progress/Progress"
 import { DefaultProps } from "../types"
 import { Title } from "../Typography/Title"
-import { readableTextColor } from "../utils"
-import { expandColor, OperationalStyleConstants } from "../utils/constants"
 import styled from "../utils/styled"
 
 export interface BaseProps extends DefaultProps {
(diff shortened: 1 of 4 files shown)


src/utils/constants.ts
```diff
@@ -151,7 +151,7 @@ const font = {
   lineHeight: 1.4,
   size: {
     /** 16px */
```
stereobooster: Please update comment as well

ImogenF: Good catch!

stereobooster:
New
Screenshot 2019-06-06 at 14 53 45

vs old
Screenshot 2019-06-06 at 14 53 55

There is no color difference in header and body, is it intended?

ImogenF:
@stereobooster According to the specs, doesn't look like there should be a colour difference. The specs all have white backgrounds though, so I could be wrong... 

image

src/Page/Page.tsx
```diff
@@ -17,10 +15,6 @@ export interface BaseProps extends DefaultProps {
   title?: string
   /** Page actions, typically `condensed button` component inside a fragment */
   actions?: React.ReactNode
-  /** Actions position */
-  actionsPosition?: "start" | "main" | "end"
```
stereobooster: this is breaking change (if I understand it right), so PR should be marked as **Breaking**

TejasQ:
# [We should probably add back the actions](https://contiamo.slack.com/archives/C0PF194MT/p1559741500014800)

PR #1018, opened 2019-06-11
**Breaking:** Page tabs

ImogenF:
# Summary

Update page content tabs to reflect new design. 

https://contiamo.atlassian.net/secure/RapidBoard.jspa?rapidView=43&projectKey=UI&modal=detail&selectedIssue=UI-52

Changes made:
- Style.
- Location - "condensed" tabs to the right of the title are no longer supported.
- Label truncation if too long.
- Page background changed to white to correspond to the new design and to ensure tabs don't look odd. 

# Related issue

# To be tested

Me

Tester 1

- [ ] Scrolling behaviour
- [ ] Label truncation

Tester 2

- [ ] Scrolling behaviour
- [ ] Label truncation

### DIFF
--- a/src/Internals/Tabs.tsx
+++ b/src/Internals/Tabs.tsx
@@ -3,6 +3,7 @@ import Spinner from "../Spinner/Spinner"
 import styled from "../utils/styled"
 import { IconComponentType } from "../Icon/Icon"
 import { inputFocus } from "../utils"
+import { useUniqueId } from "../useUniqueId"
 
 export interface Tab {
   name: string
@@ -101,6 +102,7 @@ const Tabs = ({ onTabChange, tabs, activeTabName, children }: Props) => {
   const activeTabIndex = getTabIndexByName(tabs, activeTabName)
   const [activeTab, setActiveTab] = useState(activeTabIndex)
   const [isMouseDown, setMouseDown] = useState(false)
+  const uid = useUniqueId()
 
   useEffect(() => {
     setActiveTab(activeTabIndex)
@@ -118,7 +120,9 @@ const Tabs = ({ onTabChange, tabs, activeTabName, children }: Props) => {
 
   const onKeyDown = useCallback(
     (event: React.KeyboardEvent) => {
-      const tabElements: NodeListOf<HTMLElement> = document.querySelectorAll('[role="tab"]')
+      const tabElements: NodeListOf<HTMLElement> = (document.querySelector(
+        `[id=${uid}]`,
+      ) as HTMLElement).querySelectorAll('[role="tab"]')
       let newIndex: number
       switch (event.key) {
         case "ArrowRight":
@@ -152,13 +156,20 @@ const Tabs = ({ onTabChange, tabs, activeTabName, children }: Props) => {
     [activeTab, onTabClick],
   )
 
+  const onFocus = (e: React.FocusEvent<HTMLDivElement>) => {
+    e.preventDefault()
+    if (isMouseDown) {
+      e.target.blur()
+    }
+  }
+
   // Work around: wrap return in fragment- to prevent type error and not having to change childrens return type
   // https://github.com/Microsoft/TypeScript/issues/21699
   return (
     <>
       {children({
         tabsBar: (
-          <TabsBar role="tablist" onKeyDown={onKeyDown}>
+          <TabsBar role="tablist" id={uid} onKeyDown={onKeyDown}>
             {tabs
               .filter(({ hidden }) => !hidden)
               .map((tab, index: number) => {
@@ -172,11 +183,7 @@ const Tabs = ({ onTabChange, tabs, activeTabName, children }: Props) => {
                     tabIndex={isActive ? 0 : -1}
                     onClick={() => onTabClick(index)}
                     onKeyDown={onKeyDown}
-                    onFocus={e => {
-                      if (isMouseDown) {
-                        e.target.blur()
-                      }
-                    }}
+                    onFocus={onFocus}
                     onMouseDown={() => setMouseDown(true)}
                     onMouseUp={() => setMouseDown(false)}
                   >
--- a/src/Internals/Tabs.tsx
+++ b/src/Internals/Tabs.tsx
@@ -29,41 +29,40 @@ const TabsBar = styled("div")(({ theme }) => ({
   display: "flex",
   alignItems: "flex-end",
   height: tabsBarHeight,
-  color: "inherit",
-  position: "inherit",
+  position: "relative",
   margin: `0 ${theme.space.element}px`,
   "&::before": {
     content: "''",
     display: "block",
     position: "absolute",
     width: "100%",
+    height: 0,
     top: tabsBarHeight - 1,
     left: 0,
-    borderBottom: `1px solid ${theme.color.border.default}`,
+    borderBottom: `1px solid ${theme.color.border.select}`,
   },
 }))
 
 const Tab = styled("div")<{ active?: boolean }>(({ theme, active }) => ({
-  display: "flex",
   width: 150,
   height: "100%",
-  backgroundColor: active ? "inherit" : theme.color.background.lighter,
-  color: active ? theme.color.text.action : "currentColor",
-  zIndex: 2,
-  textTransform: "capitalize",
+  display: "flex",
+  alignItems: "center",
+  backgroundColor: active ? theme.color.white : theme.color.background.lighter,
+  color: active ? theme.color.text.action : theme.color.text.dark,
   fontFamily: theme.font.family.main,
   fontSize: theme.font.size.small,
   fontWeight: active ? theme.font.weight.bold : theme.font.weight.regular,
   padding: `0px ${theme.space.element}px`,
-  border: `1px solid ${theme.color.border.default}`,
-  borderBottomColor: active ? "white" : "border.default",
+  border: `1px solid ${theme.color.border.select}`,
+  zIndex: 2,
+  borderBottomColor: active ? theme.color.white : theme.color.border.select,
   borderRightWidth: 0,
   ":last-child": {
     borderRightWidth: 1,
   },
   ":hover": {
     cursor: "pointer",
-    opacity: 1,
   },
 }))
 
@@ -116,7 +115,7 @@ const Tabs = ({ onTabChange, tabs, activeTabName, children }: Props) => {
                   ) : (
                     tab.icon && React.createElement(tab.icon, { size: 14, color: tab.iconColor, left: true })
                   )}
-                  {<TabName>{tab.name}</TabName>}
+                  <TabName>{tab.name}</TabName>
                 </Tab>
               ))}
           </TabsBar>
--- a/src/Internals/Tabs.tsx
+++ b/src/Internals/Tabs.tsx
@@ -158,7 +158,7 @@ const Tabs = ({ onTabChange, tabs, activeTabName, children }: Props) => {
     <>
       {children({
         tabsBar: (
-          <TabsBar role={"tablist"} onKeyDown={onKeyDown}>
+          <TabsBar role="tablist" onKeyDown={onKeyDown}>
             {tabs
               .filter(({ hidden }) => !hidden)
               .map((tab, index: number) => {
@@ -166,7 +166,7 @@ const Tabs = ({ onTabChange, tabs, activeTabName, children }: Props) => {
                 return (
                   <SingleTab
                     key={index}
-                    role={"tab"}
+                    role="tab"
                     active={isActive}
                     aria-selected={isActive}
                     tabIndex={isActive ? 0 : -1}
--- a/src/Internals/Tabs.tsx
+++ b/src/Internals/Tabs.tsx
@@ -35,12 +35,10 @@ export interface SingleTabProps {
   isKeyboardActive: boolean
 }
 
-const tabsBarHeight = 48
-
 const TabsBar = styled("div")(({ theme }) => ({
   display: "flex",
   alignItems: "flex-end",
-  height: tabsBarHeight,
+  height: theme.tabsBarHeight,
   position: "relative",
   margin: `0 ${theme.space.element}px`,
   // This draws the line underneath the tabs
@@ -50,7 +48,7 @@ const TabsBar = styled("div")(({ theme }) => ({
     position: "absolute",
     width: "100%",
     height: 0,
-    top: tabsBarHeight - 1,
+    top: theme.tabsBarHeight - 1,
     left: 0,
     borderBottom: `1px solid ${theme.color.border.select}`,
   },
--- a/src/Internals/Tabs.tsx
+++ b/src/Internals/Tabs.tsx
@@ -88,13 +88,13 @@ const TabContainer = styled("div")<{ active?: boolean }>(({ theme, active }) => 
   },
 }))
 
-const TabName = styled("p")(() => ({
+const TabName = styled("p")({
   width: "100%",
   whiteSpace: "nowrap",
   overflow: "hidden",
   textOverflow: "ellipsis",
   textAlign: "center",
-}))
+})
 
 const getTabIndexByName = (tabs: Tab[], tabName?: string): number => {
   if (tabName) {
--- a/src/Internals/Tabs.tsx
+++ b/src/Internals/Tabs.tsx
@@ -48,8 +48,6 @@ const Tab = styled("div")<{ active?: boolean }>(({ theme, active }) => ({
   display: "flex",
   width: 150,
   height: "100%",
-  alignItems: "center",
-  justifyContent: "center",
   backgroundColor: active ? "inherit" : theme.color.background.lighter,
   color: active ? theme.color.text.action : "currentColor",
   zIndex: 2,
@@ -68,6 +66,14 @@ const Tab = styled("div")<{ active?: boolean }>(({ theme, active }) => ({
     cursor: "pointer",
     opacity: 1,
   },
+}))
+
+const TabName = styled("p")(() => ({
+  width: "100%",
+  whiteSpace: "nowrap",
+  overflow: "hidden",
+  textOverflow: "ellipsis",
+  textAlign: "center",
 }))
 
 const getTabIndexByName = (tabs: Tab[], tabName?: string): number => {
@@ -111,7 +117,7 @@ const Tabs = ({ onTabChange, tabs, activeTabName, children }: Props) => {
                   ) : (
                     tab.icon && React.createElement(tab.icon, { size: 14, color: tab.iconColor, left: true })
                   )}
-                  {tab.name}
+                  {<TabName>{tab.name}</TabName>}
                 </Tab>
               ))}
           </TabsBar>
--- a/src/Internals/Tabs.tsx
+++ b/src/Internals/Tabs.tsx
@@ -24,6 +24,7 @@ export interface State {
 }
 
 export const tabsBarHeight = 48
+export const tabsBarMargin = 20
 
 const TabsBar = styled("div")(({ theme }) => ({
   display: "flex",
@@ -31,7 +32,7 @@ const TabsBar = styled("div")(({ theme }) => ({
   height: tabsBarHeight,
   color: "inherit",
   position: "inherit",
-  margin: theme.space.element,
+  margin: tabsBarMargin,
   "&::before": {
     content: "''",
     display: "block",
--- a/src/Internals/Tabs.tsx
+++ b/src/Internals/Tabs.tsx
@@ -103,6 +103,7 @@ const Tabs = ({ onTabChange, tabs, activeTabName, children }: Props) => {
   const [activeTab, setActiveTab] = useState(activeTabIndex)
   const [isMouseDown, setMouseDown] = useState(false)
   const uid = useUniqueId()
+  const tabsList = tabs.map(() => React.useRef<HTMLDivElement>(null))
 
   useEffect(() => {
     setActiveTab(activeTabIndex)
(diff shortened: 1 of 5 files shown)


src/Internals/Tabs.tsx
```diff
@@ -16,44 +16,66 @@ export interface Props {
...
 
-const TabsBar = styled("div")<{ condensed?: boolean }>(({ theme, condensed }) => ({
+const TabsBar = styled("div")(({ theme }) => ({
   display: "flex",
   alignItems: "flex-end",
-  height: condensed ? theme.titleHeight : tabsBarHeight,
+  height: tabsBarHeight,
   color: "inherit",
-  ...(condensed ? { paddingLeft: theme.space.element } : {}),
+  position: "inherit",
```
TejasQ: What does this do?

src/Internals/Tabs.tsx
```diff
@@ -16,44 +16,66 @@ export interface Props {
...
+  margin: tabsBarMargin,
```
TejasQ: This is 

```suggestion
  margin: theme.space.element,
```

src/Internals/Tabs.tsx
```diff
@@ -16,44 +16,66 @@ export interface Props {
...
   children: (childrenConfig: { tabsBar: React.ReactNode; activeChildren: React.ReactNode }) => React.ReactNode
 }
 
 export interface State {
   activeTab: number
 }
 
-export const tabsBarHeight = 40
+export const tabsBarHeight = 48
+export const tabsBarMargin = 20
```
TejasQ: This comes from the theme.

```suggestion
```

src/Internals/Tabs.tsx
```diff
@@ -16,44 +16,66 @@ export interface Props {
...
+const TabsBar = styled("div")(({ theme }) => ({
   display: "flex",
   alignItems: "flex-end",
-  height: condensed ? theme.titleHeight : tabsBarHeight,
+  height: tabsBarHeight,
   color: "inherit",
-  ...(condensed ? { paddingLeft: theme.space.element } : {}),
+  position: "inherit",
+  margin: tabsBarMargin,
+  "&::before": {
```
TejasQ: Nice technique. Maybe it's worth setting `height: 1px` here?

src/Internals/Tabs.tsx
```diff
@@ -16,44 +16,66 @@ export interface Props {
...
+    top: tabsBarHeight - 1,
+    left: 0,
+    borderBottom: `1px solid ${theme.color.border.default}`,
+  },
 }))
 
-const Tab = styled("div")<{ active?: boolean; condensed?: boolean }>(({ theme, active }) => ({
+const Tab = styled("div")<{ active?: boolean }>(({ theme, active }) => ({
   display: "flex",
+  width: 150,
```
TejasQ: Hmmm do we want tabs to extend in case of longer names? cc @micha-f

src/Internals/Tabs.tsx
```diff
@@ -16,44 +16,66 @@ export interface Props {
...
   height: "100%",
-  alignItems: "center",
-  justifyContent: "center",
-  color: "currentColor",
-  opacity: active ? 1 : 0.8,
-  textTransform: "uppercase",
+  backgroundColor: active ? "inherit" : theme.color.background.lighter,
```
TejasQ: Inherit from where?

src/Internals/Tabs.tsx
```diff
@@ -16,44 +16,66 @@ export interface Props {
...
+  color: active ? theme.color.text.action : "currentColor",
+  zIndex: 2,
```
TejasQ: Why the `zIndex` if it's not positioned absolutely?

src/Internals/Tabs.tsx
```diff
@@ -16,44 +16,66 @@ export interface Props {
...
+  textTransform: "capitalize",
```
TejasQ: Will this work in scenarios where English isn't the primary language? Probably best to leave this up to userland.

src/Internals/Tabs.tsx
```diff
@@ -16,44 +16,66 @@ export interface Props {
...
   fontFamily: theme.font.family.main,
   fontSize: theme.font.size.small,
-  fontWeight: theme.font.weight.medium,
+  fontWeight: active ? theme.font.weight.bold : theme.font.weight.regular,
   padding: `0px ${theme.space.element}px`,
-  borderBottom: "2px solid",
-  borderBottomColor: active ? "currentColor" : "transparent",
+  border: `1px solid ${theme.color.border.default}`,
+  borderBottomColor: active ? "white" : "border.default",
```
TejasQ: `"border.default"` is not a valid border color...

src/Internals/Tabs.tsx
```diff
@@ -85,17 +107,17 @@ const Tabs = ({ onTabChange, tabs, activeTabName, condensed, children }: Props)
...
               .map((tab, index: number) => (
-                <Tab condensed={condensed} key={index} active={activeTab === index} onClick={() => onTabClick(index)}>
+                <Tab key={index} active={activeTab === index} onClick={() => onTabClick(index)}>
                   {tab.loading ? (
                     <Spinner left size={14} />
                   ) : (
                     tab.icon && React.createElement(tab.icon, { size: 14, color: tab.iconColor, left: true })
                   )}
-                  {tab.name}
+                  {<TabName>{tab.name}</TabName>}
```
TejasQ:
```suggestion
                  <TabName>{tab.name}</TabName>
```

src/Page/Page.tsx
```diff
@@ -82,16 +78,14 @@ const TitleContainer = styled("div")(({ theme }) => ({
...
-    const calculatedTitleOffset = isInTab && !isTitleCondensed ? theme.titleHeight + tabsBarHeight : theme.titleHeight
-    return {
-      height: hasTitle ? `calc(100% - ${calculatedTitleOffset}px)` : "100%",
-      overflow: "hidden",
-      position: "relative",
-    }
-  },
-)
+const ViewContainer = styled("div")<{ isInTab?: boolean; hasTitle?: boolean }>(({ theme, isInTab, hasTitle }) => {
+  const calculatedTitleOffset = isInTab ? theme.titleHeight + tabsBarHeight + tabsBarMargin * 2 : theme.titleHeight
```
TejasQ:
```suggestion
  const calculatedTitleOffset = isInTab ? theme.titleHeight + tabsBarHeight + theme.space.element * 2 : theme.titleHeight
```

src/Page/Page.tsx
```diff
@@ -1,6 +1,6 @@
 import * as React from "react"
 
-import Tabs, { Tab, tabsBarHeight } from "../Internals/Tabs"
+import Tabs, { Tab, tabsBarHeight, tabsBarMargin } from "../Internals/Tabs"
```
TejasQ: This value comes from the theme.

```suggestion
import Tabs, { Tab, tabsBarHeight } from "../Internals/Tabs"
```

TejasQ:
## Visual Review

- Content in tabs is not centered vertically: 
image

- Colors (borders and backgrounds) are not consistent with [the design](https://app.zeplin.io/project/5cc31fad6f17562d6dac9a53/screen/5ce397351f5a851e1140eb27). Please feel free to override constants in the code to make them conform to the design.

src/Internals/Tabs.tsx
ImogenF: I wanted to keep the margin as a separate constant because if it changes here, it also needs to change here: https://github.com/contiamo/operational-ui/blob/UI-52-page-tabs/src/Page/Page.tsx#L82. I could change both instances to `theme.space.element` but then if it's changed to `theme.space.content` in one place but not the other, the layout will break.

src/Internals/Tabs.tsx
ImogenF: Had a discussion with Kemal about this already. He said fixed width, truncated text. 

image

I figured that it makes most sense to do the truncation based on the width of the tab, rather than number of characters.

ImogenF:
@TejasQ re: the colours, I don't feel comfortable changing the constants without knowing for sure that that won't impact other components (which it inevitably will). Is Kemal working with a set, small colour palette? I.e. are borders and backgrounds always one of a select few colours? If so, I think it's worth talking to him and getting the full list, and then going through the components one-by-one. Happy to do that, if it hasn't already been done to some extent.

src/Internals/Tabs.tsx
ImogenF: It inherits the page's background colour.

ImogenF: Although tbh, if the page's background colour changes, a lot of other colours like the tab text colours need to be changed as well...

src/Internals/Tabs.tsx
TejasQ: That's a really interesting problem. This sounds a bit like coupling between `Tabs` and `Page` that probably should be avoided. Is there a way we can have _only this_ component be responsible for its layout? 🤔

src/Internals/Tabs.tsx
TejasQ: Related discussion: https://github.com/contiamo/operational-ui/pull/1018#discussion_r292553119

src/Internals/Tabs.tsx
TejasQ: Ok.

src/Page/Page.tsx
TejasQ: Related discussion: https://github.com/contiamo/operational-ui/pull/1018#discussion_r292553119

src/Page/Page.tsx
TejasQ: Related discussion: https://github.com/contiamo/operational-ui/pull/1018#discussion_r292553119

TejasQ:
@ImogenF 

> Is Kemal working with a set, small colour palette? I.e. are borders and backgrounds always one of a select few colours?
Yes. The plan is to slowly replace our constants with his newer colors. (cc @micha-f)

> If so, I think it's worth talking to him and getting the full list, and then going through the components one-by-one. Happy to do that, if it hasn't already been done to some extent.
It would be **lovely** if you'd could do this. Thanks so much!

src/Internals/Tabs.tsx
ImogenF: The page needs to know how much space has been taken up by the tabs bar, so that it knows how much space is left for the other components. Can't see an obvious way to avoid this bit of communication between the two, apart from making the `tabsBarHeight` constant inclusive of the margin...

TejasQ: As discussed in real life, it seems [CSS Grid Layout](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Grid_Layout) is the solution: let `Page` as the container of `Tabs` dictate how it shows `Tabs`, and `Tabs` has no spacing information around it.

ImogenF:
@TejasQ Have address your comments and added accessibility functionality and corresponding cypress tests.

src/Internals/Tabs.tsx
```diff
@@ -23,48 +23,46 @@ export interface State {
...
 
 const TabsBar = styled("div")(({ theme }) => ({
   display: "flex",
   alignItems: "flex-end",
   height: tabsBarHeight,
-  color: "inherit",
-  position: "inherit",
-  margin: tabsBarMargin,
+  position: "relative",
+  margin: `0 ${theme.space.element}px`,
```
TejasQ: Why? What if we remove the margin entirely and leave its positioning up to its parent?

src/Internals/Tabs.tsx
```diff
@@ -23,48 +23,46 @@ export interface State {
...
+  backgroundColor: active ? theme.color.white : theme.color.background.lighter,
+  color: active ? theme.color.text.action : theme.color.text.dark,
   fontFamily: theme.font.family.main,
   fontSize: theme.font.size.small,
   fontWeight: active ? theme.font.weight.bold : theme.font.weight.regular,
   padding: `0px ${theme.space.element}px`,
-  border: `1px solid ${theme.color.border.default}`,
-  borderBottomColor: active ? "white" : "border.default",
+  border: `1px solid ${theme.color.border.select}`,
+  zIndex: 2,
```
TejasQ: Should the `zIndex` live in `theme.zIndex`?

src/Internals/Tabs.tsx
ImogenF: The parent controls the spacing between the elements, sure. But this is the spacing within the element. In this case, we want the tabs to have padding on the left of the tabs container, to align everything with the title and the content container, both of which have their own padding.

src/Internals/Tabs.tsx
ImogenF: The `zIndex` in this case is a very specific use case, to ensure the active tab is on top of the other tabs (so the highlight is displayed correctly), and so all the tabs are on top of the full-width bottom border. I could use `theme.zIndex.default`, `theme.zIndex.default + 1` etc but I don't think adding new options makes sense, nor does using `zIndex.??` since none of the other options correspond to this situation.

src/Internals/Tabs.tsx
```diff
@@ -79,25 +116,79 @@ const Tabs = ({ onTabChange, tabs, activeTabName, condensed, children }: Props)
...
+  )
+
   // Work around: wrap return in fragment- to prevent type error and not having to change childrens return type
   // https://github.com/Microsoft/TypeScript/issues/21699
   return (
     <>
       {children({
         tabsBar: (
-          <TabsBar condensed={condensed}>
+          <TabsBar role={"tablist"} onKeyDown={onKeyDown}>
```
TejasQ:
```suggestion
          <TabsBar role="tablist" onKeyDown={onKeyDown}>
```

src/Internals/Tabs.tsx
```diff
@@ -79,25 +116,79 @@ const Tabs = ({ onTabChange, tabs, activeTabName, condensed, children }: Props)
...
-                  )}
-                  {tab.name}
-                </Tab>
-              ))}
+              .map((tab, index: number) => {
+                const isActive: boolean = activeTab === index
+                return (
+                  <SingleTab
+                    key={index}
+                    role={"tab"}
```
TejasQ:
```suggestion
                    role="tab"
```

src/Internals/Tabs.tsx
```diff
@@ -79,25 +116,79 @@ const Tabs = ({ onTabChange, tabs, activeTabName, condensed, children }: Props)
     [onTabChange, tabs, setActiveTab],
   )
 
+  const onKeyDown = useCallback(
+    (event: React.KeyboardEvent) => {
+      const tabElements: NodeListOf<HTMLElement> = document.querySelectorAll('[role="tab"]')
```
stereobooster: it can be a problem if there is more than one tabs component on the screen

src/Internals/Tabs.tsx
```diff
@@ -79,25 +116,79 @@ const Tabs = ({ onTabChange, tabs, activeTabName, condensed, children }: Props)
...
+                  <SingleTab
+                    key={index}
+                    role={"tab"}
+                    active={isActive}
+                    aria-selected={isActive}
+                    tabIndex={isActive ? 0 : -1}
+                    onClick={() => onTabClick(index)}
+                    onKeyDown={onKeyDown}
+                    onFocus={e => {
+                      if (isMouseDown) {
```
stereobooster: you can as well do e.preventDefault() on `onMouseDown`

src/Page/Page.tsx
```diff
@@ -62,16 +60,16 @@ export interface PropsWithTabs extends BaseProps {
...
 
 export type PageProps = PropsWithSimplePage | PropsWithComplexPage | PropsWithTabs
 
 const Container = styled("div")(({ theme }) => ({
   height: "100%",
   position: "relative",
-  backgroundColor: theme.color.background.lighter,
+  display: "grid",
+  gridRowGap: theme.space.element,
+  backgroundColor: theme.color.white,
```
TejasQ: We should maybe add

```jsx
gridTemplateRows: `${pageHeaderHeight}px calc(100% - ${pageHeaderHeight}px)`,
```

or something similar.

src/Internals/Tabs.tsx
```diff
@@ -79,25 +118,84 @@ const Tabs = ({ onTabChange, tabs, activeTabName, condensed, children }: Props)
     [onTabChange, tabs, setActiveTab],
   )
 
+  const onKeyDown = useCallback(
+    (event: React.KeyboardEvent) => {
+      const tabElements: NodeListOf<HTMLElement> = (document.querySelector(
```
stereobooster: this is not eloquent (if I may say so) React code. We can do this with React reference. Not sure if anything would break for our use case, but  I can think of some cases where it will break, for example when there is no DOM (server side rendering)

TejasQ: >We can do this with React reference.

[Obligatory link to docs](https://reactjs.org/docs/hooks-reference.html#useref).

ImogenF: Good catch - I had forgotten about React reference 🙄

src/Internals/Tabs.tsx
```diff
@@ -103,6 +103,7 @@ const Tabs = ({ onTabChange, tabs, activeTabName, children }: Props) => {
   const [activeTab, setActiveTab] = useState(activeTabIndex)
   const [isMouseDown, setMouseDown] = useState(false)
   const uid = useUniqueId()
+  const tabsList = tabs.map(() => React.useRef<HTMLDivElement>(null))
```
stereobooster: this will break if number of tabs change across the render

src/Page/Page.tsx
```diff
@@ -118,30 +111,24 @@ class Page extends React.Component<PageProps, Readonly<typeof initialState>> {
...
                   <ActionsContainer>{actions}</ActionsContainer>
                 </TitleContainer>
-                {!condensedTitle && tabsBar}
+                {tabsBar}
               </>
             ) : (
               tabsBar
             )}
-            <ViewContainer isInTab isTitleCondensed={condensedTitle} hasTitle>
+            <ViewContainer role="tabpanel" aria-labelledby={this.props.activeTabName} tabIndex={0}>
```
stereobooster: I'm not sure if this value represents only initial state or it changes when tabs switched. If it doesn't change this is an error

src/Internals/Tabs.tsx
```diff
@@ -62,12 +98,47 @@ const getTabIndexByName = (tabs: Tab[], tabName?: string): number => {
...
+  )
+}
+
+const Tabs = ({ onTabChange, tabs, activeTabName, children }: Props) => {
   const activeTabIndex = getTabIndexByName(tabs, activeTabName)
   const [activeTab, setActiveTab] = useState(activeTabIndex)
-  useEffect(() => {
-    setActiveTab(activeTabIndex)
-  }, [activeTabIndex])
+  const uid = useUniqueId()
```
stereobooster: we don't really need it

src/Internals/Tabs.tsx
stereobooster: You can use box-shadow: inset instead of border, that way you don't need to manage z-index (if this work for the design)

src/Internals/Tabs.tsx
```diff
@@ -16,44 +18,78 @@ export interface Props {
...
+const TabsBar = styled("div")(({ theme }) => ({
   display: "flex",
   alignItems: "flex-end",
-  height: condensed ? theme.titleHeight : tabsBarHeight,
-  color: "inherit",
-  ...(condensed ? { paddingLeft: theme.space.element } : {}),
+  height: tabsBarHeight,
+  position: "relative",
+  margin: `0 ${theme.space.element}px`,
+  "&::before": {
```
stereobooster: can you add small comment what this element does? Is it to draw the line above tabs?

stereobooster:
```diff
+ import { SearchIcon } from "@operational/components"
...
- { name: "overview", children: <Tab title="Overview" />, icon: "Search" },
+ { name: "overview", children: <Tab title="Overview" />, icon: SearchIcon },
```

We need to change Icon in example with tabs (in Page readme)

src/Page/Page.tsx
ImogenF: It doesn't chance, but why is this an error? It's showing the page content for the currently selected tab...

stereobooster: `aria-labelledby` is an id of corresponding tab according to spec, right?

Screenshot 2019-06-14 at 14 10 05

Content of the tab corresponds to "overview" tab, but it says it is labelled by "jobs" tabs.

ImogenF: Ah sorry, I thought you were referring to the `tabIndex`. Good catch.

fabien0102:
image

We have a strange overflow over there

fabien0102:
image

This example is totally broken, tabs are not responding to the router anymore

ImogenF:
@fabien0102 I could have sworn I fixed the overflow... @TejasQ is that be linked to your last 2 commits?

TejasQ:
@ImogenF sorry about that. The `max-content` stretches the tabs to a height that is incorrect and forces the tab children to be incorrectly low.

Feel free to try it out locally. You'll need:

1. A page with tabs
1. A small card as a tab's child.

It may be good to add a README example with this reproduction.

TejasQ:
Example with `max-content`

image

Example with explicit height

image

TejasQ:
Let's also add `outline: none` to the tab panel to avoid this odd blue outline that @micha-f doesn't like

image

ImogenF:
I don't know how to recreate this issue. Does `outline: none` definitely solve the problem? 

> Let's also add `outline: none` to the tab panel to avoid this odd blue outline that @micha-f doesn't like
> 
> image

fabien0102:
Nice for the routing fix 👌

The last think to me to be fixed before merging is this question around `Card`

image

For a bit more context, `Card` and `Page` share some logic with `Internals/Tabs` so we need to think about this also 😅

ImogenF:
@fabien0102 I spoke to Tejas about the `Card` issue. I offered to put the old tabs back in for cards and have page tabs as a separate component, but he was of the opinion that there's no point, since it will all be affected by https://contiamo.atlassian.net/browse/UI-74 anyway.

So just the random blue line to fix now.

ImogenF:
The blue outline is not as simple as just setting `outline: blue` - it's another one of these situations where we want the outline if we're tabbing through the page contents using the keyboard, but not on click. Either we can remove the `tabIndex` for now, and deal with this at a later stage, or I'll get it fixed first thing in the morning.

micha-f:
Later stage is fine I would say. Let's not spend too much time on this right now.

TejasQ:
> @fabien0102 I spoke to Tejas about the Card issue. I offered to put the old tabs back in for cards and have page tabs as a separate component, but he was of the opinion that there's no point, since it will all be affected by https://contiamo.atlassian.net/browse/UI-74 anyway.

Yup. I'm removing all the cards and we'll figure something out soon.

TejasQ:
> I don't know how to recreate this issue. Does outline: none definitely solve the problem?

@ImogenF the example **"with tabs and no title"** here displays this issue when you click on the tab's content: https://deploy-preview-1018--operational-ui.netlify.com/#!/Page

image

Issue #1017, opened 2019-06-11
Breadcrumb with icon looks bad


micha-f:
![image](https://user-images.githubusercontent.com/200647/59267530-c9a4cc80-8c4a-11e9-9b20-e8a27284fe9a.png)

We need to improve the look here by adjusting spacing and alignment.

ImogenF:
Am redoing the breadcrumbs now so will make sure this is fixed at the same time.

PR #1020, opened 2019-06-11
**Breaking:** Breadcrumbs

ImogenF:
# Summary

Update breadcrumbs to correspond to new designs.

https://contiamo.atlassian.net/secure/RapidBoard.jspa?rapidView=43&projectKey=UI&modal=detail&selectedIssue=UI-51

Breadcrumbs are now required to have a "to" prop - they must link to somewhere. 

# Related issue

https://github.com/contiamo/operational-ui/issues/1017

# To be tested

Me

Tester 1

Tester 2


stereobooster:
Not a big deal but left caret icon is not clickable ( thought is has the same color so it seems it supposed to be clickable).

TejasQ:
Looks nice, let's add a small set of e2e tests to make sure they behave the same when we move code around in the future.

ImogenF:
@TejasQ Remind me again which ones the e2e tests are? The cypress ones? Or the snapshots?

ImogenF:
@stereobooster The caret is now clickable but currently without the cursor pointer, because setting that css option seems to override the `Icon` component styles...

TejasQ:
@ImogenF e2e (end-to-end) are Cypress because it simulates a user. Snapshots are unit tests (Jest).

src/Breadcrumb/Breadcrumb.tsx
```diff
@@ -3,62 +3,47 @@ import { IconComponentType } from "../Icon/Icon"
...
 import styled from "../utils/styled"
 
 export interface BreadcrumbProps extends DefaultProps {
   /** Navigation property à la react-router <Link/> */
-  to?: string
+  to: string
   onClick?: (ev?: React.SyntheticEvent<React.ReactNode>) => void
   children?: React.ReactNode
   icon?: IconComponentType
+  iconLeft?: boolean
```
TejasQ: It's not immediately obvious what this does.

src/Breadcrumb/Breadcrumb.tsx
```diff
@@ -3,62 +3,49 @@ import { IconComponentType } from "../Icon/Icon"
...
+  to: string
   onClick?: (ev?: React.SyntheticEvent<React.ReactNode>) => void
   children?: React.ReactNode
   icon?: IconComponentType
+  /** Optionally moves icon from default position on right, to before the text, including appropriate padding */
+  iconLeft?: boolean
 }
 
-const containerStyles = {
+const Container = styled("a")(({ theme }) => ({
```
stereobooster:
```suggestion
const Link = styled("a")(({ theme }) => ({
```
?

src/Breadcrumb/Breadcrumb.tsx
```diff
@@ -3,62 +3,49 @@ import { IconComponentType } from "../Icon/Icon"
...
-      )}
-    </OperationalContext>
-  )
-}
+const Breadcrumb: React.SFC<BreadcrumbProps> = ({ to, icon: Icon, onClick, ...props }) => (
+  <OperationalContext>
+    {ctx => (
+      <Container
+        {...props}
+        tabIndex={0}
```
stereobooster: if it is `a` it should be focusable by default, so this is redundant I guess

TejasQ: Cypress doesn't know that it's focusable sadly 😢

TejasQ: I agree

```suggestion
        tabIndex={0} // for cypress
```

could make sense

src/Breadcrumb/Breadcrumb.tsx
```diff
@@ -3,62 +3,49 @@ import { IconComponentType } from "../Icon/Icon"
...
+        href={to}
+        onClick={(ev: React.SyntheticEvent<Node>) => {
+          if (onClick) {
```
stereobooster: Allow Ctrl/Cmd click e.g. in case click event has Ctrl or Cmd modifier don't trigger handler, but allow default browser action instead

src/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -1,43 +1,45 @@
...
+    ? [head, <Slash key={`divider-${index}`}>/</Slash>, ...intersperseSlashes(index + 1)(tail)]
+    : [head]
 
-const Breadcrumbs: React.SFC<BreadcrumbsProps> = props => (
-  <Container {...props}>{intersperseSlashes(0)(React.Children.toArray(props.children))}</Container>
-)
+const Breadcrumbs: React.SFC<BreadcrumbsProps> = props => {
+  const items = React.Children.toArray(props.children)
+  return (
+    <Container aria-label={"Breadcrumb"} {...props}>
```
stereobooster: 🤔  is this helpful label? If we want more semantics we can use `<nav>` element

TejasQ: It's not helpful. I suggest `aria-label={children}` or similar if we have access to the actual text, or we can use a `nav`.

src/Breadcrumb/Breadcrumb.tsx
TejasQ: This makes it work with the router: https://github.com/contiamo/operational-ui/pull/1020/files#diff-7ec342830a821c48321c314331555858R39

src/Breadcrumb/Breadcrumb.tsx
TejasQ: @stereobooster this seems to work with Cmd + click naturally 🤔

src/Breadcrumbs/Breadcrumbs.tsx
TejasQ: I was so wrong. This actually implements the spec perfectly. My bad.

TejasQ: We can perhaps do this though

```suggestion
    <Container aria-label="Breadcrumb" {...props}>
```

src/Breadcrumb/Breadcrumb.tsx
ImogenF: @TejasQ @stereobooster are you both happy with this?

PR #1028, opened 2019-06-13
**Breaking:** Update SidenavItems to support nested items

TejasQ:
This pull request updates the `SidenavItem` component in a big way, allowing it to contain a list of `items` and function similar to a Select box, allowing our users to choose projects and other superpowers.

### To Test
- [x] is accessible
- [x] looks good
- [x] handles a large number of items
- [x] no regression on Select
- [x] no regression on TopBar
- [x] no regression on ContextMenu


fabien0102:
The multiselect behaviour is a bit strange, it close everytime…
04ba441d-d000-4892-ad9a-ed6d928ff08d

(But I'm not sure it's new, just part of my review scope ^^)

TejasQ:
@fabien0102 great catch! That's a regression!

PR #1031, opened 2019-06-13
**Fix:** Fix Foldable types

fabien0102:
# Summary

This should be a `React.ReactElement`

# Related issue

-

# To be tested

Me

Tester 1

Tester 2


src/Foldable/Foldable.tsx
```diff
@@ -10,7 +10,7 @@ export interface FoldableProps {
   }: {
     Toggler: React.ComponentType<TogglerProps>
     isFolded: boolean
-  }) => React.ReactNode
+  }) => React.ReactElement
```
TejasQ: This is the wrong type

PR #1039, opened 2019-06-17
  **Breaking:** Change accordion API

stereobooster:
# Summary

Change accordion API as discussed with @fabien0102 

# Related issue

UI-77

# To be tested

Me

Tester 1

Tester 2


src/AccordionSection/AccordionSection.tsx
```diff
@@ -0,0 +1,108 @@
...
+import styled from "../utils/styled"
+import { ChevronUpIcon, ChevronDownIcon } from "../Icon/Icon"
+import { DefaultProps } from "../types"
+import isFunction from "lodash/isFunction"
+
+export interface AccordionSectionProps extends DefaultProps {
+  title: React.ReactNode
+  children: (() => React.ReactNode) | React.ReactNode
+  // for internal use
+  expanded: boolean
```
stereobooster: @fabien0102 shall I name those differently? I can them with the help of Context API to not expose to end user

stereobooster:
@fabien0102 I saw it. Do you know what is this? I can't spot the issue

src/Accordion/README.md
```diff
@@ -10,7 +10,7 @@ import { Accordion, AccordionSection } from "@operational/components"
 
 const MyComponent = () => {
   const [expanded, setExpanded] = React.useState([true, false, false])
-  const onToggle = (index: number) => {
+  const onToggle = index => {
```
TejasQ: image

why?

src/Accordion/README.md
```diff
@@ -6,51 +6,42 @@ Accordion component implemented (mostly) according to [WAI-ARIA specification](h
...
+    const newExpanded = [...expanded]
+    newExpanded[index] = !newExpanded[index]
+    setExpanded(newExpanded)
+  }
   return (
     <div style={{ height: 400 }}>
-      <Accordion sections={sections} onToggle={onToggle} />
+      <Accordion expanded={expanded} onToggle={onToggle}>
+        <AccordionSection title="Section 1">
+          {() => (
```
TejasQ: Why is this a function? Why can't we just have a ReactNode in here?

src/AccordionSection/AccordionSection.tsx
```diff
@@ -0,0 +1,112 @@
...
+  title,
+  children,
+  isMouseRef,
+}) => {
+  const titleId = `${id}-${index}-heading`
+  const contentId = `${id}-${index}-panel`
+  const [focusFlag, setFocusFlag] = React.useState(false)
+
+  if (expanded === undefined || index === undefined || isMouseRef === undefined || toggleSection === undefined) {
+    throw new Error("Use AccordionSection inside Accordion")
```
TejasQ:
```suggestion
    throw new Error("Only AccordionSections can be used inside Accordion. See https://operational-ui.netlify.com/#!/Accordion for more info.")
```

src/Accordion/README.md
stereobooster: Can you be more specific?

src/Accordion/README.md
stereobooster: we can have both

src/Accordion/README.md
TejasQ: I do not understand this commit: https://github.com/contiamo/operational-ui/pull/1039/commits/728f18ffce5c595dd073308faeb38b58bc879de1

It says TS breaks the example and removes the type annotation from `index`. But from my screenshot this works.

Why did you remove it with that commit?

src/Accordion/README.md
TejasQ: Can we keep the README example simple and remove it in that case?

src/Accordion/README.md
stereobooster: You edit directly in playground right? Try to edit readme itself and you will get screen like @fabien0102 posted above.

image

src/Accordion/README.md
stereobooster: sure thing, I thought I would show that you can use both (rows bellow use ReactNode)

src/Accordion/README.md
TejasQ: Interesting.

src/AccordionSection/AccordionSection.tsx
```diff
@@ -0,0 +1,112 @@
...
+import styled from "../utils/styled"
+import { ChevronUpIcon, ChevronDownIcon } from "../Icon/Icon"
+import { DefaultProps } from "../types"
+import isFunction from "lodash/isFunction"
+
+export interface AccordionSectionProps extends DefaultProps {
+  title: React.ReactNode
+  children: (() => React.ReactNode) | React.ReactNode
+  // for internal use
+  index?: number
```
TejasQ: Maybe prefix with an `_`?

```suggestion
  _index?: number
```

PR #1043, opened 2019-06-18
**Feature:** DropdownButton component.

ImogenF:
# Summary

Button with down caret that triggers a `ContextMenu` dropdown. 

A few comments about this to bear in mind:

1) I've used the existing `ContextMenu` component, so the ARIA functionality provided by `useListBox` is already in there. 
2) I've set it up so that the `ComboButton` adds the right carets to the list items automatically, because that appears to be a defining feature of the component. I could, however, remove that and leave that up to the user. 
3) The styling looks a bit off because the first item of the dropdown is focussed/highlighted by default. I think this is an issue with `ContextMenu` anyway so I haven't addressed it here.

# Jira ticket

https://contiamo.atlassian.net/secure/RapidBoard.jspa?rapidView=43&projectKey=UI&modal=detail&selectedIssue=UI-75&quickFilter=56

# To be tested

Me

Tester 1

Tester 2

### DIFF
--- a/src/ComboButton/ComboButton.tsx
+++ b/src/ComboButton/ComboButton.tsx
@@ -1,0 +1,123 @@
+import * as React from "react"
+
+import { DefaultProps } from "../types"
+import { darken, inputFocus } from "../utils"
+import styled from "../utils/styled"
+import { CaretDownIcon, CaretRightIcon } from "../Icon/Icon"
+import ContextMenu, { ContextMenuProps } from "../ContextMenu/ContextMenu"
+
+export interface ComboButtonProps extends DefaultProps {
+  /** Title */
+  title?: string
+  /** Actions to display in dropdown */
+  items: ContextMenuProps["items"]
+  /** Disabled option */
+  disabled?: boolean
+  /** What is the tab index, for accessibility? */
+  tabIndex?: number
+}
+
+const BaseComboButton = styled("button")<{ disabled?: ComboButtonProps["disabled"]; isOpen: boolean }>(
+  ({ theme, disabled, isOpen }) => {
+    return {
+      backgroundColor: theme.color.primary,
+      lineHeight: "36px",
+      fontSize: theme.font.size.small,
+      fontFamily: theme.font.family.main,
+      fontWeight: theme.font.weight.medium,
+      color: theme.color.text.white,
+      display: "inline-flex",
+      alignItems: "center",
+      justifyContent: "center",
+      padding: `0 0 0 ${theme.space.element}px`,
+      borderRadius: theme.borderRadius,
+      border: 0,
+      cursor: disabled ? "not-allowed" : "pointer",
+      opacity: disabled ? 0.6 : 1.0,
+      outline: "none",
+      position: "relative",
+      marginRight: theme.space.small,
+      ...(isOpen
+        ? {
+            backgroundColor: theme.color.white,
+            color: theme.color.text.default,
+            borderColor: theme.color.border.select,
+          }
+        : {
+            backgroundColor: theme.color.primary,
+            color: theme.color.text.white,
+            borderColor: theme.color.border.invisible,
+          }),
+      // Apply styles with increased specificity to override defaults
+      ":focus": {
+        ...inputFocus({ theme }),
+      },
+      ...(!disabled
+        ? {
+            ":hover": {
+              backgroundColor: isOpen ? theme.color.white : darken(theme.color.primary, 5),
+            },
+          }
+        : {}),
+    }
+  },
+)
+
+const CaretContainer = styled("div")<{ isOpen: boolean }>(({ isOpen, theme }) => {
+  return {
+    width: "36px",
+    marginLeft: theme.space.content,
+    borderLeft: `1px solid ${isOpen ? theme.color.primary : theme.color.text.white}`,
+  }
+})
+
+const ItemWithCaret = styled("div")`
+  width: 100%;
+  display: flex;
+  align-items: baseline;
+  justify-content: space-between;
+`
+
+const ComboButton: React.SFC<ComboButtonProps> = ({ items, title, tabIndex, ...props }) => {
+  // Dropdown buttons always have right carets for every item. These are
+  const itemsWithCarets = items.map(item => {
+    return {
+      label: (
+        <ItemWithCaret>
+          {typeof item === "string" ? item : item.label}
+          <CaretRightIcon size={8} />
+        </ItemWithCaret>
+      ),
+    }
+  })
+
+  return (
+    <ContextMenu {...props} iconLocation="right" items={itemsWithCarets}>
+      {isOpen => {
+        return (
+          <BaseComboButton
+            {...props}
+            role="button"
+            aria-label={title}
+            tabIndex={tabIndex}
+            onClick={(ev: React.SyntheticEvent<React.ReactNode>) => {
+              if (props.disabled) {
+                ev.preventDefault()
+                return
+              }
+            }}
+            title={title}
+            isOpen={isOpen}
+          >
+            {title}
+            <CaretContainer isOpen={isOpen}>
+              <CaretDownIcon size={8} />
+            </CaretContainer>
+          </BaseComboButton>
+        )
+      }}
+    </ContextMenu>
+  )
+}
+
+export default ComboButton
--- a/src/ComboButton/ComboButton.tsx
+++ b/src/ComboButton/ComboButton.tsx
@@ -1,75 +1,41 @@
 import * as React from "react"
 
 import { DefaultProps } from "../types"
-import { darken, inputFocus } from "../utils"
+import { expandColor } from "../utils/constants"
 import styled from "../utils/styled"
 import { CaretDownIcon, CaretRightIcon } from "../Icon/Icon"
 import ContextMenu, { ContextMenuProps } from "../ContextMenu/ContextMenu"
+import Button from "../Button/Button"
+import { IContextMenuItem } from "../ContextMenu/ContextMenu.Item"
 
 export interface ComboButtonProps extends DefaultProps {
-  /** Title */
-  title?: string
+  /** Button color theme (hex or named color from `theme.color`) */
+  color: string
   /** Actions to display in dropdown */
   items: ContextMenuProps["items"]
   /** Disabled option */
   disabled?: boolean
   /** What is the tab index, for accessibility? */
   tabIndex?: number
+  /** onItemClick method for all menu items */
+  onItemClick?: (item: IContextMenuItem) => void
+  children?: React.ReactNode
 }
 
-const BaseComboButton = styled("button")<{ disabled?: ComboButtonProps["disabled"]; isOpen: boolean }>(
-  ({ theme, disabled, isOpen }) => {
-    return {
-      backgroundColor: theme.color.primary,
-      lineHeight: "36px",
-      fontSize: theme.font.size.small,
-      fontFamily: theme.font.family.main,
-      fontWeight: theme.font.weight.medium,
-      color: theme.color.text.white,
-      display: "inline-flex",
-      alignItems: "center",
-      justifyContent: "center",
-      padding: `0 0 0 ${theme.space.element}px`,
-      borderRadius: theme.borderRadius,
-      border: 0,
-      cursor: disabled ? "not-allowed" : "pointer",
-      opacity: disabled ? 0.6 : 1.0,
-      outline: "none",
-      position: "relative",
-      marginRight: theme.space.small,
-      ...(isOpen
-        ? {
-            backgroundColor: theme.color.white,
-            color: theme.color.text.default,
-            borderColor: theme.color.border.select,
-          }
-        : {
-            backgroundColor: theme.color.primary,
-            color: theme.color.text.white,
-            borderColor: theme.color.border.invisible,
-          }),
-      // Apply styles with increased specificity to override defaults
-      ":focus": {
-        ...inputFocus({ theme }),
-      },
-      ...(!disabled
-        ? {
-            ":hover": {
-              backgroundColor: isOpen ? theme.color.white : darken(theme.color.primary, 5),
-            },
-          }
-        : {}),
-    }
-  },
-)
+const BaseComboButton = styled(Button)<{ isOpen: boolean; textColor: string }>(({ isOpen, textColor, theme }) => ({
+  paddingRight: 0,
+  ...(isOpen
+    ? {
+        boxShadow: `0 0 0 1px ${expandColor(theme, textColor)} inset`,
+      }
+    : {}),
+}))
 
-const CaretContainer = styled("div")<{ isOpen: boolean }>(({ isOpen, theme }) => {
-  return {
-    width: "36px",
-    marginLeft: theme.space.content,
-    borderLeft: `1px solid ${isOpen ? theme.color.primary : theme.color.text.white}`,
-  }
-})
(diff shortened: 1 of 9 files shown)


stereobooster:
Dropdown is partially hidden (and view is not scrollable)

image

src/ComboButton/ComboButton.tsx
```diff
@@ -0,0 +1,123 @@
...
+  return (
+    <ContextMenu {...props} iconLocation="right" items={itemsWithCarets}>
+      {isOpen => {
+        return (
+          <BaseComboButton
+            {...props}
+            role="button"
+            aria-label={title}
+            tabIndex={tabIndex}
+            onClick={(ev: React.SyntheticEvent<React.ReactNode>) => {
```
stereobooster: `SyntheticEvent<T = Element` and `target: T;` so param should be one of HTMLElements like `HTMLDivElement`

src/ComboButton/ComboButton.tsx
```diff
@@ -0,0 +1,123 @@
...
+  justify-content: space-between;
+`
+
+const ComboButton: React.SFC<ComboButtonProps> = ({ items, title, tabIndex, ...props }) => {
+  // Dropdown buttons always have right carets for every item. These are
+  const itemsWithCarets = items.map(item => {
+    return {
+      label: (
+        <ItemWithCaret>
+          {typeof item === "string" ? item : item.label}
```
stereobooster: not sure if this is a big deal or not but `isString` from `lodash` is faster then `typeof item === "string"`

ImogenF:
@stereobooster Not sure what you mean about the dropdown being partially hidden, it looks fine to me. How can I recreate that?

src/ComboButton/ComboButton.tsx
```diff
@@ -0,0 +1,123 @@
...
+  disabled?: boolean
+  /** What is the tab index, for accessibility? */
+  tabIndex?: number
+}
+
+const BaseComboButton = styled("button")<{ disabled?: ComboButtonProps["disabled"]; isOpen: boolean }>(
+  ({ theme, disabled, isOpen }) => {
+    return {
+      backgroundColor: theme.color.primary,
+      lineHeight: "36px",
```
stereobooster: why do we need it? Can we use theme for it or use flexbox alignment instead?

stereobooster:
@ImogenF make window height smaller (my screen is 13')

ImogenF:
> @ImogenF make window height smaller (my screen is 13')

Still looks fine to me... 🤔

stereobooster:
Code is ok. Issue with partially hidden content can be addressed separately, because I guess it exists for all components which use context menu. I don't know about visual appearance, because I haven't seen design (I can't open ticket). I guess we can merge

src/ComboButton/ComboButton.tsx
```diff
@@ -0,0 +1,123 @@
...
+  /** Disabled option */
+  disabled?: boolean
+  /** What is the tab index, for accessibility? */
+  tabIndex?: number
+}
+
+const BaseComboButton = styled("button")<{ disabled?: ComboButtonProps["disabled"]; isOpen: boolean }>(
+  ({ theme, disabled, isOpen }) => {
+    return {
+      backgroundColor: theme.color.primary,
```
TejasQ: I suggest we have this customizable via a `color` prop.

src/ContextMenu/ContextMenu.tsx
```diff
@@ -76,8 +76,9 @@ const MenuContainer = styled("div")<{
   boxShadow: theme.shadows.popup,
   width: "100%",
   minWidth: "fit-content",
+  minHeight: condensed ? condensedRowHeight : rowHeight,
   display: "grid",
-  gridTemplateRows: `repeat(${numRows}, ${condensed ? condensedRowHeight : rowHeight}px)`,
+  gridTemplateRows: `repeat(${numRows}, maxcontent)`,
```
TejasQ: Isn't it `max-content`? Is that valid CSS? 🤔

TejasQ:
@stereobooster is right – this does go past my screen too:

image

src/ComboButton/README.md
```diff
@@ -0,0 +1,55 @@
...
+### Usage with complex list contents
+
+In some cases, you might want your `label` to be a little bit more clever than just a string. This example shows a `ContextMenu` with a JSX element as its `label`.
+
+```jsx
+import * as React from "react"
+import { ComboButton, DataSourceIcon } from "@operational/components"
+
+/* Anything can be a label */
+const MyLabelContainer = ({ title, children, style }) => (
```
TejasQ: The README is invalid because `style` is not really used here.

image

```suggestion
const MyLabelContainer = ({ title, children }) => (
```

src/ComboButton/README.md
```diff
@@ -0,0 +1,55 @@
+### Usage
+
+```jsx
+import * as React from "react"
+import { CaretRightIcon, ContextMenu, ComboButton } from "@operational/components"
+const menuItems = ["Project", "User"]
+;<ComboButton title="Create" items={menuItems} />
```
TejasQ: Maybe do something to indicate which option is selected on click?

src/ComboButton/ComboButton.tsx
```diff
@@ -0,0 +1,123 @@
+import * as React from "react"
+
+import { DefaultProps } from "../types"
+import { darken, inputFocus } from "../utils"
+import styled from "../utils/styled"
+import { CaretDownIcon, CaretRightIcon } from "../Icon/Icon"
+import ContextMenu, { ContextMenuProps } from "../ContextMenu/ContextMenu"
+
+export interface ComboButtonProps extends DefaultProps {
```
TejasQ: There's a missing `onClick` prop. What happens if you have a list of options that are just strings as in the first example in the README?

src/ComboButton/README.md
TejasQ: Let's add `onClick` here as well to show that it responds to clicks and keyboard events. This is kind of why e2e tests help – so we don't miss these cases.

TejasQ:
Hmmm. Lea added a new [screen](https://contiamo.atlassian.net/browse/CON-268) today that has a different hover/focus state than what `ContextMenu` gives us. I think it could make more sense to not depend on `ContextMenu`, but instead use `useListbox` and style our own children.

`useListbox` is a little bit lower level in terms of abstraction and should give us what we want. 

For reference, [SidenavItem](https://github.com/contiamo/operational-ui/blob/8803b163372e6d87ad1e44397cd4fd7be8d0ae65/src/SidenavItem/SidenavItem.tsx#L25) already works this way, for exactly the same reasons.

src/ComboButton/README.md
```diff
@@ -0,0 +1,55 @@
+### Usage
+
+```jsx
+import * as React from "react"
+import { CaretRightIcon, ContextMenu, ComboButton } from "@operational/components"
+const menuItems = ["Project", "User"]
```
TejasQ: It would be nice if these were `IContextMenuItems` with an `onClick` instead of just strings to show some real interactivity (and model the behavior from the [screens](https://app.zeplin.io/project/5cc31fad6f17562d6dac9a53/screen/5ce666f82c7c2d1dc81c2947)).

src/ComboButton/ComboButton.tsx
```diff
@@ -0,0 +1,123 @@
...
+  /** Title */
+  title?: string
```
TejasQ: This is non-idiomatic because a more common pattern is

```
<ComboButton>
  children 
</ComboButton>
```

I suggest removing this prop and using `children` instead.

TejasQ:
Can we add (or preferably, update) an example to `Page`'s README with this to show its composability when used as a page action?

ImogenF:
There are 2 open issues: 
1) The dropdown being cut off, which should be dealt with separately.
2) The down/up carets look different than in the spec, because we don't have those carets in the library yet.

src/ComboButton/README.md
ImogenF: Have added `onClick` events to individual items in the 2nd example. I'd like to leave this one with just strings to demonstrate that this works.

fabien0102:
> The down/up carets look different than in the spec, because we don't have those carets in the library yet.

You can rebase on master, I just merged the @nathanedavis1 PR with new icons 😉 cf #1050

src/ComboButton/ComboButton.tsx
TejasQ: There's still a missing `onClick` prop, though there's one in the README. As a result, the examples are red:

image

src/ComboButton/README.md
```diff
@@ -0,0 +1,78 @@
...
+import { ComboButton } from "@operational/components"
+
+const menuItems = ["Project", "User"]
+;<ComboButton
+  color="error"
+  items={menuItems}
+  onClick={() => {
+    console.log("Button clicked")
+  }}
+  onItemClick={() => {
```
TejasQ: ⚠️  This is a different pattern than `ContextMenu` employs: `IContextMenuItem`s also support on onClick prop. When a user clicks on an item, it's `onClick` handler is fired. I'm not saying this is bad, but it's a divergence of an existing pattern that might be confusing to some users.

ImogenF: @TejasQ The problem is that here, we're dealing with both a `Button` (which can have its own`onClick` event) and a `ContextMenu`, whose `onClick` events relate to clicking on individual items. So I had to distinguish between these two options in some way, hence having `onClick` for the button itself (since the component is first and foremost a button) and `onItemClick` which gets passed to the `ContextMenu` as `onClick`.

src/ComboButton/ComboButton.tsx
ImogenF: Good spot. My editor is not catching this error, for some reason...

stereobooster:
I have a bit different idea of how combo-button works

Screenshot 2019-06-21 at 11 10 55

left part is a button, and behaves like a button e.g. click/enter/space triggers main action

right part is dropdown which switches content of left button 
 - so I don't see why we would want to expose on item click callback, it suppose to switch which item is selected
 - click / down arrow/ up arrow, space show dropdown, sure can select any item, based on this content of the left part changes

`¯\_(ツ)_/¯`

ImogenF:
> I have a bit different idea of how combo-button works

@stereobooster normally, I would agree with you, but for the two use cases we've got ("Add data source", and "Create") I don't think that makes sense. I think the idea here is that you click to add a data source or create something new, and the button opens a dropdown to ask for a bit more information about what exactly it is you want to create or add...

src/ComboButton/README.md
TejasQ: 🤔 Conceptually though, this component _feels like_ a `ContextMenu`, so what I expect (I'm using it already) is that the button opens/closes and passes any other `onClick` behavior through to `ContextMenu`.

I would then get more fine-grained control if needed by giving my `items:` each their own `onClick` prop.

Here's my PR that uses this for reference: https://github.com/contiamo/contiamo-ui/pull/176/files#diff-12393d6759ed8e1bcb22f3711927014fR39

It's interesting: I'm not actually doing anything `onClick` of the button, but only on the items. Perhaps we make the `onClick` optional.

TejasQ:
> I think the idea here is that you click to add a data source or create something new, and the button opens a dropdown to ask for a bit more information about what exactly it is you want to create or add...

Yes, like a multi-stage [FAB](https://material.io/design/components/buttons-floating-action-button.html#types-of-transitions).

But I also see @stereobooster's point. Perhaps we can iterate.

src/ComboButton/ComboButton.tsx
```diff
@@ -0,0 +1,90 @@
...
+const CaretContainer = styled("div")<{ isOpen: boolean; color: string }>(({ isOpen, color, theme }) => ({
+  width: "36px",
+  marginLeft: theme.space.content,
+  borderLeft: `1px solid ${isOpen ? expandColor(theme, color) : setAlpha(0.5)(theme.color.text.white)}`,
+}))
+
+const ItemWithChevron = styled("div")`
+  width: 100%;
+  display: flex;
+  align-items: baseline;
```
TejasQ: Why not `center`? I think this causes https://github.com/contiamo/operational-ui/issues/1074.

src/ComboButton/README.md
ImogenF: Both `onClick` and `onItemClick` are optional at the moment (since click events can also be specified for each item individually rather than through the global `onItemClick` prop). 

I agree that for the use cases we currently have, there is no need for `onClick` as it currently stands. But it may be that for a future use-case, the user would want some other behaviour when the button is clicked - changing the page styling behind the dropdown, setting a flag, or whatever it may be.

TejasQ: Yeah, let's resolve it for now and iterate as we go.

ImogenF:
> But I also see @stereobooster's point. Perhaps we can iterate.

My feeling is that we're basically talking about 2 distinct components here: the one I've implemented, and the one @stereobooster is talking about. Maybe we rename this component to be `DropdownButton` and if the need arises in the future, we can have a `SwitchButton` or something where the dropdown caret is a separate clickable area and switches the action of the main part of the button.

TejasQ:
Agree.

ImogenF:
When did the `ContextMenu` width break?

TejasQ:
Not sure, but there's a fix: https://github.com/contiamo/operational-ui/pull/1072

ImogenF:
@TejasQ great, will rebase and check everything looks good in this branch

src/DropdownButton/DropdownButton.tsx
```diff
@@ -0,0 +1,90 @@
...
+import styled from "../utils/styled"
+import { CaretDownIcon, CaretUpIcon, ChevronRightIcon } from "../Icon/Icon"
+import ContextMenu, { ContextMenuProps } from "../ContextMenu/ContextMenu"
+import Button from "../Button/Button"
+import { IContextMenuItem } from "../ContextMenu/ContextMenu.Item"
+import { setAlpha } from "../utils"
+
+export interface DropdownButtonProps extends DefaultProps {
+  /** Button color theme (hex or named color from `theme.color`) */
+  color: string
```
fabien0102: Why `color` is a mandatory props? (it's actually not used in the example)

image

src/DropdownButton/DropdownButton.tsx
```diff
@@ -0,0 +1,92 @@
...
+  paddingRight: 0,
+  ...(isOpen
+    ? {
+        boxShadow: `0 0 0 1px ${expandColor(theme, textColor)} inset`,
+      }
+    : {}),
+}))
+
+const CaretContainer = styled("div")<{ isOpen: boolean; color: string }>(({ isOpen, color, theme }) => ({
+  width: "36px",
```
TejasQ:
```suggestion
  width: 36,
```

src/DropdownButton/DropdownButton.tsx
```diff
@@ -0,0 +1,92 @@
...
+  onClick?: (e?: React.SyntheticEvent<React.ReactNode>) => void
+  /** onItemClick method for all menu items */
+  onItemClick?: (item: IContextMenuItem) => void
+  children?: React.ReactNode
+}
+
+const BaseDropdownButton = styled(Button)<{ isOpen: boolean; textColor: string }>(({ isOpen, textColor, theme }) => ({
+  paddingRight: 0,
+  ...(isOpen
+    ? {
```
TejasQ: You could also use `&&` here and remove the ternary expression. More nits, not a big deal.

PR #1045, opened 2019-06-18
**Feature** UI-79 Expose onBlur callback for the custom Input option

mpotomin:
# Summary
As a developer, I'd like to be able to handle the case when User removes the focus from the Input within the Select component

Let's say we would like to prevent user leaving the custom input value empty, but still would like to allow user clean up the input before starting entering something new. We can use onInputBlur callback to implement such behaviour.

Use case: Select with custom input for Delimiter in the DataStore UI. There should be a way to handle the empty string as a custom value to re-set the previous/default value back into the Select.

# Related issue
https://contiamo.atlassian.net/browse/UI-79

# To be tested

Me

Tester 1

Tester 2


TejasQ:
Maybe we can call it `onCustomInputBlur`? Or even have it be a part of `customInput: { onBlur }`?

src/Select/README.md
```diff
@@ -342,6 +342,65 @@ ${JSON.stringify(lastChanged, null, 2)}
...
+    <Form>
+      <Select
+        data-cy="custom-select"
+        value={value}
+        options={options}
+        label="Custom Input"
+        placeholder="Choose an option"
+        customOption={{ label: "Custom...", value: customOptionValue }}
+        onChange={(newValue, lastChanged) => {
+          setValue(newValue)
```
TejasQ:
```suggestion
          setValue(String(newValue))
```

src/Select/README.md
```diff
@@ -342,6 +342,65 @@ ${JSON.stringify(lastChanged, null, 2)}
...
+          if (lastChanged && lastChanged.label === "Custom...") {
+            setCustomOptionValue(newValue)
```
TejasQ:
```suggestion
            setCustomOptionValue(String(newValue))
```

TejasQ:
@mpotomin please run the e2e tests manually to make sure they didn't break. They rely on the order of examples in the README and the order has changed so I'm expecting breakages there.

For the cleanest approach, add your new example to the bottom of the README.

mpotomin:
@TejasQ improved the API for the settings of the custom Input within the Select, also moved the test and the readme for the new usage case to the very bottom. These 3 e2e are failing the same as they do on master currently. 

Screenshot 2019-06-18 at 19 05 28

Will if I can find the way to fix them with a delay between actions, maybe on Friday, otherwise next week.

PR #1049, opened 2019-06-18
**Feature:** Remove scroll arrows from Tabs

TejasQ:
This feature adds functionality to use Tabs without a scrollbar in case we want to use it in a sidebar or something. 😋

### DIFF
--- a/src/Tabs/Tabs.tsx
+++ b/src/Tabs/Tabs.tsx
@@ -1,12 +1,20 @@
 import * as React from "react"
-import { SectionHeader } from "../Internals/SectionHeader"
 import { DefaultProps } from "../types"
-import styled from "../utils/styled"
 import { useUniqueId } from "../useUniqueId"
 import { NoIcon, PlusIcon, ChevronLeftIcon, ChevronRightIcon } from "../Icon/Icon"
 import { ScrollButton } from "./ScrollButton"
-
-const buttonWidth = 55
+import {
+  Container,
+  ScrollButtons,
+  TabContainer,
+  TabHeader,
+  TabIcon,
+  TabList,
+  TabPanel,
+  TabScroll,
+  TitleIconWrapper,
+  TitleWrapper,
+} from "./Tabs.styled"
 
 export interface Tab {
   title: string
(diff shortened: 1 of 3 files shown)


src/Tabs/Tabs.tsx
```diff
@@ -297,34 +188,36 @@ const Tabs: React.FC<TabsProps> = ({ tabs, active, onClose, onActivate, onInsert
...
+          >
+            <ChevronRightIcon size={14} />
+          </TabHeader>
+        </ScrollButtons>
+      )}
       <TabContainer>
         {tabs.map((_, i) => (
           <TabPanel hidden={i !== active} id={`TabPanel-${uid}-${i}`} aria-labelledby={`TabHeader-${uid}-${i}`} key={i}>
-            {i === active && children}
+            <TabContent>{i === active && children}</TabContent>
```
stereobooster: why separate element? You can put padding on TabPanel itself. As well we have 2 use-cases where we don't want any padding, so we need to write `margin: -20px` to undo padding. Can we remove padding? cc @fabien0102

src/Tabs/Tabs.tsx
```diff
@@ -17,146 +26,26 @@ export interface TabsProps extends DefaultProps {
   tabs: Tab[]
   active: number
   onActivate: (tabIndex: number) => void
+  noScroll?: boolean
```
stereobooster: I prefer not have negatives in name. Can we change it `scroll?: boolean` and provide `true` as default value.

So when you need to disable scroll you will do `<Tabs scroll={false}>`

src/Tabs/Tabs.tsx
```diff
@@ -297,34 +188,36 @@ const Tabs: React.FC<TabsProps> = ({ tabs, active, onClose, onActivate, onInsert
...
-          tabIndex={-1}
-          first={false}
-          aria-selected={false}
-          condensed={true}
-          onClick={scrollRight}
-        >
-          <ChevronRightIcon size={14} />
-        </TabHeader>
-      </ScrollButtons>
+      {noScroll !== true && (
```
fabien0102: `noScroll !== true` my brain just explode to try resolving this simple condition, to much computation 💥

src/Tabs/Tabs.tsx
fabien0102: I don't have all the context about this component and usecase, let's try to remove this padding, if all example are following, deal! 😉 

After if it works like this, it's also ok for me

stereobooster: Screenshot 2019-06-19 at 14 47 28
no padding here

TejasQ:
Closing in favor of #1053

PR #1051, opened 2019-06-19
**Feature:** UI-43 Id input style and a custom status icon

mpotomin:
# ToDo:
- [x] Add tests
- [x] Swap the id Icon to use the new one 

# Summary
1. Adds Id input style with an icon
1. Ensures that the icon still works with the clear callback and the button layout
1. Adds the possibility to display user-defined status icon 
1. Ensures the status icon still with fullWidth option

Screenshot 2019-06-19 at 13 45 48

# Related issue
https://contiamo.atlassian.net/browse/UI-43

# To be tested

Me

Tester 1

- [ ] No regressions can be found

Tester 2
- [ ] No regressions can be found

### DIFF
--- a/src/Input/Input.Field.tsx
+++ b/src/Input/Input.Field.tsx
@@ -5,7 +5,7 @@ import styled from "../utils/styled"
 import { InputProps } from "./Input"
 import InputButton from "./Input.Button"
 import { height } from "./Input.constants"
-import { NoIcon } from "../Icon/Icon"
+import { NoIcon, ServiceAccountIcon } from "../Icon/Icon"
 
 const width = 360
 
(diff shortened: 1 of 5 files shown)


src/Input/Input.Field.tsx
```diff
@@ -5,20 +5,33 @@ import styled from "../utils/styled"
 import { InputProps } from "./Input"
 import InputButton from "./Input.Button"
 import { height } from "./Input.constants"
-import { NoIcon } from "../Icon/Icon"
+import { NoIcon, ServiceAccountIcon } from "../Icon/Icon"
 
 const width = 360
+const iconBoxSize = height // Fit icons into the square with the side of the input height
```
TejasQ: We have a file called `Input.constants`. Maybe this and the line above it should be exported from there?

src/Input/Input.tsx
```diff
@@ -53,6 +53,10 @@ export interface BaseProps extends DefaultProps, DefaultInputProps, Omit<React.I
   errorComponent?: React.FC<{ errorMessage: string }>
   /** Is the input read-only? */
   readOnly?: boolean
+  /** Should this input get Id input style */
+  idStyle?: boolean
```
TejasQ: I think this naming leaks implementation details and as a result isn't as declarative. We don't really care about the style imho. I'd propose calling it `isId` or `isUniqueId` or something similar and hide away the `style` part of the name.

mpotomin: Agree, will help to make an API cleaner

mpotomin:
Addressed the PR review comments and swapped the [ID] Icon to use the new one from Nathan.

PR #1059, opened 2019-06-20
**Fix:** Deal with datatable without rows

fabien0102:
# Summary

image

Empty rows -> BOOOOM!

# Related issue

# To be tested

Me

Tester 1

Tester 2

### DIFF
--- a/src/DataTable/DataTable.tsx
+++ b/src/DataTable/DataTable.tsx
@@ -96,7 +96,7 @@ export function DataTable<Columns extends any[][], Rows extends any[][]>({
     [columns, rows, rowHeight],
   )
 
-  const numCells = React.useMemo(() => rows[0].length, [rows])
+  const numCells = React.useMemo(() => (rows[0] ? rows[0].length : 0), [rows])
   const VirtualRow: React.FC<ListChildComponentProps> = React.useMemo(
     () =>
       React.memo(({ style, index }) => (


fabien0102:
On prod
image

```jsx
import { DataTable, DataTableSelect, DataTableInput, Checkbox } from "@operational/components"

;<DataTable
  columns={[
    [
      "I am so long I am the longest the longest of the long LOL look how long I am my mom said I would never be long but I really am the longest KOBE BRYANT AINT GOT NOTHING ON ME HOMIE",
    ],
    ["Name"],
    ["Diet"],
  ]}
  rows={[]}
/>
```

src/DataTable/DataTable.tsx
```diff
@@ -96,7 +96,7 @@ export function DataTable<Columns extends any[][], Rows extends any[][]>({
     [columns, rows, rowHeight],
   )
 
-  const numCells = React.useMemo(() => rows[0].length, [rows])
+  const numCells = React.useMemo(() => (rows[0] ? rows[0].length : 0), [rows])
```
stereobooster: but how it gonna be displayed? Can we provide header separately?

fabien0102: I don't know I will try on the PR preview (my dev server is down ^^)

fabien0102: image

Header only, good enough for now 😁

PR #1047, opened 2019-06-18
**Fix:** Handling of clicking on ContextMenuItems

TejasQ:
There's a bug when trying to click on a ContextMenuItem. It doesn't work. This PR fixes it.

## To test:
1. Open [PR Preview](https://deploy-preview-1047--operational-ui.netlify.com/#/Components/Table)
1. Diff with master
1. Make sure you're on Table
1. Go to **With row actions**

- [ ] Click on links in action menu, should work on this version, not on master

Issue #1006, opened 2019-06-04
Consolidate styles of focus state


stereobooster:
We implemented 2 new components (Accordion, Tabs) with slightly updated style. And there is new PR https://github.com/contiamo/operational-ui/pull/1005. Those PR add 2 new styles for focus, as well we have style of focus from previous code.

We need to use one consistent style for focus state.

<img width="382" alt="Screenshot 2019-06-05 at 17 43 20" src="https://user-images.githubusercontent.com/179534/58970586-03dd1c80-87ba-11e9-83be-c1d4fa06dd85.png">
<img width="124" alt="Screenshot 2019-06-05 at 17 43 41" src="https://user-images.githubusercontent.com/179534/58970589-0475b300-87ba-11e9-8134-921edb8121c7.png">
<img width="203" alt="Screenshot 2019-06-05 at 17 43 54" src="https://user-images.githubusercontent.com/179534/58970590-0475b300-87ba-11e9-8c92-8b84cd573f3d.png">
<img width="342" alt="Screenshot 2019-06-05 at 17 44 59" src="https://user-images.githubusercontent.com/179534/58970591-0475b300-87ba-11e9-80d8-4280f28aacea.png">

After #1005
<img width="225" alt="Screenshot 2019-06-05 at 17 46 35" src="https://user-images.githubusercontent.com/179534/58970593-050e4980-87ba-11e9-8ab9-507930dd23d9.png">
<img width="202" alt="Screenshot 2019-06-05 at 17 46 40" src="https://user-images.githubusercontent.com/179534/58970594-050e4980-87ba-11e9-8630-dd701fb14b05.png">
<img width="176" alt="Screenshot 2019-06-05 at 17 46 49" src="https://user-images.githubusercontent.com/179534/58970595-050e4980-87ba-11e9-8e3e-c15da14e5682.png">

stereobooster:
As well we need to show focus only for keyboard users, but not for mouse users (like described here https://css-tricks.com/keyboard-only-focus-styles/)

TejasQ:
This is a great issue.

michaeldera:
Is this still available to work on?

TejasQ:
There isn't really an idea of what to implement here yet – it's more a design question of which style we settle on. @kemal-contiamo or @micha-f any ideas? 

When we decide which one wins, @michaeldera, would you like to implement it?

michaeldera:
Yes please. Would love to tackle it.

[...]

PR #1060, opened 2019-06-20
**Fix:** Select menu not closing

### LINKED ISSUE #1056: Select/Context menu should close when  options is selected
[Select](http://operational-ui.netlify.com/#/Components/Select) with a single option should close when the selection is made. 

Select bug

mpotomin:
# Summary
Fixes #1056 – the regression, introduced  by https://github.com/contiamo/operational-ui/pull/1047
The context menu should close now as soon as an item within the menu is selected. It should still stay opened if `keepOpenOnItemClick` was provided

# Related issue
https://github.com/contiamo/operational-ui/issues/1056

# To be tested

Me

Tester 1
- [ ] Context menu works as expected in all README.md cases (including multi-select cases)
- [ ] Select works as expected in all README.md cases (including multi-select functionalitfy)
- [ ] Action menu works as expected in all README.md cases

### DIFF
--- a/src/ContextMenu/ContextMenu.tsx
+++ b/src/ContextMenu/ContextMenu.tsx
@@ -163,7 +163,7 @@ const ContextMenu: React.FC<ContextMenuProps> = ({
       {isOpen && (
         <InvisibleOverlay
           onClick={e => {
-            e.stopPropagation() //clicking on an item should not trigger the parent's onClick
+            e.stopPropagation()
             setIsOpen && setIsOpen(false)
           }}
         />
--- a/src/ContextMenu/ContextMenu.tsx
+++ b/src/ContextMenu/ContextMenu.tsx
@@ -163,7 +163,7 @@ const ContextMenu: React.FC<ContextMenuProps> = ({
       {isOpen && (
         <InvisibleOverlay
           onClick={e => {
-            e.stopPropagation()
+            e.stopPropagation() //clicking on an item should not trigger the parent's onClick
             setIsOpen && setIsOpen(false)
           }}
         />
@@ -214,7 +214,7 @@ const ContextMenu: React.FC<ContextMenuProps> = ({
               width={width || "100%"}
               item={item}
               onClick={e => {
-                e.stopPropagation()
+                e.stopPropagation() //clicking on an item should not trigger the parent's onClick
                 if (!keepOpenOnItemClick && setIsOpen) {
                   setIsOpen(false)
                 }
--- a/src/ContextMenu/ContextMenu.tsx
+++ b/src/ContextMenu/ContextMenu.tsx
@@ -217,10 +217,16 @@ const ContextMenu: React.FC<ContextMenuProps> = ({
                 e.stopPropagation()
                 if (!isString(item) && item.onClick) {
                   item.onClick(makeItem(item))
+                  if (!keepOpenOnItemClick && setIsOpen) {
+                    setIsOpen(false)
+                  }
                   return
                 }
                 if (onClick) {
                   onClick(makeItem(item))
+                  if (!keepOpenOnItemClick && setIsOpen) {
+                    setIsOpen(false)
+                  }
                 }
               }}
               {...(getChildProps ? getChildProps(index) : {})}
--- a/src/ContextMenu/ContextMenu.tsx
+++ b/src/ContextMenu/ContextMenu.tsx
@@ -215,18 +215,15 @@ const ContextMenu: React.FC<ContextMenuProps> = ({
               item={item}
               onClick={e => {
                 e.stopPropagation()
+                if (!keepOpenOnItemClick && setIsOpen) {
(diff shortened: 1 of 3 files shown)


src/ContextMenu/ContextMenu.tsx
```diff
@@ -217,10 +217,16 @@ const ContextMenu: React.FC<ContextMenuProps> = ({
...
                 if (!isString(item) && item.onClick) {
                   item.onClick(makeItem(item))
+                  if (!keepOpenOnItemClick && setIsOpen) {
+                    setIsOpen(false)
+                  }
                   return
                 }
                 if (onClick) {
                   onClick(makeItem(item))
+                  if (!keepOpenOnItemClick && setIsOpen) {
```
mpotomin: @TejasQ If i'm correct, this should close the menu since `e.stopPropagation()` was introduced at the top of this handler. 

Also would like to add a comment above the comment above `e.stopPropagation()` to explain why it is there. but i don't have enough context on it now.

src/ContextMenu/README.md
```diff
@@ -123,7 +123,7 @@ import * as React from "react"
 import { ContextMenu, ContextMenuProps } from "@operational/components"
 
 const menuItems = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16].map(item => `Menu ${item}`)
-;<ContextMenu condensed items={menuItems} onClick={() => alert("clicked")}>
+;<ContextMenu keepOpenOnItemClick condensed items={menuItems} onClick={() => alert("clicked")}>
```
mpotomin: @TejasQ I think `keepOpenOnItemClick` was is missing in the original doc and this example in the readme, otherwise i do not understand how this part is supposed to work, since the, unlike Select component, the is no `value` prop here to check if it is an array or not to figure if it should be multi-select, and all the examples receive an array of items. If so, `<div>Many options may be selected here</div>` is too ambiguous and should  something like "It will not close when the item is clicked"

mpotomin:
@TejasQ This update fixes the behaviour for Select for both single-select and multi-select examples. I am not 100% sure that it is the case for the examples for the ActionMenu and ContextMenu, since the description in them is ambiguous. Please help test and improve them.

fabien0102:
image

The focus is missing on focus (when you select the component with tab)

fabien0102:
image

The state is not following when you remove the last caract

src/ContextMenu/README.md
stereobooster: I feel like this is escape hatch - we complicate API to make it work in all cases. Maybe this component tries to do to much things and we need to have to components or reconsider API/implementation?

src/ContextMenu/ContextMenu.tsx
TejasQ: `e.stopPropagation` was introduced so that clicking on an item (which can have its own `onClick`) does not also trigger the parent's onClick, which can, in some cases, do more than just opening the menu.

I don't understand why you're doing this twice. You can check for this condition above both cases and handle open/close once?

src/ContextMenu/README.md
TejasQ: This example isn't really needed here, but on Select. I'm not sure what you're trying to do here?

TejasQ:
Per @fabien0102's comments:

> The focus is missing on focus (when you select the component with tab)

There's an open issue for this: https://github.com/contiamo/operational-ui/issues/1006

> The state is not following when you remove the last caract

Let's open another issue for this and fix it separately. Let's not block this PR because it fixes a different problem.

src/ContextMenu/README.md
mpotomin: The example is saying "Many options may be selected here", which was not true - the menu was closing right after the first click on the item in it. Therefore I aded `keepOpenOnItemClick`, so that it at leas stays on the screen open after the first click so that any other item i selectable. 

Since you say, this example is not really needed here, lets remove it completely so it won't confuse someone else.

mpotomin: Ah, now I see, "Many options may be selected here" means the list can be very long with *many* items in it. It does not mean that there can be multiple items selected when the context menu is opened. Updated the description.

src/ContextMenu/ContextMenu.tsx
```diff
@@ -163,7 +163,7 @@ const ContextMenu: React.FC<ContextMenuProps> = ({
       {isOpen && (
         <InvisibleOverlay
           onClick={e => {
-            e.stopPropagation()
+            e.stopPropagation() //clicking on an item should not trigger the parent's onClick
```
TejasQ: That's not why this is here. In the context of this component, it has no parent – it is siblings with the menu.

It should read: clicking on this should trap clicks and not propagate them through parents – but this is kind of obvious with what `stopPropagation` does and I'm not sure it's needed here.

PR #1062, opened 2019-06-21
**Feature:** Add placeholder icon to Input.

ImogenF:
# Summary

Added a placeholder icon to the `Input` component that is displayed on the right side of the input field when the placeholder is displayed. Specific use-case: search bars. 

image

I also made the `X` button for clearing the input slightly smaller to match the size of the search icon in the specs.

https://contiamo.atlassian.net/secure/RapidBoard.jspa?rapidView=43&projectKey=UI&modal=detail&selectedIssue=UI-84

# To be tested

Me

Tester 1

Tester 2

### DIFF
--- a/src/Input/README.md
+++ b/src/Input/README.md
@@ -228,3 +228,25 @@ const [value, setValue] = React.useState("MyTable_01")
   />
 </Form>
 ```
+
+### With search icon in placeholder
+
+```jsx
+import * as React from "react"
+import { Input, SearchIcon } from "@operational/components"
+
+const SearchInput = () => {
+  const [value, setValue] = React.useState("")
+  return (
+    <Input
+      value={value}
+      onChange={setValue}
+      placeholder="Search measures & dimensions"
+      placeholderIcon={<SearchIcon size={10} />}
+      clear={() => setValue("")}
+    />
+  )
+}
+
+;<SearchInput />
+```
--- a/src/Input/Input.Field.tsx
+++ b/src/Input/Input.Field.tsx
@@ -175,6 +175,7 @@ const InputField: React.FC<InputProps> = ({
   errorComponent: ErrorComponent,
   isUniqueId,
   statusIcon,
+  placeholderIcon,
   ...props
 }) => {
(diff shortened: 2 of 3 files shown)


stereobooster:
We already have "placeholder icon" in Id fields. Do we care about this? Should we unify this? 
 
Screenshot 2019-06-21 at 10 32 05

src/Input/README.md
```diff
@@ -228,3 +228,25 @@ const [value, setValue] = React.useState("MyTable_01")
...
+import { Input, SearchIcon } from "@operational/components"
+
+const SearchInput = () => {
+  const [value, setValue] = React.useState("")
+  return (
+    <Input
+      value={value}
+      onChange={setValue}
+      placeholder="Search measures & dimensions"
+      placeholderIcon={<SearchIcon size={10} />}
```
stereobooster: not concern of this PR, but we have two distinctive patterns for icons when we pass React element ("instance") and when we pass react component ("class"). This will confuse people

ImogenF:
> We already have "placeholder icon" in Id fields. Do we care about this? Should we unify this?

The `isUniqueId` prop does more than just add the icon - it also changes the font and font weight. So I think it's a special case.

src/Input/README.md
fabien0102: True, the size should be part of the internal logic

ImogenF: In this case I decided to stick with the pattern already used here for the `statusIcon` prop. So it's at least consistent within this component.

fabien0102: I think we should move to the `Breadcrumb` API:

```tsx
export interface BreadcrumbProps extends DefaultProps {
  /** Navigation property à la react-router <Link/> */
  to: string
  onClick?: (ev?: React.SyntheticEvent<React.ReactNode>) => void
  children?: React.ReactNode
  icon?: IconComponentType
  /** Optionally moves icon from default position on right, to before the text, including appropriate padding */
  iconLeft?: boolean
}
```

You can use `IconComponentType` 😉 

The question is what flexibility we want to provide here, and I'm more to fix the color and the size in this case to avoid non-consistency

PR #1065, opened 2019-06-23
**Fix:** Sanitize user input on search filter

### LINKED ISSUE #928: [Bug] Combobox uses unescaped user input for Regexp
```ts
<OptionsList>
  {options
    .filter(option => (option.label || String(option.value)).match(RegExp(search, "i")))
```

Which will throw errors on some cases

screenshot 2019-02-20 at 12 03 20

solution is to escape regexp

```ts
const searchRegex = RegExp(search.replace(/[.*+?^${}()|[\]\\]/g, '\\$&'), "i")
```

TejasQ:
Fixes #928 by escaping RegEx on user input.

### DIFF
--- a/src/Select/Select.util.ts
+++ b/src/Select/Select.util.ts
@@ -10,7 +10,7 @@ export const truncateList = (maxLength?: number) => (options: SelectProps["optio
   maxLength ? options.slice(0, maxLength) : options
 
 export const filterList = (filter: string) => (options: SelectProps["options"]) =>
-  options.filter(option => String(option.label).match(new RegExp(filter, "ig")))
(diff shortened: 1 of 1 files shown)


src/Select/Select.util.ts
```diff
@@ -10,7 +10,7 @@ export const truncateList = (maxLength?: number) => (options: SelectProps["optio
   maxLength ? options.slice(0, maxLength) : options
 
 export const filterList = (filter: string) => (options: SelectProps["options"]) =>
-  options.filter(option => String(option.label).match(new RegExp(filter, "ig")))
+  options.filter(option => String(option.label).match(new RegExp(filter.replace(/[.*+?^${}()|[\]\\]/g, "\\$&"), "ig")))
```
stereobooster: - you create Regexp in the loop
- extract escapeRegexp function, please

PR #1067, opened 2019-06-23
**Fix:** Show items as disabled if no onClick

### LINKED ISSUE #1061: ActionMenu - Disabled items in the list look same as active
http://localhost:6060/#!/ActionMenu

1. In the example for the ActionMenu the disabled item in the list (Destroy) looks same as active, even though it can not be activated.
2. The README says `No hover behaviour is enabled for items without an `onClick` method.` which is not the case for the actual example

Screenshot 2019-06-20 at 17 53 43

TejasQ:
This PR fixes #1061 by clearly indicating disabled states on context menu items.

### DIFF
--- a/src/ContextMenu/ContextMenu.Item.tsx
+++ b/src/ContextMenu/ContextMenu.Item.tsx
@@ -1,4 +1,6 @@
 import * as React from "react"
+import isString from "lodash/isString"
+
 import { lighten } from "../utils"
 import { OperationalStyleConstants } from "../utils/constants"
 import styled from "../utils/styled"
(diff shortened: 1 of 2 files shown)


src/ContextMenu/ContextMenu.Item.tsx
```diff
@@ -28,8 +31,9 @@ export interface IContextMenuItem<TValue = any> {
   isActive?: boolean
 }
 
-const Container = styled("div")<Props>(({ align, theme, isActive, condensed, width, item }) => {
+const Container = styled("div")<Props>(({ disabled, align, theme, isActive, condensed, width, item }) => {
   const activeShadow = `0 0 0 1px ${theme.color.primary} inset`
+  const doWeHaveDescription = !isString(item) && Boolean(item.description)
```
fabien0102: I like the style, but `hasDescription` should be enough here 😅

PR #1068, opened 2019-06-23
**Fix:** Make Table actions accessible

### LINKED ISSUE #971: Make Table actions accessible
> ⚠️ This was originally posted by @mpotomin in an internal Contiamo channel.

## Motivation

Currently it is not possible to navigate the Table from the keyboard. Also, nothing gives the idea, that there are row actions attached to each item in the Table (gif attached). When triggered, the row actions menu disappears when mouse roll out from the window and there is no way to keep it on the screen (since there is no :focus handling).

FE-44_Table_actions_improvement-f22b18d3-ed76-4cc1-8b8f-86b66c37b8a8

## Suggestion:

1. Tab-cycling through the page should include the table to allow triggering the row actions from the keyboard.
2. Since we hide the context menu (should we, though?) every time the mouse pointer rolls out from the page, the style of the row should be modified so that the action menu stays on the screen and can be navigated if the Table row has the focus (focus would force the behaviour, that the table row currently shows on :hover)

https://github.com/contiamo/operational-ui/blob/39461d66c2bb861afd563b78431faa01877c7baa/src/Table/Table.tsx#L124

Possibly needs smth like & :focus here

TejasQ:
This PR fixes #971. All table rows with an `onRowClick` can now be tabbed to and highlight their respective `ActionMenu`.

### DIFF
--- a/src/ContextMenu/ContextMenu.tsx
+++ b/src/ContextMenu/ContextMenu.tsx
@@ -184,9 +184,7 @@ const ContextMenu: React.FC<ContextMenuProps> = ({
         onKeyDown={e => {
           switch (e.key) {
             case "Enter":
-              if (keepOpenOnItemClick) {
-                e.stopPropagation()
-              }
+              e.stopPropagation()
               handleSelect()
               break
           }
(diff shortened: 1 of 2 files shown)


mpotomin:
Having trouble navigating in the menu in this example:
Screenshot 2019-06-24 at 10 16 18

Maybe I am using my keyboard wrong? 🤔

TejasQ:
So what I expect in that state is:
- Enter triggers `onRowClick` handler
- Tab, then down arrow opens the menu

Does it behave differently?

PR #1083, opened 2019-06-24
**Fix:** Consolidate styles of focus state

michaeldera:
# Summary

The objective was to consolidate the focus state on elements so they are consistent throughout.
Input Focus State.
image

Action Menus Focus State
image

Tab Focus State.
image

# Related issue

#1006  Consolidate styles of focus states.

# To be tested
Visual. If the elements are appearing as they should.

Me
There are React errors on my browser console but nothing that was not there after the initial fork. 

Tester 1

Tester 2


TejasQ:
Wow, thanks for your great work, @michaeldera!

Here's some issues I found:
- [ActionMenu](https://deploy-preview-1083--operational-ui.netlify.com/#/Components/ActionMenu) has no real focus style, though it does have a hover. Let's add an identical focus style here. 
image
- [Breadcrumb](https://deploy-preview-1083--operational-ui.netlify.com/#/Components/Breadcrumb)s have a native focus style instead of our blue outline.
image
- [Button](https://deploy-preview-1083--operational-ui.netlify.com/#/Components/Button)s have no focus style at all.
  - Similar for buttons in [ButtonGroup](https://deploy-preview-1083--operational-ui.netlify.com/#/Components/ButtonGroup)
  - Buttons in [Paginator](https://deploy-preview-1083--operational-ui.netlify.com/#/Components/Paginator) have no focus as well.
- [Checkbox](https://deploy-preview-1083--operational-ui.netlify.com/#/Components/Checkbox)es have no focus styles either.
- [Chip](https://deploy-preview-1083--operational-ui.netlify.com/#/Components/Chip)s have an odd focus style – only the divider turns blue.
- [DataTable](https://deploy-preview-1083--operational-ui.netlify.com/#/Components/DataTable) selects don't have any focus style at all.
- [Select](https://deploy-preview-1083--operational-ui.netlify.com/#/Components/Select) has no focus style.
- [SimpleLink](https://deploy-preview-1083--operational-ui.netlify.com/#/Components/SimpleLink) has no focus style.
- [Stepper](https://deploy-preview-1083--operational-ui.netlify.com/#/Components/Stepper) has no focus style.
- [Switch](https://deploy-preview-1083--operational-ui.netlify.com/#/Components/Switch) is completely inaccessible.
- [Table](https://deploy-preview-1083--operational-ui.netlify.com/#/Components/Table)'s actions have no blue outline on focus.
image 
- [Textarea](https://deploy-preview-1083--operational-ui.netlify.com/#/Components/Textarea) has a completely different focus style than the other components.
image
- [Toggle](https://deploy-preview-1083--operational-ui.netlify.com/#/Components/Toggle) has no focus styles.
- [TopbarButton](https://deploy-preview-1083--operational-ui.netlify.com/#/Components/Topbar) and other Topbar components have completely different focus styles:
image

While reviewing this pull request, it seems you've correctly implemented a consolidated style for _hover_, but not _focus_ (which the issue is about). Is the distinction clear to you? Is there something I (we) can help with clarifying/explaining? **We are here to help you** so please feel free to ask any questions you may have. 💞

michaeldera:
I understand the distinction now. I will work on it later today. I think I will be able to get by. Will let you know if I comes across any hurdles.

michaeldera:
Wanted to find out if it okay if I do two different PRs for the issue @TejasQ? I can list does all the components I am satisfied with then you can review them then I proceed to work on the others.

src/Breadcrumb/Breadcrumb.tsx
```diff
@@ -20,6 +20,12 @@ const Container = styled("a")(({ theme }) => ({
   label: "breadcrumb",
   "&:hover": {
     color: darken(theme.color.primary, 5),
+    boxShadow: `0 0 0 1px ${theme.color.primary}`,
+  },
+  "&:focus": {
```
TejasQ: Hmmm, we seem to be reusing this style quite a bit. We also have this [mixin](https://github.com/contiamo/operational-ui/blob/94f66cca0431a0e1cd412e038df55c8ef96a6bb2/src/utils/mixins.ts#L8-L12). Could we repurpose and use that?

TejasQ:
> Wanted to find out if it okay if I do two different PRs for the issue @TejasQ? I can list does all the components I am satisfied with then you can review them then I proceed to work on the others.

@michaeldera this is perfectly fine! ❤️

src/Breadcrumb/Breadcrumb.tsx
michaeldera: > Hmmm, we seem to be reusing this style quite a bit. We also have this [mixin](https://github.com/contiamo/operational-ui/blob/94f66cca0431a0e1cd412e038df55c8ef96a6bb2/src/utils/mixins.ts#L8-L12). Could we repurpose and use that?

Sure I can do that. So in such cases and similar cases I do away with line 22?

TejasQ: Yes, and use the mixin instead.

michaeldera:
> > Wanted to find out if it okay if I do two different PRs for the issue @TejasQ? I can list does all the components I am satisfied with then you can review them then I proceed to work on the others.
> 
> @michaeldera this is perfectly fine! ❤️

So I am finishing up with this now. So, I remembered why I had opted to go with box-shadow instead of the mixin. So by default the breadcrumb does not have a border. So applying a a border and border color had the effect of causing sibling elements to shift by 2px right (to accomodate the border).

So my assumption had been that the mixin is meant for form input elements which has borders by default. So the approach I am now taking is to apply a transparent border which then applies the mixin without shifting. Like so 

`border: "1px solid transparent",  
  color: theme.color.primary,
  label: "breadcrumb",
  "&:hover": {
    ...inputFocus({ theme  }),
  },
  "&:focus": {
    ...inputFocus({ theme  }),
  },`

That would mean the pattern has to be repeated for all borderless elements that are focused on. Another way is to add a mixin which uses the box-shadow property that is applied to elements without borders. I am not sure which method of implementation you prefer.

TejasQ:
How about we don't do borders at all ever and stick to box-shadow only?

TejasQ:
I think so. AFAIK form input elements need outlines but those can be expressed with box-shadow instead of border with no real penalties. cc @stereobooster who may know more than me in this case.

Is it okay to replace our borders with box-shadows that look like borders? AFAIK the only thing we lose is semantics, but we gain 0px offsets when composing them together.

michaeldera:
> I think so. AFAIK form input elements need outlines but those can be expressed with box-shadow instead of border with no real penalties. cc @stereobooster who may know more than me in this case.
> 
> Is it okay to replace our borders with box-shadows that look like borders? AFAIK the only thing we lose is semantics, but we gain 0px offsets when composing them together.

I think this would work well. I am not aware of penalties either and buttons are already implemented that way.

michaeldera:
> I think so. AFAIK form input elements need outlines but those can be expressed with box-shadow instead of border with no real penalties. cc @stereobooster who may know more than me in this case.
> 
> Is it okay to replace our borders with box-shadows that look like borders? AFAIK the only thing we lose is semantics, but we gain 0px offsets when composing them together.

Any decision on this yet? There is one drawback of using box-shadow instead of border. Border has the border-color property which can be used to change color without regard for border thickness. I am not aware of a way to change box-shadow without specifying the thickness.

src/Breadcrumb/Breadcrumb.tsx
```diff
@@ -19,13 +19,13 @@ const Container = styled("a")(({ theme }) => ({
...
-    color: darken(theme.color.primary, 5),
-    boxShadow: `0 0 0 1px ${theme.color.primary}`,
+    boxShadow: "0 0 0 1px",
+    ...inputFocus({ theme }),
   },
   "&:focus": {
     outline: "none",
-    color: darken(theme.color.primary, 5),
-    boxShadow: `0 0 0 1px ${theme.color.primary}`,
+    boxShadow: `0 0 0 1px`,
```
TejasQ: Why backticks here? (`) Instead of quotes?

src/Breadcrumb/Breadcrumb.tsx
```diff
@@ -19,13 +19,13 @@ const Container = styled("a")(({ theme }) => ({
   color: theme.color.primary,
   label: "breadcrumb",
   "&:hover": {
-    color: darken(theme.color.primary, 5),
-    boxShadow: `0 0 0 1px ${theme.color.primary}`,
+    boxShadow: "0 0 0 1px",
```
TejasQ: I thought the `boxShadow` was part of the mixin?

src/ActionMenu/ActionMenu.tsx
```diff
@@ -35,6 +35,11 @@ const Container = styled("div")(({ theme }) => ({
   cursor: "pointer",
   "&:hover": {
     backgroundColor: darken(theme.color.white, 5),
+    boxShadow: `0 0 0 1px ${theme.color.primary}`,
+  },
+  "&:focus": {
+    backgroundColor: darken(theme.color.white, 5),
+    boxShadow: `0 0 0 1px ${theme.color.primary}`,
```
TejasQ: This is also repeated a bit. Could we add this to the mixin?

michaeldera: This is not done yet. I asked a question related to how you prefer the implementation to be done. Once I get that I will be able to finish up. 


> > I think so. AFAIK form input elements need outlines but those can be expressed with box-shadow instead of border with no real penalties. cc @stereobooster who may know more than me in this case.
> > Is it okay to replace our borders with box-shadows that look like borders? AFAIK the only thing we lose is semantics, but we gain 0px offsets when composing them together.
> 
> Any decision on this yet? There is one drawback of using box-shadow instead of border. Border has the border-color property which can be used to change color without regard for border thickness. I am not aware of a way to change box-shadow without specifying the thickness.

michaeldera:
Implementing using boxShadow means we have to change all the elements that already have borders yet. So does that mean we are using a new mixin or modifying the exiiting one because we will have the case where we have both boxShadows and borders. Not sure if I am being clear here.

src/ActionMenu/ActionMenu.tsx
TejasQ: Oops, sorry, I missed that!

TejasQ:
@michaeldera let's change it everywhere by modifying the existing mixin. The purpose of this PR is to consolidate the focus style so in my opinion, we have a consolidated, central focus style that we export and use everywhere – `box-shadow` or `border` is an implementation detail.

michaeldera:
> I see a lot of `outline: none` repeated everywhere. Would it make sense to add that to the mixin too? 🤔

Yes it would. Let me do that. I will need help with the select. It was a little difficult for me to follow. You  may check if the other things have the desired effect.

michaeldera:
> I see a lot of `outline: none` repeated everywhere. Would it make sense to add that to the mixin too? 🤔

On second thought, adding that to the mixin would only apply to the `:focus` state. In the `:hover` state the outline would still appear. Do we still go ahead?

TejasQ:
Yeah, I think it can be fine to remove the outline on hover. IMHO it doesn't exist on hover anyway.

michaeldera:
> Yeah, I think it can be fine to remove the outline on hover. IMHO it doesn't exist on hover anyway.

Alright, I will refactor that.

TejasQ:
@michaeldera, did you validate this?

image

michaeldera:
> Great job so far. Still,
> 
> * `ActionMenu` has no focus style.
> * `Checkbox` has no focus style.
> * `Checkbox` has no focus style.
> * `DataTableSelect` is broken:
>   image
> * `Select` is broken:
>   image
> * `Switch` is broken:
>   image
> 
> From the above screenshots, this branch is on the left, `master` is on the right. Please observe and rectify the visual regressions.
> 
> I have to say though – **great job so far!** `Textarea` looks amazing!

Having a challenge with the DataTable select.

michaeldera:
> @michaeldera, did you validate this?
> 
> image

Well... there were some react warnings before I started such as anchor tags without the href attribute. I assumed it was WIP in other areas

src/Checkbox/Checkbox.tsx
```diff
@@ -99,7 +100,8 @@ const Label = styled("label")<{ condensed: CheckboxProps["condensed"] }>`
     height: ${({ condensed }) => (condensed ? 10 : 18)}px;
     border-radius: ${props => props.theme.borderRadius}px;
     background-color: #f2f2f2;
-    border: solid 1px #c0c0c0;
+    box-shadow: 0 0 0 1px #c0c0c0;
```
TejasQ: Maybe use a theme constant here?

michaeldera: Done

TejasQ:
@michaeldera did you read this? https://github.com/contiamo/operational-ui/pull/1083#pullrequestreview-258797299

I just checked it – `ActionMenu` still has no focus style. 🤔

michaeldera:
> @michaeldera did you read this? [#1083 (review)](https://github.com/contiamo/operational-ui/pull/1083#pullrequestreview-258797299)
> 
> I just checked it – `ActionMenu` still has no focus style.

Perhaps the way I implemented is not as intended, How would you like to behave?

TejasQ:
I would expect it and other focusable elements to have the same focus style. That's kind of the point of this issue – everything looks the same when focused – with a thin blue line.

michaeldera:
> @michaeldera did you read this? [#1083 (review)](https://github.com/contiamo/operational-ui/pull/1083#pullrequestreview-258797299)
> 
> I just checked it – `ActionMenu` still has no focus style. 🤔

image
image

There is a focus style on `Action Menu` and maybe that is the problem. The `ActionMenu` Component itself implements a `ContextMenu` component. `ContextMenu` is the one with `role=button` attribute. So probably I should move all styles on the `ActionMenu` to the `ContextMenu`?

TejasQ:
Nope. I know now what the problem is. Thanks for clarifying. `ContextMenu` _wraps_ `ActionMenu` and `ContextMenu` gets the focus.

To solve this, you can fork `ContextMenu` in `ActionMenu` and style action menu from above accordingly:

```ts
const StyledContextMenu = styled(ContextMenu)`
  & :focus > div { /* target Container but when ContextMenu is focused */
    /* focus style here */
  }
`
```

Not sure this is the best approach but it works.

michaeldera:
Alright. Will do that today.

TejasQ:
@michaeldera did you test the above commit (https://github.com/contiamo/operational-ui/pull/1083/commits/dd1c6dde5710faa330fe7696b1a248743f8f3700) locally? Does it work?

michaeldera:
> @michaeldera did you test the above commit ([dd1c6dd](https://github.com/contiamo/operational-ui/commit/dd1c6dde5710faa330fe7696b1a248743f8f3700)) locally? Does it work?

Yes I did and it does. I can add a screenshot.

michaeldera:
> > @michaeldera did you test the above commit ([dd1c6dd](https://github.com/contiamo/operational-ui/commit/dd1c6dde5710faa330fe7696b1a248743f8f3700)) locally? Does it work?
> 
> Yes I did and it does. I can add a screenshot.



> @michaeldera did you test the above commit ([dd1c6dd](https://github.com/contiamo/operational-ui/commit/dd1c6dde5710faa330fe7696b1a248743f8f3700)) locally? Does it work?

Are you having issues with it?

TejasQ:
> Are you having issues with it?

Unfortunately so. 😢 

When I press `tab` and focus on the `ActionMenu`, it does not receive a blue outline. Are you able to reproduce this?

michaeldera:
I was using Devtools and the mouse to test. The problem is probably that in the `ContextMenu` itself the element which is supposed to have focus is not the parent element. The element with `...buttonProps` is wrapped inside a styled `Container` component so it will not get focus. It may work to move focus to `ContextMenu`, I could try that.

TejasQ:
The CSS [here](https://github.com/contiamo/operational-ui/pull/1083#issuecomment-509982381) is exactly the solution you need.

michaeldera:
> The CSS [here](https://github.com/contiamo/operational-ui/pull/1083#issuecomment-509982381) is exactly the solution you need.

I did this. I styled the `ContextMenu` that is implemented in `ActionMenu`. But inside of `ContextMenu` itself it appears that the top element is the one that gets focus.

TejasQ:
Do you see the difference in selectors of my snippet and yours?

**[Your Snippet](https://github.com/contiamo/operational-ui/pull/1083/files#diff-2ae3cf3143434940a6ebae9a05a8bfe8R22)**

image

**[My Snippet](https://github.com/contiamo/operational-ui/pull/1083#issuecomment-509982381)**

image

michaeldera:
Sorry about that. Had missed that.

PR #1092, opened 2019-06-25
**Fix:** SectionHeader height

### LINKED ISSUE #1087: Fix SectionHeader height
There is [SectionHeader](https://github.com/contiamo/operational-ui/blob/master/src/Internals/SectionHeader.tsx) which uses

```
height: theme.space.element * 2,
```

This is wrong because we not suppose to use "space" (something around the content) for content size. 

Task is to change it to use `36px` from new design (do we need constant for it or we already have one?), as well find other occurrences of `theme.space.element * 2` and fix them.

PS SectionHeader is reused for Card headers, Tabs and Accordion

march213:
# Summary
Fixes #1087.
I changed all occurrences of `theme.space.element * 2` to `36px`.
Also in this file https://github.com/contiamo/operational-ui/blob/master/src/Tabs/Tabs.styled.ts#L23 I changed it to `56px` because the code was `height: ${({ theme }) => theme.space.element * 2 + 20}px` and remove the overwritten `height` property above.

# Related issue

https://github.com/contiamo/operational-ui/issues/1087

# To be tested

Me
- [x] Tabs component looks and works the same
- [x] Accordion component looks and works the same
- [x] HeaderBar component looks and works the same
- [x] Check other components that use HeaderMenu or SectionHeader

Tester 1

Tester 2

### DIFF
--- a/src/Internals/SectionHeader.tsx
+++ b/src/Internals/SectionHeader.tsx
@@ -1,4 +1,5 @@
 import styled from "../utils/styled"
+import { headerHeight } from "../utils/constants"
 
 export const SectionHeader = styled("div")(({ theme }) => ({
   fontFamily: theme.font.family.main,
@@ -12,7 +13,7 @@ export const SectionHeader = styled("div")(({ theme }) => ({
 
   // This ensures that the card header text and card controls are placed in opposite corners.
   justifyContent: "space-between",
-  height: "36px",
+  height: headerHeight,
   padding: `0 ${theme.space.element}px`,
   lineHeight: 1,
 }))
--- a/src/Internals/SectionHeader.tsx
+++ b/src/Internals/SectionHeader.tsx
@@ -12,7 +12,7 @@ export const SectionHeader = styled("div")(({ theme }) => ({
 
   // This ensures that the card header text and card controls are placed in opposite corners.
   justifyContent: "space-between",
-  height: theme.space.element * 2,
+  height: "36px",
   padding: `0 ${theme.space.element}px`,
   lineHeight: 1,
 }))
--- a/src/Tabs/Tabs.styled.ts
+++ b/src/Tabs/Tabs.styled.ts
@@ -6,21 +6,20 @@ const buttonWidth = 55
 export const Container = styled.div`
   label: Tabs;
   display: grid;
-  grid-template-rows: ${({ theme }) => `${theme.space.element * 2}px 1fr`};
+  grid-template-rows: 36px 1fr;
   position: relative;
   height: 100%;
 `
 
 export const TabList = styled.div<{ scroll: boolean }>`
   display: flex;
-  height: ${({ theme }) => theme.space.element * 2}px;
   overflow-x: auto;
   max-width: ${({ scroll }) => (scroll ? `calc(100% - ${buttonWidth * 2}px)` : "none")};
   scroll-behavior: smooth;
   border-left: solid 1px ${({ theme }) => theme.color.separators.default};
   overflow-y: hidden;
   /* magic number to hide scroll bar underneath tabpanel */
-  height: ${({ theme }) => theme.space.element * 2 + 20}px;
+  height: 56px;
   -webkit-overflow-scrolling: auto;
   ::-webkit-scrollbar {
     display: none;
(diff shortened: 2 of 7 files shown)


src/Internals/SectionHeader.tsx
```diff
@@ -12,7 +12,7 @@ export const SectionHeader = styled("div")(({ theme }) => ({
 
   // This ensures that the card header text and card controls are placed in opposite corners.
   justifyContent: "space-between",
-  height: theme.space.element * 2,
+  height: "36px",
```
fabien0102:
```suggestion
  height: 36,
```

You don't need to specify `px`, a number is enough here 😉

TejasQ:
Wow! Great work, @march213!

I noticed an odd side effect as soon as I opened the [PR preview](https://deploy-preview-1092--operational-ui.netlify.com/) and clicked `diff with master`: `Card` titles are somehow smaller. 🤔 

Check it out: 
image

stereobooster:
@TejasQ is it ok that we reuse 36 everywhere instead of introducing const? (And all of those 36 are the same thing height of header-like element)

TejasQ:
Yeah @stereobooster it looks a little too wet (not DRY) for me too that it's concerning. Do you think it makes more sense to add it to the theme? I don't think it makes sense in the theme because it's not really something we want to be customizable, right? Maybe we can just export a `const`?

stereobooster:
Const works as well. My may concern is that this value represent the same concept but from code it is hard to tell, because it's just a number 36

TejasQ:
That's a great point. So then @march213, let's not add it to the theme but 

`export const headerHeight = 36`

from [here](https://github.com/contiamo/operational-ui/blob/64ff059da232c3eaa68b5efcceee6e0c741f3776/src/utils/constants.ts) and use it everywhere to convey the meaning behind the number.

Great observation, @stereobooster!

src/Internals/SectionHeader.tsx
march213: thanks! didn't notice that

march213:
@TejasQ Done ✅

src/utils/constants.ts
```diff
@@ -268,6 +268,8 @@ const constants = {
   },
 }
 
+export const headerHeight: number = 36
```
TejasQ: We can remove the `: number` from here but not a big deal.

march213: Yeah, actually I put it here, because in some places it should be a number ... But can remove it 🤔

TejasQ: It's fine.

stereobooster: TypeScript knows that `headerHeight` is a number without using signature in this case. They use so called [Type Inference](https://www.typescriptlang.org/docs/handbook/type-inference.html). Type checker can "guess" the type in this case, but it can't guess type in all cases, when it can't guess it will assign type `any` silently, unless there is `noImplicitAny` flag is enabled (it is in this repository). So you can skip all annotation and type checker will complain about those it can't guess. (As well sometime it guesses wrong, like `string` instead of string literal)

src/Tabs/Tabs.styled.ts
```diff
@@ -1,26 +1,26 @@
...
 import { SectionHeader } from "../Internals/SectionHeader"
+import { headerHeight } from "../utils/constants"
 
 const buttonWidth = 55
 
 export const Container = styled.div`
   label: Tabs;
   display: grid;
-  grid-template-rows: ${({ theme }) => `${theme.space.element * 2}px 1fr`};
+  grid-template-rows: 36px 1fr;
```
stereobooster:
```suggestion
  grid-template-rows: ${headerHeight}px 1fr;
```

src/utils/constants.ts
stereobooster:
```suggestion
export const headerHeight = 36
```

march213:
Should I make new PR ? I’ve made suggested changes

src/Tabs/Tabs.styled.ts
march213: Ooops, my bad, sorry about that

TejasQ: Not a problem.

TejasQ:
@stereobooster did the rest: https://github.com/contiamo/operational-ui/pull/1092

Thanks for your amazing work, @march213!

PR #1095, opened 2019-06-26
**Feature: ** Add icons to TopbarSelect.

ImogenF:
# Summary

Updated `TopbarSelect` component to enable icons in the button itself. Component styling has also been updated to reflect new specs.

image

# Related issue

# To be tested

Me

Tester 1

Tester 2

### DIFF
--- a/src/TopbarSelect/README.md
+++ b/src/TopbarSelect/README.md
@@ -24,16 +24,6 @@ const displayItems = [
   { label: "Waterfall Chart", icon: WaterfallChartIcon },
 ]
 
-const icons = {
-  Table: TableIcon,
-  "Bar Chart": BarChartIcon,
-  "Line Chart": LineGraphIcon,
-  Counter: CounterIcon,
-  "Pie Chart": PieChartIcon,
-  "Cohort Chart": CohortChartIcon,
-  "Waterfall Chart": WaterfallChartIcon,
-}
-
 const [activeFruit, setActiveFruit] = React.useState(undefined)
 const [activeDisplay, setActiveDisplay] = React.useState("Table")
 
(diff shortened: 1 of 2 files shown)


src/TopbarSelect/README.md
```diff
@@ -1 +1,59 @@
...
+      selected={activeFruit}
+      items={["apples", "oranges"].map(name => ({ label: name }))}
+      onChange={item => setActiveFruit(item)}
+    />
+  }
+  right={
+    <TopbarSelect
+      label="Display:"
+      selected={activeDisplay}
+      selectedIcon={icons[activeDisplay]}
```
stereobooster: I feel like this is wrong to require provide `selectedIcon` if we already provide `selected` and `items`

ImogenF: Good point. Will change it this evening.

src/TopbarSelect/TopbarSelect.tsx
```diff
@@ -38,24 +41,23 @@ const TopbarSelectValue = styled("div")`
...
 `
 
-const TopbarSelectValueSpan = styled("span")<{ active?: boolean }>`
-  color: ${props => (props.active ? props.theme.color.text.dark : props.theme.color.text.lighter)};
+const TopbarSelectValueSpan = styled("span")`
+  margin-right: ${props => props.theme.space.element}px;
+`
+
+const TopbarSelectIcon = styled("div")`
+  paddingleft: ${props => props.theme.space.base}px;
```
TejasQ: This is invalid CSS. I think you meant

```suggestion
  padding-left: ${props => props.theme.space.base}px;
```

stereobooster:
Label "jumps" do we want to address this in the PR or create issue?
Screenshot 2019-06-26 at 15 32 57
Screenshot 2019-06-26 at 15 33 05

TejasQ:
Depends on how much time effort it is? If it's just an alignment we can fix it in this PR.

stereobooster:
It can take some time, because right now there is no concept of max-width in this component, so we need to add that and decide if it suppose to be turned on by default or not

ImogenF:
Was just taking a look at this now. Think it would be good to go ahead and merge so I can put this component in Explore, but I'll try and fix it tonight, unless there's anything more urgent?

PR #1105, opened 2019-06-28
**Fix:** 1078 code examples

bradd123:
# Summary

# Related issue

# To be tested

Me

Tester 1

Tester 2


src/Autocomplete/README.md
```diff
@@ -43,6 +43,11 @@ const MyComponent = () => {
       })
   }, [dogToFetch])
 
+  const onResultClickUtil = result => {
```
TejasQ: Use `useCallback` here to not recompute the function every time.

src/CardSection/CardSection.tsx
```diff
@@ -9,7 +9,7 @@ import { headerHeight } from "../utils/constants"
 
 export interface CardSectionProps extends DefaultProps, DragProps {
   /** Column title */
-  title?: string
+  title?: React.ReactElement<any> | string
```
TejasQ: Would `React.ReactNode` work here?

src/ContextMenu/README.md
```diff
@@ -62,7 +62,7 @@ import * as React from "react"
 import { AddIcon, ContextMenu, ContextMenuProps, Contact, Hint, Button } from "@operational/components"
 
 /* Anything can be a label now, even some DIV you style yourself */
-const MyLabelContainer = ({ children, style }) => (
+const MyLabelContainer = ({ children, style = {} }) => (
```
TejasQ: We can just remove the `style` prop and its spread on the next line.

src/Page/README.md
```diff
@@ -200,7 +200,7 @@ const Tab = props => (
   </PageContent>
 )
 
-class Router extends React.Component {
+class Router extends React.Component<{}, any> {
```
TejasQ: Would you be willing to rewrite this with hooks? It could sharpen your hooks skills too!

src/Select/README.md
```diff
@@ -22,7 +22,7 @@ const options = [
 ]
 
 const MyComponent = () => {
-  const [value, setValue] = React.useState("one")
+  const [value, setValue] = React.useState < string > "one"
```
TejasQ: Is this valid TS? It looks more like elm...

bradd123: That was prettier in work. I wanted to write `React.useState<string>("one")`.

src/Page/README.md
bradd123: Cool. will do that.

TejasQ:
@bradd123 what up do you need help with anythang

TejasQ:
A week later, paging captiain @bradd123. How's it going? Can I help with anything?

bradd123:
Hi, Most of the things are done. I am having difficulty understanding the following component type errors. `DataTable`, `DataTableInput`, `HeaderMenu`, `LabelText`, `Page`, `Select`, `Tree`, `FinePrint`.

bradd123:
I will create a separate PR to convert `import * as React from 'react'` to `import React from 'react'`.

bradd123:
I have seen multiple type errors like this. `Type 'number' is not assignable to type 'SetStateAction<string>'` What is the solution for these types? Any good references would be welcome.

TejasQ:
> I have seen multiple type errors like this. Type 'number' is not assignable to type 'SetStateAction<string>'

Pointing me to where you see these errors would be a start to helping me help you. 💞

TejasQ:
To fix the type errors:

## About `DataTable`
- `DataTableFooter` apparently has no `onClick` prop! Odd! Maybe we should add one. Open an issue if you don't want to fix it in this PR.
- You'll need to rewrite the `getNewRows` function like so:

```diff
const getNewRows = (lastIndex = 0): Array<[string, number, number]> =>
  Array.from({ length: lastIndex > 1000000 ? 1 : (lastIndex || 1) * 10 }, (_, i) => [
    `Cell ${i + lastIndex + 1}`,
    Math.random(),
    Math.random(),
-  ])
+  ]) as Array<[string, number, number]>
```

TypeScript doesn't really infer types of matrices well enough so we need to add a type assertion here and everything will be fine.

## About `DataTableInput`
Its [`value` prop](https://github.com/contiamo/operational-ui/blob/7f3a10e1e24ff04b90a54940644d3c9340ebaa1c/src/Input/Input.tsx#L20) is a `string`, but in this README example, we're [passing it a `number`](https://github.com/contiamo/operational-ui/blame/7f3a10e1e24ff04b90a54940644d3c9340ebaa1c/src/DataTableInput/README.md#L11). The correct solution would be to coerce the `number` we pass to `useState` to a `string`:

```diff
-  const [age, setAge] = React.useState(new Date().getFullYear() - 1993)
+  const [age, setAge] = React.useState(String(new Date().getFullYear() - 1993))
```

## About `HeaderMenu`

TypeScript has no idea what [`projectOptions`](https://github.com/contiamo/operational-ui/blame/7f3a10e1e24ff04b90a54940644d3c9340ebaa1c/src/HeaderMenu/README.md#L9) is and it can't figure it out on its own. We need to help it out here by telling it:

```diff
- const projectOptions = [
+ const projectOptions: ContextMenuProps["items"] = [
```

## About `LabelText`

There is no valid `htmlFor` prop on `span` elements – [which it is](https://github.com/contiamo/operational-ui/blame/7f3a10e1e24ff04b90a54940644d3c9340ebaa1c/src/LabelText/LabelText.tsx#L4). The solution in this case is to remove the prop from the README.

## About `Page`
1. `styled` is not imported from Operational UI so TypeScript doesn't know what it is. The solution is to add it to the list of imports.
1. You can remove the whole `Actions` component and its import to fix the other error. It's an internal component that we should either externalize or unexpose.

## About `Select`

A `value` given to the first argument of `onChange` can be `String | Number | ...`. What you'll need to do is assert what the new value actually is when you call `setState` like this:

```diff
<Select
        id="basic"
        label="Basic"
        data-cy="basic-select"
        value={value}
        options={options}
        onChange={(newValue, lastChanged) => {
-          setValue(newValue)
+          setValue(newValue as string)
          setLastChanged(lastChanged)
        }}
      />
```

## About `Tree`

This is a little more involved. Let's make an issue and deal with it later.

## About `FinePrint`

This is an odd error. I opened an issue about it on emotion (https://github.com/emotion-js/emotion/issues/1431). Let's see what they say. We can open an issue for this and continue with the PR.

bradd123:
Thanks for the detailed explanation. I will make the changes and push tonight.

bradd123:
`HeaderMenu` has a weird problem. Directly editing works, but changing the file is not working. The following error is showing.

> Error: import or require() statements can be added only by editing a Markdown example file: @operational/components

TejasQ:
I know this problem. I think @stereobooster once solved it really well. I wonder where his code is... 👀

TejasQ:
Here!

https://github.com/contiamo/operational-ui/commit/728f18ffce5c595dd073308faeb38b58bc879de1

Apparently removing a type annotation resolves it.

bradd123:
What if we need the type annotation? Remove annotation throws type errors.

TejasQ:
I honestly don't know 😂

src/Card/Card.tsx
```diff
@@ -13,6 +13,7 @@ export interface BaseProps extends DefaultProps {
    */
   fullSize?: boolean
   onClick?: (event: React.MouseEvent<HTMLDivElement>) => void
+  style?: React.CSSProperties
```
TejasQ: Wy do we need this?

bradd123:
What is your preferred way to merge the code? Do you rebase or directly merging?

src/OperationalUI/README.md
```diff
@@ -23,9 +23,9 @@ To set up routing that is automatically wired up to the `to` props of all nested
 
 ```jsx
-import { Button } from "@operational/components"
+import { Button, OperationalUI, OperationalUIProps } from "@operational/components"
 
-class RoutingComponent extends React.Component {
+class RoutingComponent extends React.Component<OperationalUIProps, any> {
```
TejasQ: no `any`. Write a definition of the state. Should just be `{ path: string }`

TejasQ: Even better, this could be rewritten as hooks if you want the knowledge/fun.

bradd123:
The remaining components are `DataTable`, `HeaderMenu`, `LabelText`, `Tree`, and `FinePrint`. Each one has separate issues. We will deal with them separately. Please review the existing ones.

Next, I will create a separate PR that will convert all `import * as React from "react"` to `import React from "react"`. I have already changed compiler options in this PR.

src/Card/Card.tsx
bradd123: We don't actually need that. I added that a week back. I will update this file and doc.

src/OperationalUI/README.md
bradd123: Sure, will convert it into a hook

bradd123: Done

src/Card/Card.tsx
bradd123: Updated this component.

TejasQ:
Great! If you want to deal with all the other issues separately, would you go ahead and make issues for them so we can keep track of them and open corresponding PRs? That way we can merge this one.

bradd123:
Sure, I will create separate issues for them.

TejasQ:
Finally, it would be great if you filled this out:

image

Attention to detail is something we value and encourage on this project. I think once that's done we can take a train to merge city. 🚋

bradd123:
I am having the following warnings in the console.
Operational_UI_and_Slack_-_RemoteIndian
Operational_UI
Operational_UI
Operational_UI

TejasQ:
Right – those existed before. Great. I think then we're good.

Issue #1099, opened 2019-06-27
TopBarSelect types are broken


stereobooster:
Consider following code example

```ts
import { BarChartIcon, TableIcon, TopbarSelect } from "@operational/components"

const displayItems = [{ label: "Table", icon: TableIcon }, { label: "Bar Chart", icon: BarChartIcon }]

const [activeDisplay, setActiveDisplay] = React.useState<any>("Table")

;<TopbarSelect
  label="Display:"
  selected={activeDisplay}
  items={displayItems}
  onChange={setActiveDisplay}
/>
```

It doesn't work without `any` in `React.useState`

march213:
Hi 👋🏻 
Can I take this one too ?

ImogenF:
@march213 Go for it!

march213:
@ImogenF Thank you !

march213:
@stereobooster Would it be right to change `onChange` type to this

```
onChange?: (
    newLabel:
      | string
      | React.ReactElement<any>
      | React.Dispatch<React.SetStateAction<string | React.ReactElement<any>>>
  ) => void
```

It works, because in this scenario you are passing `setActiveDisplay` to `onChange`. But `setActiveScenario` has type `React.Dispatch<React.SetStateAction<string>>` which is incompatible with current `onChange?: (newLabel: string | React.ReactElement<any>) => void`

I'm not a TS expert yet 🤓, so maybe I'm wrong.

stereobooster:
@march213 consider following code (standard React)

```js
function App() {
  const [state, setState] = React.useState('mango');
  const onChange = React.useCallback((event) => setState(event.target.value), [])

  return (
    <select value={state} onChange={onChange}>
      <option value="grapefruit">Grapefruit</option>
      <option value="lime">Lime</option>
      <option value="coconut">Coconut</option>
      <option value="mango">Mango</option>
    </select>
  );
}
```

React standard components have to deal with event, because they are building blocks. But in our case we can simplify API and onChange can be called with value directly instead of event. (Does it make sense so far?)

So we should be able to write something like this

```
const displayItems = [{ label: "Table", icon: TableIcon }, { label: "Bar Chart", icon: BarChartIcon }];
const [activeDisplay, setActiveDisplay] = React.useState("Table");

<TopbarSelect
  label="Display:"
  selected={activeDisplay}
  items={displayItems}
  onChange={setActiveDisplay}
/>
```

[...]

PR #1108, opened 2019-06-28
**Fix:** TopBarSelect types are broken

march213:
# Summary
Fixes https://github.com/contiamo/operational-ui/issues/1099
Change type signature for `TopbarSelect`.
Change the provided function parameter type from `string` to `IContextMenuItem`.

# Related issue

https://github.com/contiamo/operational-ui/issues/1099

# To be tested

Me
- [x] No TS errors

Tester 1

- [ ] Should not be any TS errors
  

Tester 2

- [ ] Should not be any TS errors

### DIFF
--- a/src/Topbar/README.md
+++ b/src/Topbar/README.md
@@ -9,7 +9,7 @@ import {
   TopbarSelect,
   TopbarButton,
   TopbarSeparator,
-  styles,
+  styled,
   NoIcon,
   UndoIcon,
   RedoIcon,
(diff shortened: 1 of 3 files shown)


stereobooster:
@march213 can you please fix example in the playground as well?

Screenshot 2019-07-01 at 17 44 22

march213:
I fixed following examples:
- `Topbar/README.md`
- `TopbarSelect/README.md`

stereobooster:
Screenshot 2019-07-02 at 10 06 34

If you will go to online playground you will see that there is TS error. TS complains that value can string and object, but we handle only object case.

march213:
Yeah, I was checking online playground locally, making sure it is working ... And didn’t notice any ts errors 🤔
I’ll fix it

march213:
I'm a little bit confusing.
I'm not sure how to fix it, sorry. 😢

The following code fix that problem:
```ts
const [activeDisplay, setActiveDisplay] = React.useState(undefined)
```

But if you would set the default parameter to a `string`, you'll get an error.

TejasQ:
I think as a first iteration, coercing `item` to a string in the README example, so it becomes

```js
onChange={item => setActiveDisplay(String(item.label))}
```

will be enough to get this merged. At a later stage, we might want to think about improving the interfaces of ContextMenu and components that use it to harness the power of generics to dictate what their items and their labels are. This, I think is a bigger topic. Let's open an issue for it.

Not sure how deep your TypeScript interest is, @march213, but that could be fun if you want to take it on separate to this PR.

stereobooster:
@TejasQ proposed solution here https://github.com/contiamo/operational-ui/issues/1099#issuecomment-506776580

march213:
@stereobooster Yeah, when I tried this one `onChange={item => typeof item.label === "string" && setActiveDisplay(item.label)}` it was based on the proposed solution. But if I move typeof checking to a function, it stop working 😥

image

march213:
@TejasQ My Typescript interest is very high, because I see the benefits of using it and I really want to learn it properly.

march213:
I used this one for fix:
```ts
onChange={item => setActiveDisplay(String(item.label))}
```

Issue #1117, opened 2019-07-03
Need to fix unfiltered props in styled components


stereobooster:
# The Problem

<img width="636" src="https://user-images.githubusercontent.com/179534/60587780-96f77b00-9d95-11e9-9d92-ef36d326463a.png" alt="index.js:3671 Warning: Received `false` for a non-boolean attribute `fill`. If you want to write it to the DOM, pass a string instead: fill='false' or fill={value.toString()}. If you used to conditionally omit it with fill={condition && value}, pass fill={condition ? value : undefined} instead.    in div (created by Context.Consumer)    in Styled(div) (created by Tabs)">

# The Solution

Let's use [Emotion's `shouldForwardProp`](https://emotion.sh/docs/styled#customizing-prop-forwarding) to filter out non-standard attributes on DOM nodes across the library.

# How do I know it's fixed?

There should be no errors or warnings in the console on `localhost:6060` on each component page in the styleguide.

march213:
Hi 👋🏻
I'd like to take this one too.

TejasQ:
Sure! Appreciate your great work!

march213:
@stereobooster @TejasQ 

Hi! I found a problem:
- An Icon component accepts right and left prop. But as I can see, icons are created on build stage and all props are spreaded to an `<svg>`:

```ts
export const RemoveIcon = ({ size = 18, color, left, right, ...props }: IconProps) => {
  const iconColor: string = expandColor(constants, color) || "currentColor"
  const style = {
    // theme.space.small
    marginLeft: right ? 8 : 0,
    // theme.space.small
    marginRight: left ? 8 : 0,
    cursor: Boolean(props.onClick) ? "pointer" : "default",
    transition: "fill .075s ease",
  }
  return (
    <svg viewBox="0 0 17 17" fill={iconColor} width={size} height={size} style={style} {...props}>
      ...
    </svg>
  )
}

```

I'm not sure, how to modify `build-icons.ts` script to prevent it... Any suggestions ?

stereobooster:
@march213 what the problem exactly? 

<img width="875" alt="Screenshot 2019-07-05 at 17 20 52" src="https://user-images.githubusercontent.com/179534/60732025-458fed00-9f49-11e9-826c-382b861ee830.png">

`...props` contains everything except `size, color, left, right`

march213:
@stereobooster I thought that the problem is here:

`Button.tsx`
```ts
[...]

PR #1121, opened 2019-07-05
**Fix:** Fix unfiltered props in styled-components

march213:
fix icons import and use in playgrounds

# Summary

Fixes https://github.com/contiamo/operational-ui/issues/1117

Fix unfiltered props in styled-components
Fix icons import and use in playgrounds

# Related issue

https://github.com/contiamo/operational-ui/issues/1117

# To be tested

Me
- [ ] No more error or warning in the console on `localhost:6060` about props

Tester 1

Tester 2

### DIFF
--- a/src/Card/Card.tsx
+++ b/src/Card/Card.tsx
@@ -119,7 +119,7 @@ const Container = styled("div")(({ theme }) => ({
   },
 }))
 
-const Content = styled("div")<{ fullSize?: boolean }>`
+const Content = styled("div", { shouldForwardProp: prop => prop !== "fullSize" })<{ fullSize?: boolean }>`
   padding: ${({ theme }) => theme.space.element}px;
   height: ${props => (props.fullSize ? "100%" : "auto")};
   white-space: pre-wrap;
--- a/src/Card/Card.tsx
+++ b/src/Card/Card.tsx
@@ -119,7 +119,7 @@ const Container = styled("div")(({ theme }) => ({
   },
 }))
 
-const Content = styled("div", { shouldForwardProp: prop => prop !== "fullSize" })<{ fullSize?: boolean }>`
+const Content = styled.div<{ fullSize?: boolean }>`
   padding: ${({ theme }) => theme.space.element}px;
   height: ${props => (props.fullSize ? "100%" : "auto")};
   white-space: pre-wrap;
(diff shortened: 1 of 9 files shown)


src/Card/Card.tsx
```diff
@@ -119,7 +119,7 @@ const Container = styled("div")(({ theme }) => ({
   },
 }))
 
-const Content = styled("div")<{ fullSize?: boolean }>`
+const Content = styled("div", { shouldForwardProp: prop => prop !== "fullSize" })<{ fullSize?: boolean }>`
```
stereobooster:
```suggestion
const Content = styled.div<{ fullSize?: boolean }>`
```

this should work as well

march213:
@TejasQ Sure! Didn't know that.

march213:
I've changed it for the Page component but it doesn't work.

__Error:__
image

__Component:__
image

stereobooster:
So the issue here is following:

`styled.div` uses built in `shouldForwardProp` which filters out everything except valid html props. `fill` is valid prop for svg elements, so it's passed down `¯\_(ツ)_/¯`

You can use `styled.div` where there are no standard props and `styled("div", { shouldForwardProp: ...` otherwise

march213:
@stereobooster Thank you! I learn so much from you guys ☺️

march213:
I made changes.
Where it is possible I use `.div` instead of ("div").

I'm not sure about `SimpleLink` component. How to write `.button` with this syntax:

```ts
const BaseSimpleLink = styled<"a" | "button">("button", {
  shouldForwardProp: prop => !["color_", "left_", "right_"].includes(prop),
})
```

TejasQ:
How about 

```diff
- const BaseSimpleLink = styled<"a" | "button">("button", {
+ const BaseSimpleLink = (styled<"a" | "button">).button``
```

march213:
I'll try this one.

march213:
I tried this one, but it doesn’t work. 
If you don’t mind, I’d like to read typescript docs on the weekend  for better understanding how it works.

TejasQ:
That's a good idea. Maybe we can not block merging because of this though – how about we open a separate issue for this one case and deal with it separately? Then we can merge this PR too. 🎉

march213:
I’d like to work on it separately 🚀

TejasQ:
GREAT. Open the relevant issue and let's get this merged.

march213:
I opened the separate issue https://github.com/contiamo/operational-ui/issues/1134

PR #1125, opened 2019-07-09
**Fix:** Datepicker component: open the datepicker component on the correct month and year

### LINKED ISSUE #909: Open the DatePicker on the correct page if `max` is set
## Summary

Actually the `DatePicker` component always open on the current date, even if we have the `max` props set.

It's very annoying so let's fix this! 😄

rimildeyjsr:
Add the getStartMonthYearInWidget(props) method to Datepicker.tsx to set start month, year the datepicker should be opened to, based on the availability of start, end, max, min props. If no props are available, the current date is used.
Fixes #909

# Summary

Fixes #909

The datepicker component always opens on the current month and year when the start props is not available. Ideally, it should open on the month and year of any one of the props available, and should fallback to the current month and year only when none of the props are available.

# Related issue
#909 

# To be tested

Me

Tester 1

Tester 2

### DIFF
--- a/src/DatePicker/DatePicker.tsx
+++ b/src/DatePicker/DatePicker.tsx
@@ -7,7 +7,7 @@ import { Label } from "../utils/mixins"
 import { keyCodes } from "../utils"
 import Month from "./DatePicker.Month"
 import { Container, DatePickerCard, IconContainer, Input, MonthNav, Toggle } from "./DatePicker.styles"
-import { changeMonth, months, toDate, toYearMonthDay, validateDateString } from "./DatePicker.utils"
+import { changeMonth, months, toDate, validateDateString, getStartMonthYearInWidget } from "./DatePicker.utils"
 
 export interface DatePickerProps extends DefaultProps {
   id?: string
@@ -43,7 +43,7 @@ class DatePicker extends React.Component<DatePickerProps, State> {
     this.validate(props) // Start year month is either based on the start date
     // or the current month if no start date is specified.
 
-    const startYearMonthInWidget = this.getStartMonthYearInWidget(props)
+    const startYearMonthInWidget = getStartMonthYearInWidget(props)
 
     this.state = {
       ...startYearMonthInWidget,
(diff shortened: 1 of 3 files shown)


src/DatePicker/DatePicker.tsx
```diff
@@ -78,6 +71,36 @@ class DatePicker extends React.Component<DatePickerProps, State> {
     }
   }
 
+  public getStartMonthYearInWidget(props?: DatePickerProps) {
+    // set start month and year of the widget based on the availability of start, end, max, or min props.
+    // if none of the props are available, set it to the current month and year.
+    const validatedProps = props || this.props
```
TejasQ: So this function is going to be recomputed on every render. This isn't ideal. 

I think to make it a little more performant, you could move this function to [DatePicker.utils.ts](https://github.com/contiamo/operational-ui/blob/1b5389bae8c376761e43c9c14fca802892b634c9/src/DatePicker/DatePicker.utils.ts) and call it from inside here, passing props to it.

Does that make sense?

rimildeyjsr: Yup that does! will make the changes and add them in!

rimildeyjsr:
@TejasQ - Moved the `getStartMonthYearInWidget()` function out of the class and added unit tests for the `getStartMonthYearInWidget()`

Let me know if there are any issues 😃

TejasQ:
Also @rimildeyjsr did you test this locally?
image

rimildeyjsr:
@TejasQ - I had thought about subsets, but I included all so as to not miss any edge case. Any minor case missed would lead it back to having the current date in the date picker. Also, subsets quickly got complicated so I used the if-else ladder.

Can we merge?  😅

rimildeyjsr:
@TejasQ - there are errors in the console, but the same errors are also there for the `master` branch. No new warnings or errors were introduced after my changes

TejasQ:
Really good work, @rimildeyjsr! Amazing!

rimildeyjsr:
@TejasQ - thanks! I learned a lot! Also, first time writing unit tests 😄

TejasQ:
@rimildeyjsr weird – this worked on the CI but broke on `master`. If you checkout `master` and `yarn jest`, you'll see the unit test is broken. Would you like to fix it or shall I?

rimildeyjsr:
@TejasQ - I will check it out!

rimildeyjsr:
@TejasQ - I found the issue, the months and days are zero-indexed, hence the tests were failing. For eg, the unit test expected June(6) to return 5 (because 0 based) but was expecting 6. How should I add the code, do I raise a new PR?

TejasQ:
Yup.

TejasQ:
I still don’t know how the tests passed on CI for the previous PR.

rimildeyjsr:
@TejasQ - Fix for the broken unit tests

https://github.com/contiamo/operational-ui/pull/1130

PR #1130, opened 2019-07-12
**Fix:** DatePicker unit test

rimildeyjsr:
# Summary
The unit tests for Datepicker component was breaking. The months and days are zero-based, hence subtracted 1 from the expected value in the tests.

# Related issue
Was mentioned on PR #1125

# To be tested

Me
- [ ] No error or warning in the console on `localhost:6060`

Tester 1

- [ ] Things look good on the demo.
  

Tester 2

- [ ] Things look good on the demo.

PR #1131, opened 2019-07-12
**Fix:** tab focus steal

TejasQ:
This PR fixes this:

Contiamo

# [Click Here for Demo](https://deploy-preview-1131--operational-ui.netlify.com/#!/Tabs/5)

### DIFF
--- a/src/Tabs/Tabs.tsx
+++ b/src/Tabs/Tabs.tsx
@@ -181,6 +181,7 @@ const Tabs: React.FC<TabsProps> = ({
               aria-selected={false}
               condensed={true}
               onMouseDown={e => {
+                userAction.current = true
                 e.preventDefault()
                 onInsert(tabs.length - 1)
               }}
--- a/src/Tabs/Tabs.tsx
+++ b/src/Tabs/Tabs.tsx
@@ -65,9 +65,9 @@ const Tabs: React.FC<TabsProps> = ({
 
   const onKeyDown = React.useCallback(
     (event: React.KeyboardEvent) => {
-      userAction.current = true
       switch (event.key) {
         case "ArrowRight":
+          userAction.current = true
           event.preventDefault()
           if (active + 1 >= tabs.length) {
             onActivate(0)
@@ -76,6 +76,7 @@ const Tabs: React.FC<TabsProps> = ({
           }
           break
         case "ArrowLeft":
+          userAction.current = true
           event.preventDefault()
           if (active - 1 < 0) {
             onActivate(tabs.length - 1)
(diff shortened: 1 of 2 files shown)


TejasQ:
Unit tests are failing because of @rimildeyjsr's broken unit test on `master`. There's already a PR for it: https://github.com/contiamo/operational-ui/pull/1130

Let's not block merging this because of CI because `master` is also red until we merge #1130. 😄

src/Tabs/Tabs.tsx
```diff
@@ -65,9 +65,9 @@ const Tabs: React.FC<TabsProps> = ({
 
   const onKeyDown = React.useCallback(
     (event: React.KeyboardEvent) => {
-      userAction.current = true
```
stereobooster: ouch

TejasQ: I just moved it down into each key. I wasn't sure why we wanted to set it true on all keystrokes because then we would listen to a child textarea's keystrokes too... 😬

src/Tabs/Tabs.tsx
```diff
@@ -178,7 +182,6 @@ const Tabs: React.FC<TabsProps> = ({
               condensed={true}
               onMouseDown={e => {
                 e.preventDefault()
-                userAction.current = true
```
stereobooster: Go to first example and click plus sign you will see a bug https://deploy-preview-1131--operational-ui.netlify.com/#!/Tabs

stereobooster: We can as well create a ticket for it, because this bug is less disturbing then still of focus

TejasQ: You are amazing.

TejasQ: You are amazing.

TejasQ: Is the bug that thew new tab has no focus?

src/Tabs/Tabs.tsx
stereobooster: Yeah I understood, I said "ouch", because I added it there. 

But now on the second thought 

> we would listen to a child textarea's keystrokes too... 😬

but they are in different subtrees, aren't they? so it actually won't listen to textarea. (the change is good independently of this)

src/Tabs/Tabs.tsx
stereobooster: yes. I can create a bug report.

TejasQ:
> for which I will create a ticket

I just fixed it 😍

src/Tabs/Tabs.tsx
micha-f: Could also leave it where it was and add a default case that sets it to `false`.

PR #1147, opened 2019-07-17
**Fix:** Allow using a react component as a DataTable cell

mpotomin:
# Summary
Fixes the regression, introduced by https://github.com/contiamo/operational-ui/pull/1132/ and allows back using a React component as a custom cell for the `<DataTable/>` instead of getting `[object Object]`

# Related issue
https://contiamo.atlassian.net/browse/PLAT-732

# To be tested

Me

Tester 1

Tester 2

### DIFF
--- a/src/DataTable/DataTable.tsx
+++ b/src/DataTable/DataTable.tsx
@@ -40,6 +40,10 @@ export interface DataTableProps<Columns, Rows> {
   className?: string
 }
 
+const stringifyIfNeeded = (value: any) => {
+  return typeof value === "boolean" ? String(value) : value
+}
+
 export function DataTable<Columns extends any[][], Rows extends any[][]>({
   columns,
   rows,
@@ -121,7 +125,7 @@ export function DataTable<Columns extends any[][], Rows extends any[][]>({
                 cell={cellIndex + 1}
                 height={rowHeight}
               >
-                {truncate(maxCharactersInCell)(String(cell))}
+                {truncate(maxCharactersInCell)(stringifyIfNeeded(cell))}
               </Cell>
             ))}
         </Row>
--- a/src/DataTable/DataTable.tsx
+++ b/src/DataTable/DataTable.tsx
@@ -41,6 +41,7 @@ export interface DataTableProps<Columns, Rows> {
 }
 
 const stringifyIfNeeded = (value: any) => {
+  // We compare booleans like this and without typeof for perf
   return value === true || value === false ? String(value) : value
 }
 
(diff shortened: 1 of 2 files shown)


src/DataTable/DataTable.tsx
```diff
@@ -40,6 +40,10 @@ export interface DataTableProps<Columns, Rows> {
   className?: string
 }
 
+const stringifyIfNeeded = (value: any) => {
+  return typeof value === "boolean" ? String(value) : value
```
stereobooster:
```suggestion
  return (value === true || value === false) ? String(value) : value
```

would be faster, but this is optional thing

stereobooster: Should we worry about other special values like, null, NaN, undefined, {}, []. Or TS won't let you use those?

stereobooster: Screenshot 2019-07-17 at 14 32 54

mpotomin: `null`, `undefined`, etc are valid React children, so I expect that to be handled in the userland. Original https://github.com/contiamo/operational-ui/pull/1132 was dealing with booleans only

src/DataTable/DataTable.tsx
```diff
@@ -40,6 +40,10 @@ export interface DataTableProps<Columns, Rows> {
   className?: string
 }
 
+const stringifyIfNeeded = (value: any) => {
+  return value === true || value === false ? String(value) : value
```
TejasQ:
```suggestion
  // We compare booleans like this and without typeof for perf
  return value === true || value === false ? String(value) : value
```

src/DataTable/README.md
```diff
@@ -22,7 +29,7 @@ import { DataTable, DataTableSelect, DataTableInput, Checkbox } from "@operation
     [String(Math.random()).repeat(1000), "Imogen Mason", "Good Stuff", true],
     [String(Math.random()).repeat(1000), "Fabien Bernard", "🥖🥐🧀🍷", false],
     [String(Math.random()).repeat(1000), "STEREO BOOSTER", "☕️", true],
-    [String(Math.random()).repeat(1000), "Mischa Potomin", "null", false],
+    [String(Math.random()).repeat(1000), <CustomCell>{"Mischa Potomin"}</CustomCell>, "null", false],
```
TejasQ:
```suggestion
    [String(Math.random()).repeat(1000), <CustomCell>Mischa Potomin</CustomCell>, "null", false],
```

PR #1163, opened 2019-07-26
**Fix:** for TypeError: Cannot read property 'length' of null

stereobooster:
# Summary

The problem is that data in Row is any, and this any flows through throw the whole code of DataTable and TypeScript will not help. Can we restrict DataTable to `string | number | null | boolean`?

# Related issue

# To be tested

Me

Tester 1

Tester 2


TejasQ:
This is a nice “halfway” fix. Opened an issue to fix the root of the issue.

PR #1168, opened 2019-07-30
**Feature:** Tree a11y

### LINKED ISSUE #1085: Tree is completely inaccessible
Tabbing around [here](https://deploy-preview-1083--operational-ui.netlify.com/#/Components/Tree) doesn't allow us to do anything with the tree (expand or collapse nodes) via keyboard.

This is _bad_.

Let's fix it.

### LINKED ISSUE #1166: Tree focus
A sibling to #1085 – a [tree item](https://github.com/contiamo/operational-ui/blob/master/src/Tree/TreeItem.tsx) would look nicer if it had a blue focused state like so:

image

### LINKED ISSUE #1167: Listen for Double click on Tree Items
Sometimes, you'd want to be able to double-click on [tree items](https://github.com/contiamo/operational-ui/blob/master/src/Tree/TreeItem.tsx) and have something happen. Let's implement this!

TejasQ:
# Summary
This PR:
- fixes #1167 
- fixes #1166 
- fixes #1085

We:
- add a focus state to tree
- support `onDoubleClick`
- make it keyboard accessible

# Related issue

# To be tested

Me

Tester 1

- [ ] Tree works as described in this PR and in the README.
  

Tester 2

- [ ] Tree works as described in this PR and in the README.

### DIFF
--- a/src/Tree/ChildTree.tsx
+++ b/src/Tree/ChildTree.tsx
@@ -34,18 +34,8 @@ const ChildTree: React.SFC<Props> = ({
 }) => {
   const [isOpen, setIsOpen] = React.useState(Boolean(initiallyOpen))
   const hasChildren = Boolean(childNodes && childNodes.length)
-  const onNodeClick =
-    !disabled && (hasChildren || onClick)
-      ? (e: React.MouseEvent<HTMLDivElement>) => {
-          e.stopPropagation()
-          if (hasChildren) {
-            setIsOpen(!isOpen)
-          }
-          if (onClick) {
-            onClick()
-          }
-        }
-      : undefined
+  const clickCount = React.useRef(0)
+  const timeoutForDoubleClick = 200
 
   const onNodeContextMenu = React.useMemo(
     () =>
@@ -57,6 +47,32 @@ const ChildTree: React.SFC<Props> = ({
         : undefined,
     [disabled, onContextMenu],
   )
+
+  const onNodeClick =
+    !disabled && (hasChildren || onClick)
+      ? (e: React.MouseEvent<HTMLDivElement>) => {
+          e.stopPropagation()
+          if (e.altKey && onNodeContextMenu) {
+            onNodeContextMenu(e)
+            return
+          }
+          clickCount.current++
+          setTimeout(
+            () => {
+              if (clickCount.current === 1) {
+                if (hasChildren) {
+                  setIsOpen(!isOpen)
+                }
+                if (onClick) {
+                  onClick()
+                }
+              }
+              clickCount.current = 0
+            },
+            onDoubleClick ? timeoutForDoubleClick : 0,
+          )
+        }
+      : undefined
 
   const onNodeDoubleClick = React.useMemo(
     () =>
--- a/src/Tree/ChildTree.tsx
+++ b/src/Tree/ChildTree.tsx
@@ -25,6 +25,7 @@ const ChildTree: React.SFC<Props> = ({
   droppableProps,
   onClick,
   onContextMenu,
+  onDoubleClick,
   onRemove,
   cursor,
   searchWords,
@@ -57,6 +58,17 @@ const ChildTree: React.SFC<Props> = ({
     [disabled, onContextMenu],
   )
 
+  const onNodeDoubleClick = React.useMemo(
+    () =>
+      !disabled && onDoubleClick
+        ? (event: React.MouseEvent<HTMLDivElement, MouseEvent>) => {
+            event.preventDefault()
+            onDoubleClick(event)
+          }
+        : undefined,
+    [disabled, onContextMenu],
+  )
+
   return (
     <Container ref={forwardRef} disabled={Boolean(disabled)} hasChildren={hasChildren} {...props}>
       <TreeItem
@@ -64,6 +76,7 @@ const ChildTree: React.SFC<Props> = ({
         searchWords={searchWords}
         onNodeClick={onNodeClick}
         onNodeContextMenu={onNodeContextMenu}
+        onNodeDoubleClick={onNodeDoubleClick}
         highlight={Boolean(highlight)}
         hasChildren={hasChildren}
         isOpen={isOpen}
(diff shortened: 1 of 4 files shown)


src/Tree/README.md
```diff
@@ -1,6 +1,22 @@
...
+- <kbd>Tab</kbd> navigates to the next node
+- <kbd>Shift</kbd>+<kbd>Tab</kbd> navigates to the previous node
+- <kbd>Space</kbd> triggers the `onClick` handler, usually expanding a node
+- <kbd>Enter</kbd> triggers the `onDblClick` action
+- <kbd>Alt</kbd>+<kbd>Enter</kbd> triggers the right-click action
+- <kbd>Delete</kbd> or <kbd>Backspace</kbd> triggers the "remove" action, if available
+
+Try it yourself!
+
+- "Region" has a double click/<kbd>Enter</kbd> handler
```
mpotomin: `Region` reacts to single "Enter" keystroke with the double-click handler for me. Hence, I also can not close/open that part of the tree with my keyboard (single clicking with mouse still works as expected though)

src/Tree/ChildTree.tsx
```diff
@@ -33,25 +34,52 @@ const ChildTree: React.SFC<Props> = ({
 }) => {
   const [isOpen, setIsOpen] = React.useState(Boolean(initiallyOpen))
   const hasChildren = Boolean(childNodes && childNodes.length)
+  const clickCount = React.useRef(0)
+  const timeoutForDoubleClick = 200
```
stereobooster: Is there a reason why we need to reimplement doubleclick instead of using built in [`onDoubleClick`](https://reactjs.org/docs/events.html#mouse-events)?

src/Tree/README.md
TejasQ: Space should handle expand and collapse, while enter triggers the “action” on double click. This is outlined in the README.md. 😉

src/Tree/ChildTree.tsx
TejasQ: We’re using both, but our requirements are to only expand/collapse on single click, while doing something else on double click. The single click action **should not happen** on double click. 

This is the only way that I’ve seen so far to accomplish this. 

Definitely open to suggestions to solve this in an alternate way.

src/Tree/README.md
TejasQ: Perhaps I could reword things to make them a little clearer. Or do you have an idea to make the behavior more like what you expect?

mpotomin: I have another question. The item "Legal Entity" right below the "Region" reacts to "Enter" key with opening/closing. It feels like a pretty inconsistent behaviour to me. I would expect to use `Space` key only everywhere for opening/closing the tree items, especially since there is no visual sing or marker, telling me, that "Region" has a handler for a double-click/enter attached. What's your take on this @TejasQ?

mpotomin: Apart from the click/space/enter inconsistency, the rest looks good to me, I would leave it as it is 👍

src/Tree/ChildTree.tsx
stereobooster: > The single click action should not happen on double click.

Makes a lot of sense. Maybe add comment about this, or even create separate component which provides this behaviour.

src/Tree/README.md
TejasQ: I think we could leave that up to the users. I'm also not that convinced about the behavior. It is currently:

- if it's foldable, space toggles, enter double-clicks
  - if no double-click specified, enter falls back to click

I totally understand what you're talking about but I figure we'll have an issue about it from a more concrete use case in the future if its truly confusing. I think it's a little hard to say at this stage.

src/Tree/ChildTree.tsx
TejasQ: [You got it bro](https://github.com/contiamo/operational-ui/pull/1168/commits/67c0c42d5b5be6c2c19ebf4b0c6e6502df9bfbfe).

stereobooster:
Sometimes I get following behaviour on double click:

- it collapse the tree
- then alert
- after alert it uncollapses the tree

Maybe a race condition or something

src/Tree/ChildTree.tsx
```diff
@@ -33,25 +34,60 @@ const ChildTree: React.SFC<Props> = ({
...
-            setIsOpen(!isOpen)
-          }
-          if (onClick) {
-            onClick()
+          if (e.altKey && onNodeContextMenu) {
+            onNodeContextMenu(e)
+            return
           }
+          clickCount.current++
+          setTimeout(
```
stereobooster: Do we want to cancel previous timeout before launching new one?

TejasQ:
@stereobooster awesome. I think it's maybe that the timeout is not long enough. 😢 Unfortunately if I make it longer, the UX seems choppy. I don't really know how to solve it. I'll make an issue.

src/Tree/ChildTree.tsx
```diff
@@ -33,25 +34,60 @@ const ChildTree: React.SFC<Props> = ({
...
+            () => {
+              if (clickCount.current === 1) {
+                if (hasChildren) {
+                  setIsOpen(!isOpen)
```
stereobooster: This theoretically can be triggered after component unmount, which will trigger the error

src/Tree/TreeItem.tsx
```diff
@@ -75,51 +82,95 @@ const TreeItem: React.SFC<TreeItemProps> = ({
...
+        case "Delete":
+        case "Backspace":
+          e.preventDefault()
+          if (onRemove) {
+            onRemove(e)
+          }
+          return
+      }
+    },
+    [onNodeContextMenu, onNodeDoubleClick, onNodeClick],
```
stereobooster:
```suggestion
    [onNodeContextMenu, onNodeDoubleClick, onNodeClick, onRemove],
```

we need linter for this

fabien0102:
Why unit-tests are not played on this PR 🤔 Is it new?

fabien0102:
image

The double click action fold the tree section for me (and expanded when I close the alert)

stereobooster:
Side question: how do you trigger context menu with keyboard?

TejasQ:
@stereobooster please [read the PR](https://github.com/contiamo/operational-ui/pull/1168/files#diff-da0c8b02385f203827104fc313ccb636R11).

TejasQ:
So there's no clear direction on how to handle use cases with both **double click** _and_ **right click** behavior. I will close this and focus on other stuff until it makes sense to spend time here again.

micha-f:
@TejasQ We'll make figuring this out a priority. I'm sure some of this PR will live on 👍 .

PR #1169, opened 2019-07-30
**Breaking:** Update ContextMenu

TejasQ:
# Summary
This PR:

- updates the look of ContextMenu
- adds a `separator` property to `IContextMenuItem` that allows `"top"`, `"bottom"`, or `"both"`, allowing users to compose separators according to their requirements.
- breaks the API removing the possibility to have `string` as menu items
  - we don't really use this and it was only left in the codebase to preserve backwards compatibility, but since we're "breaking" the design with a new look, we can break it properly. 😋 

# To be tested

Me

Tester 1

- [x] No regression on `ActionMenu`
- [x] No regression on `ContextMenu`
- [x] No regression on `DropdownButton`
- [x] No regression on `Select`
- [x] No regression on `TopbarSelect`
- [x] No regression on `DataTableSelect`

  

Tester 2

- [ ] No regression on `ActionMenu`
- [ ] No regression on `DropdownButton`
- [ ] No regression on `Select`
- [ ] No regression on `TopbarSelect`
- [ ] No regression on `DataTableSelect`


mpotomin:
@TejasQ  when navigating the context menu with arrows up and down, the separator the selection cursor dwells on a separator for an iteration instead of jumping to the next available item. Is it an expected behaviour? 

00-ContextMenu

fabien0102:
I don't find any screens, everything looks bold but I can check if it's wanted or not 😕

fabien0102:
image

But I don't think we should apply bold here for example, this looks messy

fabien0102:
image

The grey is very light, I'm really afraid about contrast here 😕

TejasQ:
@mpotomin, great catch! I'll look into skipping the separator.

@fabien0102 – this is the design specification from @kemal-contiamo and @micha-f . I propose we merge these and continue iterating – especially the concern about contrast. I discussed this with @micha-f and it's the same as Google Drive. I don't really think this is a problem because the contrast between the _foreground text_ and the _background color_ is still favorable. Also, people will know what is hovered/focused because it is screenreader accessible. Brief, there's no real drawback to the contrast the way it is.

Further, There is no real spec for `DropdownButton` or even use case for it now if I understand correctly since we removed the button from the product. Maybe we should remove it from Operational?

TejasQ:
@mpotomin updated the keyboard navigability.

micha-f:
@TejasQ Looks good.

We still need the dropdown button (albeit with slightly different styling) and it should show a dropdown that is or is styled like a context menu.

kemal-contiamo:
> image
> 
> But I don't think we should apply bold here for example, this looks messy

It shouldn't apply bold this much indeed @fabien0102 , as the screen for this one looks like this originally: 
Screenshot 2019-07-30 at 21 33 20

micha-f:
@kemal-contiamo But we are not using that screen anymore?

micha-f:
@kemal-contiamo @TejasQ could this not be the new style of the DropdownButton?

image

kemal-contiamo:
> @kemal-contiamo But we are not using the screens you posted anymore?

@micha-f yes that is the previous screens indeed. But, last week in product JF, we agreed to use the existing dropdown style for at least until the demo right? I will create a new story to update all the dropdown button style to match the new ones after the demo. At least as I understood it, this review is for tomorrow the changes to be used in the demo.

TejasQ:
Let's deal with the DropdownButton separately – this PR is about ContextMenu. The bold text in the DropdownButton example is a concern of userland – it's code from the README to illustrate that you can put _whatever you want_ inside a ContextMenu's label – including ugly bold children.

For this reason, I don't think [this comment](https://github.com/contiamo/operational-ui/pull/1169#issuecomment-516483920) is relevant to this PR.

mpotomin:
There is a regression in the `Autocomplete` - the new one opens empty in the example

Screenshot 2019-07-31 at 12 00 07
Screenshot 2019-07-31 at 12 00 12

TejasQ:
That's awesome, @mpotomin! Thanks for your help!

fabien0102:
image

PR #1170, opened 2019-07-30
**Breaking:** Update Sidenav

### LINKED ISSUE #1119: Make sidenav narrower
Let's update the [Sidenav](https://operational-ui.netlify.com/#!/Sidenav) in compact mode to:

- make it thinner – 64px
- remove labels, use tooltips instead

This will look better I think.

TejasQ:
# Summary
This PR fixes #1119:

- makes `Sidenav` narrower
- moves labels into tooltips

## Breaking Changes
- `Sidenav` looks different
- `Tooltip` no longer supports arbitrary positioning props but instead relies on a single `position` prop.
- `Tooltip` no longer supports `"smart"` option and expects users to know ahead of time where to place the tooltip, or calculate it in userland.

image

# To be tested

Me

Tester 1

- [ ] No regression on `Sidenav`
  

Tester 2

- [ ] No regression on `Sidenav`


src/SidenavItem/Tooltip.tsx
```diff
@@ -0,0 +1,50 @@
...
+    transform: translateY(-50%);
+    display: block;
+    width: 0;
+    height: 0;
+    border: 6px solid transparent;
+    border-right-color: ${({ theme, dark }) => (dark ? theme.color.primaryDark : theme.color.white)};
+  }
+`
+
+const SidenavTooltip: React.FC<{ dark: boolean }> = ({ children, ...props }) => {
```
mpotomin: Why are we creating a new Tooltip here instead of applying the [one we already have?](https://deploy-preview-1170--operational-ui.netlify.com/#/Components/Tooltip) Would extending the functionality of the existing `<Tooltip/>` make sense?

fabien0102:
Probably a question for our designer, but we can't see the tooltip when we have a submenu ->

image

TejasQ:
Good point. @kemal-contiamo what should we do?

kemal-contiamo:
> Good point. @kemal-contiamo what should we do?

@fabien0102 @TejasQ any chance we put the tooltip in even higher layer than the sub-menu? so in theory, we should still see the tooltip referring to that particular item.

TejasQ:
@kemal-contiamo that doesn't really make any sense because it would overlay the submenu. What if we just add a "title" to the submenus that are equivalent to the tooltips? 

So "Fabien" [here](https://user-images.githubusercontent.com/1761469/62444729-9e2bf300-b75e-11e9-888a-9153202ddd3e.png) would be the title: "Settings" or somthing?

kemal-contiamo:
> @kemal-contiamo that doesn't really make any sense because it would overlay the submenu. What if we just add a "title" to the submenus that are equivalent to the tooltips?
> 
> So "Fabien" [here](https://user-images.githubusercontent.com/1761469/62444729-9e2bf300-b75e-11e9-888a-9153202ddd3e.png) would be the title: "Settings" or somthing?

@TejasQ better not do it like that, because its not consistent style with the other ones then. Perhaps we can place the tooltip either to the bottom or top (only in the case of menu having sub-menu that is)?

TejasQ:
I don't understand.

kemal-contiamo:
> I don't understand.

@TejasQ what I meant is we can do it like these examples: 
Screenshot 2019-08-05 at 15 50 55
Screenshot 2019-08-05 at 15 51 13
Screenshot 2019-08-05 at 15 51 23

As long as the arrow of the tooltip is pointing towards the intended nav, it does the job in terms of UX in my opinion.

TejasQ:
Respectfully, I think that is ugly and confusing. As a user if I see that, I will:

- think the product is broken
  - negatively judge the developer and designer responsible for that
    - feel bad about it
    - have an existential crisis
      - _am i a bad person_
        - train myself to assume better intentions
          - try to be a nice person

I would like to suggest we iterate in a different direction here. I apologize because I cannot yet offer an alternative look apart from what I've already recommended, but I certainly think a tooltip on top of a popup is a huge no-no.

This is my best idea (so far) in that it answers the question "what am I hovering over" by placing a heading there with the answer, but I'm sure there's more.

image

micha-f:
@fabien0102 @kemal-contiamo @TejasQ I'm glad we're discussing this!

I also think a tooltip on top of a submenu looks broken. I would like to explore two options:

1) A sub menu title - which appears not clickable (that's the challenge I guess)

2) No tooltip on a sidenav item that opens a submenu on hover and no submenu title either. The tooltip is meant to tell the user what you will see after a click on the item. In the case of a submenu that opens on hover that is not necessary - you already see the content of the submenu on hover.

kemal-contiamo:
> @fabien0102 @kemal-contiamo @TejasQ I'm glad we're discussing this!
> 
> I also think a tooltip on top of a submenu looks broken. I would like to explore two options:
> 
> 1. A sub menu title - which appears not clickable (that's the challenge I guess)
> 2. No tooltip on a sidenav item that opens a submenu on hover and no submenu title either. The tooltip is meant to tell the user what you will see after a click on the item. In the case of a submenu that opens on hover that is not necessary - you already see the content of the submenu on hover.

@micha-f @TejasQ I would vote for second approach that Michael mentioned, which is not to show the tooltip at all for the menu item that has sub-menu. Not the most perfect solution, but in my view, it is not making the style inconsistent and also won't look broken.
In my view, the point of having the tooltip is more like secondary helping tool to identify what the icon on the menu item representing. So for item that has sub-menu, I don't think further explanation won't be necessary the moment they clicked on that item, as the sub-menu would reveal what menu item that is. So yeah +1 with what Michael already said basically.

src/SidenavItem/Tooltip.tsx
```diff
@@ -0,0 +1,50 @@
...
+  border-radius: 2px;
+  align-items: center;
+  justify-content: center;
+  padding: ${({ theme }) => theme.space.small}px;
+  background: ${({ theme, dark }) => (dark ? theme.color.primaryDark : theme.color.white)};
+  box-shadow: ${({ theme }) => theme.shadows.popup};
+  color: ${({ theme, dark }) => (dark ? theme.color.white : theme.color.text.default)};
+  z-index: ${({ theme }) => theme.zIndex.tooltip};
+
+  ::before {
```
stereobooster: this is arrow, right? Maybe put a comment

src/SidenavItem/Tooltip.tsx
```diff
@@ -0,0 +1,50 @@
...
+`
+
+const SidenavTooltip: React.FC<{ dark: boolean }> = ({ children, ...props }) => {
+  const [rect, setRect] = React.useState<ClientRect | null>(null)
+  const $container = React.useRef<HTMLDivElement>(null)
+
+  React.useLayoutEffect(() => {
+    const currentContainer = $container.current
+    if (currentContainer) {
+      setRect(currentContainer.getBoundingClientRect())
```
stereobooster: I wonder if `currentContainer.getBoundingClientRect()` returns new instance every time, so this can be performance issue

stereobooster: Just checked it, seems to be ok

mpotomin:
Screenshot 2019-08-13 at 16 55 03

The difference is very subtle!

mpotomin:
:/

mpotomin:
It fixes after I enforce the re-render though, so i think it's fine?!

TejasQ:
I'll fix it.

PR #1175, opened 2019-08-06
 **Fix:** DataTable sticky header bug

TejasQ:
# Summary

Before:
before

Now:
after

# To be tested

Me

Tester 1

Tester 2

### DIFF
--- a/src/Checkbox/Checkbox.tsx
+++ b/src/Checkbox/Checkbox.tsx
@@ -38,6 +38,8 @@ const toggleCheckboxAnimation = keyframes`
 const Input = styled("input")<{ condensed: CheckboxProps["condensed"] }>`
   position: absolute;
   left: -100vw;
+  opacity: 0;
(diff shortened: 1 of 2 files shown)


fabien0102:
Was broken before, but since we are on it:

image


And I can't scroll horizontally on any example (so it's hard to validate your fix)

src/Checkbox/Checkbox.tsx
```diff
@@ -38,6 +38,8 @@ const toggleCheckboxAnimation = keyframes`
 const Input = styled("input")<{ condensed: CheckboxProps["condensed"] }>`
   position: absolute;
   left: -100vw;
+  opacity: 0;
+  visibility: hidden;
```
stereobooster: make sure it's hidden

stereobooster:
Screenshot 2019-08-12 at 16 37 55

some text sneaked in

PR #1185, opened 2019-08-09
**Feature:** Update DataTable condensed

TejasQ:
We want condensed DataTable to be a bit smaller.

image


micha-f:
Just one note - it says "update data table condensed". But my understanding is that it updates the font size for all data tables (not just condensed). Correct? That's fine! But the title doesn't communicate it.

stereobooster:
rebase?

PR #1189, opened 2019-08-12
**Fix:** Remove sidenav border in dark mode.

ImogenF:
# Summary

https://contiamo.atlassian.net/browse/PLAT-811

Grey border on the right of the sidebar is now only there if the sidenav has a white background - in dark mode, it is gone.

image

image

# Related issue

# To be tested

Me

Tester 1

Tester 2

### DIFF
--- a/src/Sidenav/Sidenav.tsx
+++ b/src/Sidenav/Sidenav.tsx
@@ -28,12 +28,10 @@ const Container = styled("div")<SidenavProps>(({ theme, compact, dark }) => {
     overflow: "auto",
     width: compact ? theme.compactSidebarWidth : theme.sidebarWidth,
     height: "100%",
-    ...(dark
-      ? {}
-      : {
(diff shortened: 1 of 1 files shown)


src/Sidenav/Sidenav.tsx
```diff
@@ -28,8 +28,12 @@ const Container = styled("div")<SidenavProps>(({ theme, compact, dark }) => {
     overflow: "auto",
     width: compact ? theme.compactSidebarWidth : theme.sidebarWidth,
     height: "100%",
-    borderRight: "1px solid",
-    borderRightColor: theme.color.separators.default,
+    ...(dark
+      ? {}
```
TejasQ: I think a logical and operator makes more sense than a one-sided ternary in this case. 🤔

TejasQ:
image

Is this correct?

PR #1192, opened 2019-08-13
**Fix:** Update DropdownButton.

ImogenF:
# Summary

DropdownButton updated to conform to specs. 

# Related issue

# To be tested

Me

Tester 1

Tester 2

### DIFF
--- a/src/Button/Button.tsx
+++ b/src/Button/Button.tsx
@@ -39,7 +39,7 @@ export interface ButtonProps extends DefaultProps {
   children?: React.ReactNode
 }
 
-const makeColors = (theme: OperationalStyleConstants, color: string) => {
+export const makeColors = (theme: OperationalStyleConstants, color: string) => {
   const defaultColor = theme.color.white
 
   const backgroundColors: { [key: string]: string } = {
@@ -81,7 +81,7 @@ const BaseButton = styled<"button" | "a">("button")<{
   const { background: backgroundColor, foreground: foregroundColor } = makeColors(theme, color_ || "")
   return {
     backgroundColor,
-    lineHeight: `${condensed ? 28 : 36}px`,
+    lineHeight: `${condensed ? 26 : 34}px`,
     minWidth: "max-content",
     fontSize: theme.font.size.small,
     fontFamily: theme.font.family.main,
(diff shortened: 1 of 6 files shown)


TejasQ:
Odd. When it opens, it pushes everything around it. This'll move things around quite a bit in the workbench or explore.

blo

ImogenF:
Oh, my bad. Good catch.

src/Button/Button.tsx
```diff
@@ -52,12 +52,21 @@ export const makeColors = (theme: OperationalStyleConstants, color: string) => {
     warning: theme.color.error,
   }
 
+  const borderColors: { [key: string]: string } = {
+    ghost: setAlpha(0)(theme.color.ghost),
```
TejasQ: Isn't this `transparent`?

ImogenF: 🤦‍♀ Yes. I thought `transparent` was causing some weird behaviour but it turns out that was something else.

src/ContextMenu/ContextMenu.tsx
```diff
@@ -71,7 +71,7 @@ const MenuContainer = styled.div<{
   isOpen: boolean
 }>(({ theme, numRows, align, embedChildrenInMenu, isOpen }) => ({
   position: "absolute",
-  top: embedChildrenInMenu ? 0 : "100%",
+  top: embedChildrenInMenu ? 0 : "calc(100% - 1px)",
```
TejasQ: why? Maybe leave a comment in code explaining this?

src/ContextMenu/ContextMenu.tsx
```diff
@@ -71,7 +71,8 @@ const MenuContainer = styled.div<{
   isOpen: boolean
 }>(({ theme, numRows, align, embedChildrenInMenu, isOpen }) => ({
   position: "absolute",
-  top: embedChildrenInMenu ? 0 : "100%",
+  // Context menu is moved up by one pixel to cover the bottom border of dropdown buttons.
```
TejasQ: Sorry but this hack breaks a lot of the existing library. Here's one example:

image


Everything that uses the ContextMenu (Select, ActionButton, etc.) appears to be affected.

src/DropdownButton/DropdownButton.tsx
```diff
@@ -27,16 +26,25 @@ export interface DropdownButtonProps extends DefaultProps {
   align?: "left" | "right"
 }
 
-const BaseDropdownButton = styled(Button)<{ isOpen: boolean; textColor: string }>(({ isOpen, textColor, theme }) => ({
+const BaseDropdownButton = styled(Button)<{ isOpen: boolean }>(({ isOpen, theme }) => ({
   paddingRight: 0,
-  ...(isOpen && { boxShadow: `0 0 0 1px ${expandColor(theme, textColor)} inset` }),
+  marginRight: 0,
+  ...(isOpen && {
+    borderBottom: "1px solid #fff",
```
TejasQ: This would come back to bite us in the end when we want to implement the dark mode unless we change it like so:

```suggestion
    borderBottom: `1px solid ${({theme}) => theme.color.white}`,
```

PR #1200, opened 2019-08-14
**Fix:** Style of DataTable

TejasQ:
# Summary
This PR updates the grey on DataTable and allows the options to open.

# Related issue
https://contiamo.atlassian.net/browse/CON-484

# To be tested

Me

Tester 1

- [x] No regression on `DataTable`.
  

Tester 2

- [ ] No regression on `DataTable`.


stereobooster:
```
main.js:936 [Deprecation] Percentages row tracks and gutters for indefinite height grid containers will be resolved against the intrinsic height instead of being treated as auto and zero respectively. This change will happen in M70, around October 2018. See https://www.chromestatus.com/feature/6708326821789696 for more details.

main.js:871 Attempting to truncate something that is not a JS primitive: this will fail for functions, objects and arrays. Please reconsider the item you are truncating: {$$typeof: Symbol(react.element), type: {…}, key: null, ref: null, props: {…}, …}
```

PR #1209, opened 2019-08-19
**Fix:** Small fixes for accordion and tabs

stereobooster:
# Summary

- Add hover for plus
- Made icons in accordion unfocusable
- Fix warning in the console

# Related issue

# To be tested

Me

Tester 1

Tester 2

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,6 +16,7 @@
   "scripts": {
     "start": "ts-node -P tsconfig.node.json ./scripts/start.ts",
     "build": "run-s build:*",
+    "clean:icons": "ts-node -P tsconfig.node.json ./scripts/build-icons.ts -c ./src/Icon",
     "build:icons": "ts-node -P tsconfig.node.json ./scripts/build-icons.ts -o ./src/Icon",
     "build:package": "rimraf lib .tsbuildinfo && tsc -d",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -16,6 +16,7 @@
   "scripts": {
     "start": "ts-node -P tsconfig.node.json ./scripts/start.ts",
     "build": "run-s build:*",
+    "clean:icons": "ts-node -P tsconfig.node.json ./scripts/build-icons.ts -c ./src/Icon",
```
fabien0102: I don't see how this can works, you missed the `-o` option, personnally I just do `yarn build:icons -c` when I need a clean build 😉

fabien0102:
Can you also add some other littles fixes to the tabs here?

The specs for this:
> Let’s improve the hover style of the “X” for closing a tab. It should always be the same color (whether tab is active or inactive): #BFCBD2. The circle around the X should be 24px not 30px (it’s too big right now).

package.json
stereobooster: I had no idea

fabien0102:
https://app.zeplin.io/project/5cc31fad6f17562d6dac9a53/screen/5d42f85de37a238930a05444

PR #1218, opened 2019-08-22
**Feature**: improve modal sizes

stereobooster:
# Summary

Added fullSize property for Modal, added Close icon, added border radius, still need to change padding, but I want to do it in separate PR

# Related issue

# To be tested

Me

Tester 1

Tester 2

### DIFF
--- a/src/Modal/Modal.tsx
+++ b/src/Modal/Modal.tsx
@@ -39,14 +39,6 @@ const Modal: React.RefForwardingComponent<HTMLDivElement, ModalProps> = (
     console.warn("width, height and anchor are ignored when used together with fullSize")
   }
 
-  if (
-    process.env.NODE_ENV !== "development" &&
-    fullSize &&
-    (height !== undefined || width !== undefined || anchor !== undefined)
-  ) {
-    console.warn("width, height and anchor are ignored when used together with fullSize")
-  }
-
   // Focus the modal on open
   React.useEffect(() => {
     const currentModalContainer = $modalContainer.current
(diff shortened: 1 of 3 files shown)


src/Modal/Modal.tsx
```diff
@@ -41,20 +55,32 @@ const Modal: React.RefForwardingComponent<HTMLDivElement, ModalProps> = (
...
+  }, [isOpen, anchor, fullSize])
+
+  // Resize
+  React.useEffect(() => {
+    if (isOpen && fullSize) {
+      const handleResize = () => setSize([0, 0, window.innerWidth - margin, window.innerHeight])
+      window.addEventListener("resize", handleResize)
+      return () => window.removeEventListener("resize", handleResize)
+    }
+  }, [isOpen, fullSize])
```
fabien0102:
```suggestion
  }, [isOpen, fullSize, setSize])
```

(I know it's the same, but I would like to introduce the eslint hook rule later, so if we can avoid to much works over there ^^)

src/Modal/Modal.tsx
```diff
@@ -1,34 +1,48 @@
...
   ref,
 ) => {
   const $modalContainer = React.useRef<HTMLDivElement>(null)
-  const [size, setSize] = React.useState<[Top, Left, Width, Height]>([50, 0, 480, height])
+  const [size, setSize] = React.useState<[Top, Left, Width, Height]>(
+    fullSize ? [0, 0, window.innerWidth - margin, window.innerHeight] : [50, 0, 480, height],
+  )
+
+  if (
+    process.env.NODE_ENV !== "development" &&
```
fabien0102: Only warn if we are not in dev? 🤔

fabien0102:
image

the corner is not rounded!

PR #1219, opened 2019-08-22
**Fix:** strange bug with state on operational ui

stereobooster:
# Summary

When we use state to track focus it breaks tree and dropdown button in some cases, we wasn't able to find why, instead we decided to manipulate HTML directly. This is hacky we will investigate the issue later

# Related issue

N/A

# To be tested

Me

Tester 1

Tester 2


fabien0102:
@stereobooster snapshots…

PR #1223, opened 2019-08-27
**Fix:** the max-height of the modal

fabien0102:
# Summary

image

The max-height was not correct, so the action button can be outside of the screen.

# Related issue

# To be tested

Me

Tester 1

Tester 2


mpotomin:
Let's add a ticket to remove the anchored mode for all Modals since they look broken. Particularly this change with heights looks ok to me

mpotomin:
Accepting since the anchored mode is obsolete and will be addressed with https://contiamo.atlassian.net/browse/UI-129

PR #1235, opened 2019-09-18
**Breaking:** Update DataTable truncation

TejasQ:
# Summary
This PR implements this spec for `DataTable`:
image

Of course, implementing the spec was not as straightforward as it seemed. 🤦‍♂ It required a lot more rewiring and updating of related components in the following ways:

## `DataTableHeader`

`DataTable`'s column headers were always `ReactNode`s – meaning they could be `string`s, or other React components. For this reason, we create a new `DataTableHeader` that can go into a column header that supports:

- icons
- text with truncation
- "view more" functionality
- a context menu with actions

## `useSticky`

There has been a recurring need in the product for **true absolute positioning** since Operational UI began: we used to have tooltips that get cut off by a sidenav overflow, various modal issues, and most recently, _super weird behavior_ with `position: sticky;` on the `DataTable` headers. For this reason, I introduced a `useSticky` hook. The behavior was already implemented in [SidenavItem/Popout.tsx](https://github.com/contiamo/operational-ui/blob/a5ddf74251c90be4427c48e9177c16da5d991f55/src/SidenavItem/Popout.tsx#L27-L50) and I started writing the same code before I decided to turn it into a hook.

This hook takes any component. Its order of execution is the following:
- component mounts with initial position (`absolute`, `relative`, whatever)
- get the rect of this component
- set its position to `fixed` and update all coordinates (`top`, `left`, etc.)
- use a [portal](https://reactjs.org/docs/portals.html) and render it outside of our current DOM tree, just above `</body>`

The above flow ensures that `position: fixed` elements are **always on top of things and in the right place**. This was painful to get right, but I think from this point, positioning things will become a whole lot easier.

## `useViewMore`
We have "view more" functionality in two places: 
- `DataTable` cells
- `DataTableHeader`

We grouped the following behavior into a hook for reuse:
- get mouse coordinates
- display popup at coordinates
- display on a side with enough space always
- click outside to close
- do not be longer than the screen

# Breaking Changes
Since the truncation of strings is now dynamic (related to the width of the cell) instead of static (via a magic number), [the `maxCharactersInCell` prop](https://github.com/contiamo/operational-ui/blob/a5ddf74251c90be4427c48e9177c16da5d991f55/src/DataTable/DataTable.tsx#L50) is removed.

# Known issues
The "down arrow" icon is not 100% according to the spec. @nathanedavis1 will provide the right SVG soon and we can solve this in another PR.

# To be tested

Me
- [x] No regression on `DataTable`
- [x] No regression on `ContextMenu`
- [x] No regression on `Autocomplete`
- [x] No regression on `Select`
- [x] No regression on `DropdownButton`
- [x] No regression on `TopbarSelect`
- [x] No regression on `SidenavItem`

Tester 1

- [x] No regression on `DataTable`
- [x] No regression on `ContextMenu`
- [x] No regression on `Autocomplete`
- [x] No regression on `Select`
- [x] No regression on `DropdownButton`
- [x] No regression on `TopbarSelect`
- [x] No regression on `SidenavItem`
  

Tester 2

- [ ] No regression on `DataTable`
- [ ] No regression on `Select`
- [ ] No regression on `DropdownButton`
- [ ] No regression on `TopbarSelect`
- [ ] No regression on `SidenavItem`

### DIFF
--- a/src/ContextMenu/ContextMenu.Popout.tsx
+++ b/src/ContextMenu/ContextMenu.Popout.tsx
@@ -1,0 +1,80 @@
+import * as React from "react"
+import { createPortal } from "react-dom"
+
+import styled from "../utils/styled"
+import { ContextMenuProps } from "./ContextMenu"
+import useSticky from "../useSticky/useSticky"
+
+export interface ContextMenuPopoutProps {
+  embedChildrenInMenu?: ContextMenuProps["embedChildrenInMenu"]
+  numRows: number
+  align: ContextMenuProps["align"]
+  condensed: ContextMenuProps["condensed"]
+  rowHeight: number
+  container?: React.RefObject<HTMLDivElement>
+  children?: React.ReactNode
+}
+
+interface PositionProps {
+  left: string
+  top: string
+  position: string
+  width?: string
+}
+
+const Container = styled.div<ContextMenuPopoutProps & PositionProps>`
+  position: ${({ position }) => position};
+  top: ${({ top }) => top};
+  left: ${({ left }) => left};
+  max-height: 50vh;
+  overflow: auto;
+  box-shadow: ${({ theme }) => theme.shadows.contextMenu};
+  min-width: fit-content;
+  width: ${({ width }) => width};
+  max-width: 90vw;
+  min-height: ${({ rowHeight }) => rowHeight}px;
+  display: grid;
+  grid-template-rows: repeat(${({ numRows }) => numRows}, max-content);
+  background-color: ${({ theme }) => theme.color.white};
+  padding: ${({ theme }) => theme.space.small}px 0;
+  z-index: ${({ theme }) => theme.zIndex.selectOptions + 2};
+`
+
+const ContextMenuPopout = React.forwardRef(
+  (
+    { align, children, condensed, container, embedChildrenInMenu, numRows, rowHeight }: ContextMenuPopoutProps,
+    forwardRef: React.Ref<HTMLDivElement>,
+  ) => {
+    const fallbackRef = React.useRef<HTMLDivElement | null>(null)
+    const ref = forwardRef || fallbackRef
+
+    const { left, position, top, width } = useSticky(ref, {
+      position: "absolute",
+      left: align === "left" ? "0" : "auto",
+      top: embedChildrenInMenu ? "0" : "100%",
+      width: "100%",
+      alignment: "flex-start",
+    })
+
+    const areWeFixedYet = position === "fixed"
+    const popOut = (
+      <Container
+        align={align}
+        condensed={condensed}
+        numRows={numRows}
+        rowHeight={rowHeight}
+        left={left}
+        width={width}
+        top={top}
+        position={position}
+        ref={ref}
+      >
+        {children}
+      </Container>
+    )
+
+    return areWeFixedYet && container && container.current ? createPortal(popOut, container.current) : popOut
+  },
+)
+
+export default ContextMenuPopout
--- a/src/ContextMenu/ContextMenu.Popout.tsx
+++ b/src/ContextMenu/ContextMenu.Popout.tsx
@@ -12,6 +12,7 @@ export interface ContextMenuPopoutProps {
   align: ContextMenuProps["align"]
   condensed: ContextMenuProps["condensed"]
   rowHeight: number
+  anchored: boolean
   container?: React.RefObject<HTMLDivElement>
   children?: React.ReactNode
 }
@@ -38,22 +39,25 @@ const Container = styled.div<ContextMenuPopoutProps & PositionProps>`
   grid-template-rows: repeat(${({ numRows }) => numRows}, max-content);
   background-color: ${({ theme }) => theme.color.white};
   padding: ${({ theme }) => theme.space.small}px 0;
-  z-index: ${({ theme }) => theme.zIndex.selectOptions + 2};
+
+  ${({ theme, anchored }) => (anchored ? "" : `z-index: ${theme.zIndex.selectOptions + 2};`)}
 `
 
 const ContextMenuPopout = React.forwardRef<HTMLDivElement, ContextMenuPopoutProps>(
-  ({ align, children, condensed, container, embedChildrenInMenu, numRows, rowHeight }, forwardRef) => {
+  ({ align, children, condensed, container, embedChildrenInMenu, numRows, rowHeight, anchored }, forwardRef) => {
     const $fallback = React.useRef<HTMLDivElement | null>(null)
     const $el = isRefRefObject(forwardRef) ? forwardRef : $fallback
 
+    const initialValue = {
+      position: "absolute",
+      left: align === "left" ? "0" : "auto",
+      top: embedChildrenInMenu ? "0" : "100%",
+      width: "100%",
+    } as const
+
     const { left, position, top, width } = useSticky({
       $el: $el,
-      initialValue: {
-        position: "absolute",
-        left: align === "left" ? "0" : "auto",
-        top: embedChildrenInMenu ? "0" : "100%",
-        width: "100%",
-      },
+      initialValue,
     })
 
     const areWeFixedYet = position === "fixed"
@@ -64,17 +68,21 @@ const ContextMenuPopout = React.forwardRef<HTMLDivElement, ContextMenuPopoutProp
         condensed={condensed}
         numRows={numRows}
         rowHeight={rowHeight}
-        left={left}
-        width={width}
-        top={top}
-        position={position}
+        anchored={anchored}
+        left={anchored ? initialValue.left : left}
+        width={anchored ? initialValue.width : width}
+        top={anchored ? initialValue.top : top}
+        position={anchored ? initialValue.position : position}
         ref={$el}
+        role="listbox"
       >
         {children}
       </Container>
     )
 
-    return areWeFixedYet && container && container.current ? createPortal(popOut, container.current) : popOut
+    return !anchored && areWeFixedYet && container && container.current
+      ? createPortal(popOut, container.current)
+      : popOut
   },
 )
 
--- a/src/ContextMenu/ContextMenu.Popout.tsx
+++ b/src/ContextMenu/ContextMenu.Popout.tsx
@@ -48,12 +48,16 @@ const ContextMenuPopout = React.forwardRef(
     const fallbackRef = React.useRef<HTMLDivElement | null>(null)
     const ref = forwardRef || fallbackRef
 
-    const { left, position, top, width } = useSticky(ref, {
-      position: "absolute",
-      left: align === "left" ? "0" : "auto",
-      top: embedChildrenInMenu ? "0" : "100%",
-      width: "100%",
-      alignment: "flex-start",
+    const { left, position, top, width } = useSticky({
+      inputRef: ref,
+      options: { shouldAvoidToggler: true },
+      initialValue: {
+        position: "absolute",
+        left: align === "left" ? "0" : "auto",
+        top: embedChildrenInMenu ? "0" : "100%",
+        width: "100%",
+        alignment: "flex-start",
+      },
     })
 
     const areWeFixedYet = position === "fixed"
--- a/src/ContextMenu/ContextMenu.Popout.tsx
+++ b/src/ContextMenu/ContextMenu.Popout.tsx
@@ -4,6 +4,7 @@ import { createPortal } from "react-dom"
 import styled from "../utils/styled"
 import { ContextMenuProps } from "./ContextMenu"
 import useSticky from "../useSticky/useSticky"
+import { isRefRefObject } from "../utils/isRefRefObject"
 
 export interface ContextMenuPopoutProps {
   embedChildrenInMenu?: ContextMenuProps["embedChildrenInMenu"]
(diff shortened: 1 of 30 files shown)


src/DataTable/CellContent.tsx
```diff
@@ -0,0 +1,52 @@
...
+  cell: React.ReactNode
+  open: (content: string) => (e: React.MouseEvent<Element, MouseEvent>) => void
+}
+
+const stringifyIfNeeded = (value: any) => {
+  // We compare booleans like this and without typeof for perf
+  return value === true || value === false ? String(value) : value
+}
+
+const CellContent: React.FC<CellContentProps> = ({ cell, open }: any) => {
```
mpotomin: Do we need this `any` here?

src/DataTable/DataTable.styled.ts
```diff
@@ -77,23 +91,12 @@ export const DataWrapper = styled("div")<{ numHeaders: number; rowHeight: DataTa
   top: ${({ numHeaders, rowHeight }) => numHeaders * getHeaderRowHeight(rowHeight)}px;
 `
 
-export const ViewMoreToggle = styled("div", { shouldForwardProp: prop => prop !== "height" })<{ height: number }>`
-  position: absolute;
-  right: 0;
-  top: 0;
-  height: ${({ height }) => height}px;
-  width: 40px;
+export const ViewMoreToggle = styled("div")`
```
mpotomin:
```suggestion
export const ViewMoreToggle = styled.div`
```

src/DataTable/DataTable.styled.ts
```diff
@@ -43,12 +43,26 @@ export const Cell = styled.div<{
...
   color: ${({ theme }) => theme.color.text.default};
   grid-column: ${({ cell }) => cell};
   background-color: ${({ theme }) => theme.color.white};
 `
 
+export const CellGrid = styled.div<{ canTruncate: boolean }>`
+  display: ${({ canTruncate }) => (canTruncate ? "grid" : "flex")};
+  align-items: center;
+  width: 100%;
+  ${({ canTruncate }) => (canTruncate ? "grid-template-columns: calc(100% - 36px) 36px" : "")};
```
mpotomin: Could we use a variable / sum of variables from the `theme` or at least move out this 36px outside this line and explain why it is exactly 36?

src/DataTableHeader/DataTableHeader.styled.ts
```diff
@@ -0,0 +1,39 @@
+import styled from "../utils/styled"
+
+export const Container = styled.div<{ hasIcon: boolean }>`
+  display: grid;
+
+  /**
+   * The columns are: type, name, "see more" icon, and caret for icons
+   */
+  grid-template-columns: ${({ hasIcon }) => (hasIcon ? `36px auto 36px 36px` : `auto 36px 36px`)};
```
mpotomin: Another 36px, looks like it is really asking for an own exported variable within the DataTable context

src/useSticky/useSticky.ts
```diff
@@ -0,0 +1,77 @@
...
+ *
+ * @param inputRef - a ref to the component we would like to stick
+ * @param initialValue - initial positioning CSS values
+ */
+const useSticky = ({
+  inputRef,
+  options,
+  initialValue,
+}: {
+  inputRef: React.Ref<HTMLDivElement>
```
mpotomin: Would it make sense to use `<HTMLElement>` here instead of `<HTMLDivElement>`? i.e. if I would like to use this hook for sticking around non-styled element or a non-div?

src/DataTable/CellContent.tsx
TejasQ: UFFF I MISSED THIS SO BAD SORRY

src/DataTable/DataTable.styled.ts
TejasQ: love you

src/DataTable/DataTable.styled.ts
TejasQ: 100% sure

src/DataTableHeader/DataTableHeader.styled.ts
TejasQ: got you

src/useSticky/useSticky.ts
TejasQ: the truth is that you're right and i love you

mpotomin:
Looks like a bug to me, watch the first row of numbers:

There is no `...` in the row since initially the number fits the cell, then the number gets truncated when i resize, but there is no `...` appearing to let me view the non-truncated value

src/ContextMenu/ContextMenu.tsx
```diff
@@ -138,6 +119,20 @@ const ContextMenu: React.FC<ContextMenuProps> = ({
     initiallyOpen: open,
   })
 
+  React.useEffect(() => {
```
stereobooster:
```ts
React.useEffect(() => {
  if (isOpen && setIsOpen) {
    const hideOnScroll = () => setIsOpen(false)
    document.addEventListener("scroll", hideOnScroll)
    return () => {
      document.removeEventListener("scroll", hideOnScroll)
    }
  }
}, [isOpen, setIsOpen])
```

src/useSticky/useSticky.ts
```diff
@@ -0,0 +1,77 @@
...
+    left: initialValue ? initialValue.left : "100%",
+    width: initialValue ? initialValue.width : "100%",
+    position: initialValue ? initialValue.position : "absolute",
+    alignment: initialValue ? initialValue.alignment : "flex-start",
+  }
+
+  const [displaySettings, setDisplaySettings] = React.useState(defaultDisplaySettings)
+
+  React.useLayoutEffect(() => {
+    const node = inputRef && (inputRef as React.RefObject<HTMLDivElement>).current // ts can't figure out it's a refObject and thinks it's a function ref.
```
stereobooster: 👃 it should be able to figure out type, because you declared it `inputRef: React.Ref<HTMLDivElement>`. Also `inputRef` declared as non-optional why we need to check `inputRef &&...`

src/useSticky/useSticky.ts
```diff
@@ -0,0 +1,77 @@
...
+ * @param initialValue - initial positioning CSS values
+ */
+const useSticky = ({
+  inputRef,
+  options,
+  initialValue,
+}: {
+  inputRef: React.Ref<HTMLDivElement>
+  options?: {
+    shouldAvoidToggler: boolean
```
stereobooster: What is `Toggler` and why we should avoid it?

src/DataTable/CellContent.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+  return (
+    <CellGrid ref={$cell} canTruncate={isString(cell)}>
+      {isString(cell) ? <CellTruncator>{stringifyIfNeeded(cell)}</CellTruncator> : stringifyIfNeeded(cell)}
+      {isString(cell) && isTextProbablyOverflowing ? (
+        <ViewMoreToggle onClick={open(cell)}>
+          <DotMenuHorizontalIcon
+            color="color.text.lighter"
+            size={20}
+            onClick={noop} // for the hover/focus effect
```
stereobooster: We need to remove those hacks of repurposing onClick handler as styling flag

src/DataTable/CellContent.tsx
```diff
@@ -0,0 +1,52 @@
...
+          />
+        </ViewMoreToggle>
+      ) : (
+        <div />
```
stereobooster:
```suggestion
        null
```
?

src/useSticky/useSticky.ts
```diff
@@ -0,0 +1,77 @@
...
+        draftSettings.alignment = "flex-end"
+      } else {
+        draftSettings.top = `${rect.top}px`
+        draftSettings.alignment = "flex-start"
+      }
+
+      draftSettings.position = "fixed"
+      setDisplaySettings({ ...displaySettings, ...draftSettings })
+    }
+  }, [inputRef])
```
stereobooster:
```suggestion
  }, [inputRef, get(options, "shouldAvoidToggler", false)])
```

?

src/ContextMenu/ContextMenu.tsx
TejasQ: More context would be great here.

src/useSticky/useSticky.ts
TejasQ: > 👃 it should be able to figure out type, because you declared it inputRef: React.Ref<HTMLDivElement>

It cannot because `React.Ref<HTMLDivElement>` is a union between `((instance: HTMLElement | null) => void)` and `React.RefObject<HTMLElement>` and `undefined`. So TypeScript cannot figure out that it is a `RefObject` instead of a function.

We need the emptiness check for the same reason: `React.Ref<HTMLDivElement>` is effectively defined like so:

```ts
  type React.Ref<HTMLDivElement> = ((instance: HTMLElement | null) => void) | React.RefObject<HTMLElement> | undefined
```

Does that clear things up?

src/DataTable/CellContent.tsx
TejasQ: Shall we solve this in a separate issue?

cc @fabien0102

src/DataTable/CellContent.tsx
TejasQ: No. We need a `div` here so that CSS' grid algorithm places the next element in the right place. Shall I add a comment?

src/useSticky/useSticky.ts
TejasQ: I can just watch all the options.

src/useSticky/useSticky.ts
TejasQ: This is a really good question! I accidentally introduced conceptual coupling here – `useSticky` really shouldn't know about any `ContextMenu` or `Sidenav` concepts like a toggler.

Will refine the implementation a bit. Great catch!

src/ContextMenu/ContextMenu.Popout.tsx
```diff
@@ -0,0 +1,84 @@
...
+  padding: ${({ theme }) => theme.space.small}px 0;
+  z-index: ${({ theme }) => theme.zIndex.selectOptions + 2};
+`
+
+const ContextMenuPopout = React.forwardRef(
+  (
+    { align, children, condensed, container, embedChildrenInMenu, numRows, rowHeight }: ContextMenuPopoutProps,
+    forwardRef: React.Ref<HTMLDivElement>,
+  ) => {
+    const fallbackRef = React.useRef<HTMLDivElement | null>(null)
```
fabien0102:
```suggestion
    const fallbackRef = React.useRef<HTMLDivElement>(null)
```

This should be fine for typescript

src/DataTable/CellContent.tsx
```diff
@@ -0,0 +1,68 @@
...
+  open: (content: string) => (e: React.MouseEvent<Element, MouseEvent>) => void
+}
+
+const stringifyIfNeeded = (value: any) => {
+  // We compare booleans like this and without typeof for perf
+  return value === true || value === false ? String(value) : value
+}
+
+const CellContent: React.FC<CellContentProps> = ({ cell, open }) => {
+  const $cell = React.useRef<HTMLDivElement>(null)
```
fabien0102: We have two conventions in this project:
 - `$myvar`
 - `myvarRef`

I quickly search accross our project, `myvarRef` won, can you convert the few badboys? 😇

src/utils/isRefRefObject.ts
```diff
@@ -0,0 +1,3 @@
+export function isRefRefObject<T>(ref: React.Ref<T>): ref is React.RefObject<T> {
```
stereobooster: Why RefRef?

TejasQ: It checks if `React.Ref` is a `RefObject`.

So "is a ref a ref object?" is the name of the function, but shorter.

TejasQ: More info here: 

https://github.com/contiamo/operational-ui/pull/1235#discussion_r326075597

src/DataTableHeader/DataTableHeader.tsx
```diff
@@ -0,0 +1,86 @@
+import * as React from "react"
+import { createPortal } from "react-dom"
+import noop from "lodash/noop"
+
+import { ActionsContainer, Container, TitleContainer, GhostTitleContainer } from "./DataTableHeader.styled"
+import { IconComponentType, DotMenuHorizontalIcon, ChevronDownIcon } from "../Icon"
+import { IContextMenuItem } from "../ContextMenu/ContextMenu.Item"
+import isString = require("lodash/isString")
```
stereobooster:
```suggestion
import isString from "lodash/isString"
```

src/SidenavItem/Popout.tsx
```diff
@@ -24,33 +26,29 @@ const ScrollTrap = styled("div")`
...
+  const { height } = useWindowSize()
 
   React.useLayoutEffect(() => {
-    const node = containerNode.current
+    const node = $container.current
     if (node) {
       const rect = node.getBoundingClientRect()
-      setLeft(`${rect.left}px`)
-
+      console.log(node, rect.top, height)
```
stereobooster:
```suggestion
```

stereobooster:
Screenshot 2019-09-19 at 16 43 41

Why scroll is so big?

TejasQ:
> Why scroll is so big?

More context would help

stereobooster:
Screenshot 2019-09-19 at 16 48 28

stereobooster:
Focus Autocomplete field, type "a", scroll outside of field, type something, dropdown doesn't appear anymore

mpotomin:
> > Why scroll is so big?
> 
> More context would help

Screenshot 2019-09-19 at 16 53 07

stereobooster:
`Select`
Screenshot 2019-09-19 at 16 57 47

TejasQ:
> Focus Autocomplete field, type "a", scroll outside of field, type something, dropdown doesn't appear anymore

@stereobooster is this a regression or does it exist like this on `master`?

TejasQ:
> SidenavItem (in the master it is the same)

let's fix that in a separate issue in this case

PR #1236, opened 2019-09-23
**Fix:** Update truncation behavior

TejasQ:
# Summary
All truncated popups must display on `hover`.

This PR implements this.

# To be tested

Me

Tester 1

- [ ] Click on `Tree`'s truncated text (4th example from the top) triggers popover
- [ ] Hover on `Tree`'s truncated text (4th example from the top) does not trigger popover

### DIFF
--- a/src/DataTable/useViewMore.ts
+++ b/src/DataTable/useViewMore.ts
@@ -22,7 +22,7 @@ const useViewMore = () => {
   )
 
   const close = () => {
-    closeTimeout = window.setTimeout(() => setViewMorePopup(false), 300)
+    closeTimeoutId = window.setTimeout(() => setViewMorePopup(false), 150)
   }
 
   React.useEffect(
--- a/src/DataTable/useViewMore.ts
+++ b/src/DataTable/useViewMore.ts
@@ -39,6 +39,7 @@ const useViewMore = () => {
     viewMorePopup,
     toggle: (content: string) => (viewMorePopup ? setViewMorePopup(false) : openViewMore(content)),
     open: openViewMore,
+    close: () => setViewMorePopup(false),
   }
 }
 
--- a/src/DataTable/useViewMore.ts
+++ b/src/DataTable/useViewMore.ts
@@ -7,26 +7,6 @@ import * as React from "react"
 const useViewMore = () => {
   const [viewMorePopup, setViewMorePopup] = React.useState<{ content: string; x: number; y: number } | false>(false)
 
-  React.useEffect(() => {
-    if (!viewMorePopup) {
-      return
-    }
-
-    const close = () => {
-      setViewMorePopup(false)
-    }
-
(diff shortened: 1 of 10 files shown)


TejasQ:
lol

mpotomin:
It's red 🔴 

Screenshot 2019-09-23 at 14 24 06

mpotomin:
@TejasQ Just a though, not necessarily should be implemented now: actually, we could copy the value to the clipboard when we click three horizontal dot's, still showing the non-truncated value on hover

src/DataTable/useViewMore.ts
```diff
@@ -5,39 +5,40 @@ import * as React from "react"
  * triggered by a click event.
  */
 const useViewMore = () => {
+  let closeTimeout = 0
```
mpotomin: First I thought it's a timeout duration, but then figured from the code below that it is actually a close timeout Id. Shall we rename it to avoid confusion?

```suggestion
  let closeTimeoutId = 0
```

src/DataTable/useViewMore.ts
```diff
@@ -5,39 +5,40 @@ import * as React from "react"
...
+        content,
+        x: e.clientX > window.innerWidth / 2 ? e.clientX - 8 : e.clientX + 8,
+        y: e.clientY > window.innerHeight / 2 ? e.clientY - 8 : e.clientY + 8,
+      })
+    },
+    [viewMorePopup, closeTimeout],
+  )
+
+  const close = () => {
+    closeTimeout = window.setTimeout(() => setViewMorePopup(false), 300)
```
mpotomin: From testing in the PR preview, 300 feels pretty slow and gives me the feeling that we have some performance issues. Also made try if I still could reach the pop-over to select and copy the value from it. Was left frustrated  since having mouse over the pop-over does not keep the pop-over opened.

If there is no requirement to keep it 300, I'd go with smaller value to make it feel snappier. Would 150ms also work here?

```suggestion
    closeTimeout = window.setTimeout(() => setViewMorePopup(false), 150)
```

src/DataTable/useViewMore.ts
```diff
@@ -5,39 +5,40 @@ import * as React from "react"
...
+  }
+
+  React.useEffect(
+    () => () => {
+      if (closeTimeout) {
+        window.clearTimeout(closeTimeout)
```
mpotomin: Shouldn't we also set `closeTimeout` to `undefined` here so next time this check runs it will return false?

TejasQ: I don't understand? `0` is also false... This check only runs on unmount. 🤔 

Could you maybe say more about what the concern is?

mpotomin: I think in this case I forgot, that `0` gets coerced to `false` in this check. Makes more sense now

PR #1239, opened 2019-09-25
**Fix:** Title component width

TejasQ:
# Summary
#1238 broke `Title`, leading to odd alignment issues. We want `Title` to have a full-width on `Page`, but nowhere else. This PR corrects this regression.

More concretely, we fix this:
regression

# To be tested

Me
- [ ] No new error or warning in the console on `localhost:6060`

Tester 1

Tester 2


TejasQ:
Great catch! Turns out you found a much bigger issue. Fixed. It's supposed to have a smaller font-size than `master`. This is a _progression_.

mpotomin:
> Great catch! Turns out you found a much bigger issue. Fixed. It's supposed to have a smaller font-size than `master`. This is a _progression_.

PRO-gression! 💯

PR #1242, opened 2019-09-26
**Fix:** Select in Modal

TejasQ:
# Summary
This PR fixes #1240.

# Related issue
#1240

# To be tested

Me
- [x] No error or warning in the console on `localhost:6060`

Tester 1

- [ ] Things look good on the demo.
  

Tester 2

- [ ] Things look good on the demo.

PR #1245, opened 2019-10-10
**Fix:** Anchor Autocomplete AND FIX CARDSECTION

TejasQ:
The same fix as #1242 but for Autocomplete. 🙄

### DIFF
--- a/src/Autocomplete/Autocomplete.tsx
+++ b/src/Autocomplete/Autocomplete.tsx
@@ -91,6 +91,7 @@ export function Autocomplete<T>({
 
   return (
     <Container
+      anchored
       open
       iconLocation="right"
       fullWidth={Boolean(fullWidth)}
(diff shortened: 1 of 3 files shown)


mpotomin:
Code looks good :). Waiting for the build deploy

TejasQ:
@mpotomin the reason that occurs is because the right panel of our documentation isn't high enough. We can fix it (the documentation) by:

- adding more examples, or
- extending the height of the content area to `100vh` (which it currently isn't, seen below)
image

The `Autocomplete` itself isn't broken: it has a `max-height` of `50vh` and will never overflow the screen. Unfortunately in case of our docs, it overflows the docs because they're not high enough. 

Does that make sense?

TejasQ:
sneaky? me? never 😛 

🔥 🤘 🤠 🤘 🔥

PR #1248, opened 2019-10-18
**Feature:** Add reorder functionality to Table

TejasQ:
# Summary
I added reorder to the table so that I can reorder stuff by dragging.

# To be tested

Me

Tester 1
- [ ] I can drag and drop to reorder in the example


mpotomin:
@TejasQ The demo looks good and works ass expected, but it's not keyboard accessible at all. Is it expected, do we plan to add later?

PR #1253, opened 2019-10-29
**Feature:** Add syntax to support left & right icon in table headers

fabien0102:
# Summary

Add `icon: {left: MyIcon, right: MyOtherIcon}` syntax for the `DataTableHeader` to be able to have multiples icons.

This is needed to be able to implement this:

image

This feature is totally retro-compatible, so nothing should break 😁 

I also remove the strange behaviour around the truncation, now the hover is on the text itself

image

This feel more natural and let the space for the custom icon.

# Related issue

# To be tested

Me

Tester 1

- [x] I can have an icon on right
- [x] I can have an icon on left
- [x] I can have two icons!
- [x] The truncate hover is beautiful

### DIFF
--- a/src/DataTable/CellContent.tsx
+++ b/src/DataTable/CellContent.tsx
@@ -1,9 +1,7 @@
 import * as React from "react"
 import isString from "lodash/isString"
-import noop from "lodash/noop"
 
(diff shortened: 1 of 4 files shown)


fabien0102:
I know what I forgot… `onClick` on icon…

PR #1255, opened 2019-11-02
**Fix:** Select component border

### LINKED ISSUE #1249: <Select /> looks broken
From the big green example [in the docs](http://operational-ui.netlify.com/#!/Modal/13):

image

Let's fix this.

miqh:
Hello,

This small fix attempts to address a visual issue with the select component appearing to have a missing border on one side.

The `box-shadow` property being used to style the borders doesn't count as part of the element dimensions. Consequently, depending on what the select component is used inside, different sides of the border can be clipped.

The fix is simply to use an `inset` based `box-shadow`.

This should also resolve a one pixel offset issue with the select component when used in conjunction with other elements. This is a zoomed in example to illustrate:

image

Closes #1249

### DIFF
--- a/src/Select/Select.styled.ts
+++ b/src/Select/Select.styled.ts
@@ -23,7 +23,8 @@ export const Combobox = styled("div")<{ naked: boolean; isOpen: boolean; hasCust
   align-items: stretch;
   box-shadow: 0 0 0 1px
     ${({ theme, isOpen, hasCustomOption }) =>
-      !hasCustomOption && isOpen ? theme.color.primary : theme.color.border.select};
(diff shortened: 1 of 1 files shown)


TejasQ:
Wow! Thanks for your great contribution, @miqh! We'll review this quickly and let you know! The code looks good though!

stereobooster:
Not an issue of your code, but corners of background are now visible

Screenshot 2019-11-04 at 12 11 28

Inset shadow feels a bit thiner than border (1px border ~ 1.2px inset shadow at least on my screen)

Screenshot 2019-11-04 at 12 13 47

But I think this is still better than before

TejasQ:
@miqh wins today! Thanks for your lovely contribution! 🎉

PR #1258, opened 2019-11-13
**Fix:** Fix datatable actions placements

fabien0102:
# Summary

Before:
image

image

After:
image

# Related issue

# To be tested

Me

Tester 1

Tester 2

### DIFF
--- a/src/DataTableHeader/DataTableHeader.styled.ts
+++ b/src/DataTableHeader/DataTableHeader.styled.ts
@@ -40,13 +40,4 @@ export const ActionsContainer = styled.div`
   margin: auto;
   width: 100%;
   height: 100%;
-
-  /* The ContextMenu */
-  div {
-    display: flex;
-    align-items: center;
-    justify-content: center;
-    width: 100%;
-    height: 100%;
-  }
 `
--- a/src/useSticky/useSticky.ts
+++ b/src/useSticky/useSticky.ts
@@ -1,4 +1,5 @@
 import * as React from "react"
+import constants from "../utils/constants"
 
 /**
  *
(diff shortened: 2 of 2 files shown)


src/DataTableHeader/DataTableHeader.styled.ts
```diff
@@ -40,13 +40,4 @@ export const ActionsContainer = styled.div`
   margin: auto;
   width: 100%;
   height: 100%;
-
-  /* The ContextMenu */
```
fabien0102: @TejasQ I'm not really sure why we introduce this peace of code, so let's check the preview with care 😅

src/useSticky/useSticky.ts
```diff
@@ -40,7 +40,22 @@ const useSticky = ({
 
     if (node) {
       const rect = node.getBoundingClientRect()
-      setDisplaySettings({ position: "fixed", left: `${rect.left}px`, width: `${rect.width}px`, top: `${rect.top}px` })
+      const needOffset = rect.left + rect.width > window.innerWidth
+      if (needOffset) {
+        setDisplaySettings({
+          position: "fixed",
+          left: `${window.innerWidth - rect.width - 8}px`,
```
mpotomin: Can we use https://github.com/contiamo/operational-ui/blob/af46bdcb42d72577bfcd5e1c925333c0ba4af3fc/src/utils/constants.ts#L198 instead?

fabien0102: Sure

mpotomin:
@fabien0102 Would it make sense to have a clickable test case to avoid regression in the future?

mpotomin:
> @fabien0102 Would it make sense to have a clickable test case to avoid regression in the future?

found one in the DataTable!

PR #1259, opened 2019-11-14
**Fix:** Remove all focus rings if mouse is being used

TejasQ:
# Summary
This PR uses CSS' `:focus-visible` selector to hide/show focus styles based on input modality ([further reading](https://twitter.com/TejasKumar_/status/1186326187364229120)).

**Left half:** this PR
**Right half:** `master`

wow

If we're using a mouse, we'll never see a focus ring on anything.
If we're tabbing around with a keyboard, we will.

# To be tested

Me

Tester 1

Tester 2


TejasQ:
My GIF was from Chrome.

PR #1260, opened 2019-11-14
**Feature:** Add support for "active" table rows

TejasQ:
# Summary
This PR creates the possibility for this type of behavior from `Table`:
image

I added a demo to the README, so it should be visible in the PR preview.

# To be tested

Me

Tester 1

Tester 2


TejasQ:
It's the first thing that came to my mind and serves the purpose. Please assume good intentions as opposed to relying on negative assumptions. Better for a team.

PR #1262, opened 2019-11-25
**Fix:** Make useURLState update on URL change

TejasQ:
# Summary
This PR implements "two way data binding" with useURLState:

1) It updates when the URL changes,
2) It updates on `setValue`.here -->

# To be tested

Me

Tester 1

Tester 2


fabien0102:
I tried some setups locally but nothing really successful/clean, the conclusion of this search was to use react-router to handle this kind of state than need to be updated from url with any component re-render. Let's close this and keep this component as is, simple!

PR #1263, opened 2019-11-26
  **Feature:** Exposing FixedProgress component

mpotomin:
# Summary
This PR co-locates and exports  the `<FixedProgress />` component to make it available for use not only within the current `<Page />`, but also externally and in custom Page implementations 

# Related issue
No issue

# To be tested

Me

Tester 1

Tester 2


fabien0102:
Let's maybe add a little example about this in the doc, not sure about the usage from your PR ;)

mpotomin:
> Let's maybe add a little example about this in the doc, not sure about the usage from your PR ;)

Good point! Was just thinking about! Makes sense for testing and making sure everything works

PR #1265, opened 2019-11-26
**Fix:** Give AvatarGroup tooltips

### LINKED ISSUE #1264: AvatarGroup needs to show a full name
Currently, AvatarGroup only shows initials of its Avatars and makes a user **guess** what the initials mean.

Let's show a tooltip on hover to display a user's name.

TejasQ:
# Summary
This PR fixes #1264 and allows users to know which avatars mean what.

# Related issue
#1264

# To be tested

Me

Tester 1

Tester 2


micha-f:
@TejasQ Would it be possible to list additional names in a tooltip of the last "+7" circle?

Maybe using a simple rule: Up to 10 lines in that tooltip, then "and more..."

PR #1266, opened 2019-11-29
**Improvement:** Improve table row focus, clickable and draggable style.

micha-f:
Slightly adjusting the table row's focus style as we have a lot of clickable table rows and the browser's default focus is not that pretty.


fabien0102:
Let's improve this PR by using the real focus manager introduced there -> https://github.com/contiamo/operational-ui/pull/1214

So we don't have any non-wanted focus state on click

PR #1268, opened 2019-12-02
**Fix**: Fallback on the first tab if not exists

fabien0102:
# Summary

On advanced usecase, when the number of tabs change dynamically, the active tab can be undefined and crash the application.

This little fix make sure to fallback correctly.

# Related issue

# To be tested

Me

Tester 1

Tester 2

### DIFF
--- a/src/Internals/Tabs.tsx
+++ b/src/Internals/Tabs.tsx
@@ -190,6 +190,10 @@ const Tabs = ({ onTabChange, tabs, activeTabName, children }: Props) => {
     [activeTab, onTabClick],
   )
 
+  if (tabsWithId.length === 0 || tabs.length === 0) {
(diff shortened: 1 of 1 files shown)


src/Internals/Tabs.tsx
```diff
@@ -212,8 +212,8 @@ const Tabs = ({ onTabChange, tabs, activeTabName, children }: Props) => {
               ))}
           </TabsBar>
         ),
-        activeTabId: tabsWithId[activeTab].id,
-        activeChildren: tabs[activeTab].children,
+        activeTabId: (tabsWithId[activeTab] || tabsWithId[0]).id,
```
mpotomin: Could it be that we have no tabs at all?

PR #1271, opened 2019-12-11
  **Fix:** Fix hanging popups

stereobooster:
# Summary

`DotMenuHorizontalIcon` shows up and hide based on `useLayoutEffect`, so it can be there even if it is not needed

Screenshot 2019-12-11 at 11 32 24

when you hover it, it will show popup, but the same time `useLayoutEffect` kicks in and `DotMenuHorizontalIcon` disappears, so there is no more dom element to handle `onMouseLeave={close}` and popups left hanging there forever. 

# Related issue

# To be tested

Me

Tester 1

Tester 2

### DIFF
--- a/src/DataTable/useViewMore.ts
+++ b/src/DataTable/useViewMore.ts
@@ -5,28 +5,27 @@ import * as React from "react"
  * triggered by a click event.
  */
 const useViewMore = () => {
-  const [closeTimeoutId, setCloseTimeoutId] = React.useState<number>()
+  const closeTimeoutIdRef = React.useRef<number>()
   const [viewMorePopup, setViewMorePopup] = React.useState<{ content: string; x: number; y: number } | false>(false)
 
   const close = React.useCallback(() => {
     if (viewMorePopup) {
-      setCloseTimeoutId(window.setTimeout(() => setViewMorePopup(false), 150))
+      closeTimeoutIdRef.current = window.setTimeout(() => setViewMorePopup(false), 150)
     }
   }, [viewMorePopup])
-  React.useEffect(() => () => window.clearTimeout(closeTimeoutId), [closeTimeoutId])
+  React.useEffect(() => () => window.clearTimeout(closeTimeoutIdRef.current), [])
 
   const open = React.useCallback(
     (content: string) => (e: React.MouseEvent) => {
       e.stopPropagation()
-      window.clearTimeout(closeTimeoutId)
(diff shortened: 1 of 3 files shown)


src/DataTable/useViewMore.ts
```diff
@@ -5,39 +5,36 @@ import * as React from "react"
  * triggered by a click event.
  */
 const useViewMore = () => {
-  let closeTimeoutId = 0
+  const [closeTimeoutId, setCloseTimeoutId] = React.useState<number>()
```
fabien0102: State will cause an useless re-render, we can optimize this

src/DataTable/useViewMore.ts
```diff
@@ -5,39 +5,36 @@ import * as React from "react"
...
-  React.useEffect(
-    () => () => {
-      if (closeTimeoutId) {
-        window.clearTimeout(closeTimeoutId)
-      }
-    },
     [closeTimeoutId],
   )
 
+  const toggle = React.useCallback((content: string) => (viewMorePopup ? close() : open(content)), [viewMorePopup])
```
fabien0102: we should put `close` and `open` as deps here

src/DataTable/useViewMore.ts
```diff
@@ -5,39 +5,36 @@ import * as React from "react"
...
  */
 const useViewMore = () => {
-  let closeTimeoutId = 0
+  const [closeTimeoutId, setCloseTimeoutId] = React.useState<number>()
   const [viewMorePopup, setViewMorePopup] = React.useState<{ content: string; x: number; y: number } | false>(false)
 
-  const openViewMore = React.useCallback(
+  const close = React.useCallback(() => {
+    if (viewMorePopup) {
+      setCloseTimeoutId(window.setTimeout(() => setViewMorePopup(false), 150))
```
fabien0102: I don't really see the point of this timeout, @TejasQ you may have more context ^^

src/DataTable/useViewMore.ts
stereobooster: they never change though... but I can fix this

src/DataTable/useViewMore.ts
TejasQ: The timeout fixes this issue: https://github.com/contiamo/operational-ui/pull/1236#pullrequestreview-291771889

fabien0102:
image

Not totally convince 🙄

PR #1272, opened 2019-12-11
**Feature:** Search input

fabien0102:
# Summary

New component to support our fancy search feature, with categories please! 

This component should also be accessible with the proper aria helpers. ("should" because I'm still beginning in this area, so it's probably not perfect)

c1c77fb2-2ca3-4e8b-bb8e-91337ee4492f

# Related issue

# To be tested

Me

Tester 1

- [ ] The component is accessible (keyboard)


micha-f:
@fabien0102 Looks great! Just one idea: when the user clicks on the select dropdown, I would actually expect it to close after selecting a value, not jump into the search input text box. In our use case, it is common for people to not enter a search string at all but use the select to filter filter the result types.

icons/Enter.svg
```diff
@@ -0,0 +1,15 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 13 8.92" height="30" width="30">
+  <defs>
+    <style>
```
stereobooster: I guess we can remove style and id and other props/tags

src/SearchInput/README.md
```diff
@@ -0,0 +1,58 @@
+Specialized input for advanced search ux.
```
stereobooster:
```suggestion
Specialized input for advanced search UX.
```

src/SearchInput/SearchInput.tsx
```diff
@@ -0,0 +1,274 @@
...
+  const toggleOpen = React.useCallback(
+    (event: React.MouseEvent<HTMLDivElement, MouseEvent>) => {
+      event.stopPropagation()
+      setIsOpen(i => !i)
+    },
+    [setIsOpen],
+  )
+
+  React.useEffect(() => {
+    if (isOpen && setIsOpen) {
```
stereobooster:
```suggestion
    if (isOpen) {
```

src/SearchInput/SearchInput.tsx
```diff
@@ -0,0 +1,274 @@
...
+      const hideOnScroll = () => setIsOpen(false)
+      document.addEventListener("scroll", hideOnScroll)
+      return () => {
+        document.removeEventListener("scroll", hideOnScroll)
+      }
+    }
+  }, [isOpen, setIsOpen])
```
stereobooster:
```suggestion
  }, [isOpen])
```

src/SearchInput/SearchInput.tsx
```diff
@@ -0,0 +1,274 @@
...
+  useHotkey(containerRef, { key: "ArrowDown" }, () => {
+    if (!isOpen) {
+      setIsOpen(true)
+    }
+    setActiveItemIndex(prev => (prev + 1) % (props.categories || []).length)
+  })
+
+  useHotkey(containerRef, { key: "ArrowUp" }, () =>
+    setActiveItemIndex(prev =>
+      prev === 0 ? (props.categories || []).length : (prev - 1) % (props.categories || []).length,
```
mpotomin: `|| : [] % . : % || [] `

my eyes hurt 😢 

Can we change this fancy one-liner into multi-liner function that is easier to read?

src/SearchInput/SearchInput.tsx
```diff
@@ -0,0 +1,274 @@
...
+      return
+    }
+    props.onChange({ search: props.value, category: props.categories[activeItemIndex] })
+    setIsOpen(false)
+  })
+
+  useHotkey(containerRef, { key: "Escape" }, () => setIsOpen(false))
+
+  if (props.categories && !props.categories.includes(props.category!)) {
+    throw new Error("[SearchInput] `categories` and `category` props doesn't match!")
```
mpotomin:
```suggestion
    throw new Error("[SearchInput] `categories` and `category` props don't match!")
```

src/SearchInput/SearchInput.tsx
```diff
@@ -0,0 +1,274 @@
...
+  const containerRef = React.useRef<HTMLDivElement>(null)
+
+  useHotkey(containerRef, { key: "ArrowDown" }, () => {
+    if (!isOpen) {
+      setIsOpen(true)
+    }
+    setActiveItemIndex(prev => (prev + 1) % (props.categories || []).length)
+  })
+
+  useHotkey(containerRef, { key: "ArrowUp" }, () =>
```
stereobooster:
```suggestion
  useHotkey(containerRef, { key: "ArrowUp" }, () => {
    setIsOpen(true);
```

src/SearchInput/SearchInput.tsx
```diff
@@ -0,0 +1,274 @@
...
+        />
+      )}
+    </Container>
+  )
+}
+
+const Container = styled.div<{ hasCategory: boolean; isOpen: boolean }>`
+  display: grid;
+  cursor: text;
+  grid-template-columns: ${props => (props.hasCategory ? "16px min-content auto" : "16px auto")};
```
mpotomin: 16px = constants.space.content

src/SearchInput/SearchInput.tsx
```diff
@@ -0,0 +1,274 @@
...
+    }
+  }, [isOpen, setIsOpen])
+
+  const [activeItemIndex, setActiveItemIndex] = React.useState(0)
+
+  const inputRef = React.useRef<HTMLInputElement>(null)
+  const containerRef = React.useRef<HTMLDivElement>(null)
+
+  useHotkey(containerRef, { key: "ArrowDown" }, () => {
+    if (!isOpen) {
```
stereobooster: you can skip check current value, because React will check it anyway and will not update if it is the same as current one

src/SearchInput/SearchInput.tsx
```diff
@@ -0,0 +1,274 @@
...
+  )
+}
+
+const Container = styled.div<{ hasCategory: boolean; isOpen: boolean }>`
+  display: grid;
+  cursor: text;
+  grid-template-columns: ${props => (props.hasCategory ? "16px min-content auto" : "16px auto")};
+  border: 1px solid transparent; /* Avoid jump with hover state */
+  border-bottom: 1px solid ${({ theme }) => theme.color.separators.light};
+  gap: ${({ theme, hasCategory }) => theme.space.content + (hasCategory ? 8 : 0)}px;
```
mpotomin: 8px = constans.space.small

src/SearchInput/SearchInput.tsx
```diff
@@ -0,0 +1,274 @@
...
+  display: flex;
+  align-items: center;
+  position: relative;
+  cursor: pointer;
+
+  ${({ highlighted, theme }) => (highlighted ? `background-color: ${theme.color.background.lighter};` : "")}
+
+  :hover {
+    background-color: ${({ theme, highlighted }) =>
+      highlighted ? theme.color.background.grey : lighten(theme.color.primary, 54)};
```
mpotomin: we use the color that is `lighten(theme.color.primary, 54)` all over the Catalog. I think it's time to give it it's own constant as well in @operational

src/SearchInput/SearchInput.tsx
```diff
@@ -0,0 +1,274 @@
...
+  border: 1px solid ${({ theme }) => theme.color.separators.light};
+`
+
+const DropdownItem = styled.div<{ isActive: boolean }>`
+  height: 48px;
+  display: grid;
+  align-items: center;
+  width: 100%;
+  grid-template-columns: 16px min-content auto;
+  grid-gap: ${({ theme }) => theme.space.content + 8}px;
```
mpotomin: why 8 and not 6 or 10?

src/SearchInput/SearchInput.tsx
```diff
@@ -0,0 +1,274 @@
...
+  background-color: ${({ theme }) => theme.color.white};
+  border: 1px solid ${({ theme }) => theme.color.separators.light};
+`
+
+const DropdownItem = styled.div<{ isActive: boolean }>`
+  height: 48px;
+  display: grid;
+  align-items: center;
+  width: 100%;
+  grid-template-columns: 16px min-content auto;
```
mpotomin: why 16 and not `constant.space.content`?

src/SearchInput/SearchInput.tsx
```diff
@@ -0,0 +1,274 @@
...
+         * So we need to trap the focus before and after our `Input`
+         */
+        <InvisibleOverlay
+          onClick={e => {
+            e.stopPropagation()
+            setIsOpen(false)
+          }}
+          onFocus={() => setIsOpen(false)} /* Close dropdown if the focus is away */
+          tabIndex={0}
+          aria-hidden="true"
```
stereobooster: I'm not an expert either, but if something is focusable it can't be `aria-hidden`. Screen reader goes through focusable items and announce them

src/SearchInput/SearchInput.tsx
stereobooster: > CSS Grid Layout initially defined the grid-gap property. This prefixed property is being replaced by gap. However, in order to support browsers that implemented grid-gap and not gap for grid, you will need to use the prefixed property as in the interactive example above.

https://developer.mozilla.org/en-US/docs/Web/CSS/gap

src/SearchInput/SearchInput.tsx
```diff
@@ -0,0 +1,274 @@
...
+  height: 48px;
+  display: grid;
+  align-items: center;
+  width: 100%;
+  grid-template-columns: 16px min-content auto;
+  grid-gap: ${({ theme }) => theme.space.content + 8}px;
+  padding: 0 ${({ theme }) => theme.space.content}px;
+  cursor: pointer !important;
+
+  ${({ isActive }) => (isActive ? "" : ":hover {")}
```
stereobooster: I kind of get it, but it is a bit tricky to read

icons/Enter.svg
fabien0102: This is the raw svg, the optimization is part of building tool to avoid this kind of manual micro optimisation ;)

src/SearchInput/SearchInput.tsx
fabien0102: It's just a good practice to put every deps here, don't change anything if this doesn't change, so…

src/SearchInput/SearchInput.tsx
fabien0102: I know, but it was this or a huge duplication…

src/SearchInput/SearchInput.tsx
fabien0102: image

Do we really want to debate about this?

src/SearchInput/SearchInput.tsx
fabien0102: I don't really know, it was like this in our ContextMenu and it seams to works 🙄

src/SearchInput/SearchInput.tsx
fabien0102: But since I also use `grid-gap` above, let's be consistant

src/SearchInput/SearchInput.tsx
fabien0102: I'm just following the design at some point, try to make this pretty, and it's all the point of CSS 🙄

src/SearchInput/SearchInput.tsx
fabien0102: I took this value from the catalog to be consistant, so I guess it make a lot of sense! What about `lighterPrimary`?

src/SearchInput/SearchInput.tsx
fabien0102: We can ^^

src/SearchInput/SearchInput.tsx
fabien0102: ???

fabien0102:
> @fabien0102 Looks great! Just one idea: when the user clicks on the select dropdown, I would actually expect it to close after selecting a value, not jump into the search input text box. In our use case, it is common for people to not enter a search string at all but use the select to filter filter the result types.

It's actually a bit more subtil, if you have a value and click bellow, I close the dropdown BUT if you don't have any value set and click bellow, I focus the search so you can set your filter ;)

src/SearchInput/SearchInput.tsx
fabien0102: I will do this in another PR, I'm not how to set up this variable to follow the primary color easily with `constant.ts`

PR #1281, opened 2020-01-15
**Feature:** Improve multiselect

stereobooster:
# Summary

Screenshot 2020-01-15 at 12 02 31

# Related issue

# To be tested

Me

Tester 1

Tester 2

### DIFF
--- a/src/Select/Select.styled.ts
+++ b/src/Select/Select.styled.ts
@@ -75,6 +75,11 @@ export const ClearButton = styled.div`
 
   :focus {
     box-shadow: none;
+    outline: none;
+
+    svg {
+      fill: ${({ theme }) => theme.color.primary};
+    }
   }
 `
 
--- a/src/Select/Select.styled.ts
+++ b/src/Select/Select.styled.ts
@@ -102,7 +102,7 @@ export const DropdownButton = styled.div<{ naked: boolean; isOpen: boolean; hasC
 
 DropdownButton.defaultProps = { role: "button", "aria-disabled": false, "aria-label": "Expand" }
 
-export const CheckboxConteiner = styled.div`
+export const CheckboxContainer = styled.div`
   height: ${rowHeight}px;
   display: flex;
   align-items: center;
--- a/src/Select/Select.styled.ts
+++ b/src/Select/Select.styled.ts
@@ -60,6 +60,22 @@ export const FilterInput = styled(Input)`
   max-width: none;
 `
 
+export const ClearButton = styled.div`
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  margin: -1px 0;
+  border: none;
+  border-top-right-radius: ${({ theme }) => theme.borderRadius}px;
+  border-bottom-right-radius: ${({ theme }) => theme.borderRadius}px;
+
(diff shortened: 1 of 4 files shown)


src/Select/Select.styled.ts
```diff
@@ -84,3 +85,9 @@ export const DropdownButton = styled("div")<{ naked: boolean; isOpen: boolean; h
 `
 
 DropdownButton.defaultProps = { role: "button", "aria-disabled": false, "aria-label": "Expand" }
+
+export const CheckboxConteiner = styled.div`
```
mpotomin:
```suggestion
export const CheckboxContainer = styled.div`
```
I tiny typo

mpotomin:
Would be great to have a button to deselect all the selected items with one click 💃

stereobooster:
Screenshot 2020-01-15 at 17 09 30

stereobooster:
Aaah you caught that bug I saw it but wasn't sure if its still there

mpotomin:
I like how the X (clear) button works in this case. 

I can imagine though, that for a user it is not obvious from the first look that clicking this button will clear all the selected options. They might think that it will close the drop down. Would it make sense to keep both controls for this case - "arrow up" and the "clear" button. I'd also suggest making X same active (shade of blue) colour for consistency. 

Whats your take on it, @kemal-contiamo?

kemal-contiamo:
> I like how the X (clear) button works in this case.
> 
> I can imagine though, that for a user it is not obvious from the first look that clicking this button will clear all the selected options. They might think that it will close the drop down. Would it make sense to keep both controls for this case - "arrow up" and the "clear" button. I'd also suggest making X same active (shade of blue) colour for consistency.
> 
> Whats your take on it, @kemal-contiamo?

@stereobooster I second what @mpotomin suggested, I think we need to show both clear and arrow up button for more clarity. Perhaps the order would be (from left to right); the X button and then the Up arrow, when the dropdown is opened.

stereobooster:
Screenshot 2020-01-16 at 13 58 00

mpotomin:
> Clicking the "chevron up" button does not close the select anymore, whereas the case with no multi-select works as expected
> 
> 



Added this issue to the context menu [refactoring CIP](https://www.notion.so/contiamo/CIPs-d771790a0b164bbe9e2aeeb7be93179e?p=0afec0b5417a4cf5b12959ee817013b2)

stereobooster:
Custom options "jumps", but it is the same in the master

mpotomin:
> Custom options "jumps", but it is the same in the master

Screenshot 2020-01-16 at 17 33 58

👀 👀 👀 This focus ring though  👀 👀 👀

stereobooster:
Screenshot 2020-01-16 at 17 53 35

mpotomin:
The custom focus styling works! 🎉 

Getting this weird behaviour when trying to use the X button with my keyboard:
Triggering X with my keyboard only clears the last set value `Option 5` while keeping previously set `Option 1` and `Option 3` enabled

stereobooster:
It would take a bit more work to implement clear button right, so for now let's merge without it and add it in the next iteration.

PR #1285, opened 2020-01-20
**Feature:** Allow advanced use-case on Table

fabien0102:
# Summary

We have this use-case in our product:

image

To do this, we need a way to tweak a bit the table data element (to pass `backgroundColor` and `colSpan`. The idea of this PR is to permit a bit more control to support non-standard use-cases. If our non-standard use cases become standard, we will extend the API later 😉

We can also send an explicit `false` to skip a `td` (needed for the `colSpan` use-case)

Note: Since it's not really standard, I would prefer to not expose this to the official doc.

# Related issue

# To be tested

Me

Tester 1

Tester 2

### DIFF
--- a/src/Table/Table.tsx
+++ b/src/Table/Table.tsx
@@ -356,7 +356,7 @@ function Table<T>({
                                 key={columnIndex}
                                 {...(column.tdProps ? column.tdProps(dataEntry, dataEntryIndex) : {})}
                               >
-                                {column.cell(dataEntry, dataEntryIndex)}
+                                {cell}
                               </Td>
                             )
                           })}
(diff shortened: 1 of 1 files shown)


src/Table/Table.tsx
```diff
@@ -340,11 +344,22 @@ function Table<T>({
...
+                            if (cell === false) return
+
+                            return (
+                              <Td
+                                coloredBorders={shouldTdHaveColoredBorders}
+                                cellWidth={column.width}
+                                key={columnIndex}
+                                {...(column.tdProps ? column.tdProps(dataEntry, dataEntryIndex) : {})}
+                              >
+                                {column.cell(dataEntry, dataEntryIndex)}
```
stereobooster:
```suggestion
                                {cell}
```

PR #1289, opened 2020-01-24
**Fix:** Increase LabelText font size

fabien0102:
# Summary

So the label is not smaller than a Body 🙄 

# Related issue

# To be tested

Me

Tester 1

Tester 2


mpotomin:
Screenshot 2020-01-24 at 10 40 12

fabien0102:
This was a mistake, I should not code on Friday 😅

PR #1297, opened 2020-02-17
**Fix:** bug when tooltips get stuck on scroll

stereobooster:
# Summary

The problem was due to dependencies in open/close functions so they got called way to much times.

# Related issue

# To be tested

Me

Tester 1

- [ ] Resize window to make table horizontally scrollable, check that tooltip disappears on all scroll directions

Tester 2

- [ ] Resize window to make table horizontally scrollable, check that tooltip disappears on all scroll directions

### DIFF
--- a/src/DataTable/DataTable.tsx
+++ b/src/DataTable/DataTable.tsx
@@ -52,6 +52,11 @@ export function DataTable<Columns extends any[][], Rows extends any[][]>({
 }: DataTableProps<Columns, Rows>) {
   const { open, close, viewMorePopup } = useViewMore()
   const rowHeight = React.useMemo(() => getRowHeight(initialRowHeight), [initialRowHeight])
+  React.useEffect(() => {
+    const onScroll = () => close(true)
+    window.addEventListener("scroll", onScroll)
+    return () => window.removeEventListener("scroll", onScroll)
+  }, [close])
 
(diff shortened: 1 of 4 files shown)


src/DataTable/DataTable.tsx
```diff
@@ -50,12 +50,19 @@ export function DataTable<Columns extends any[][], Rows extends any[][]>({
...
 }: DataTableProps<Columns, Rows>) {
+  const ref = React.useRef<HTMLDivElement>(null)
   const { open, close, viewMorePopup } = useViewMore()
   React.useEffect(() => {
-    const onScroll = () => close()
-    window.addEventListener("scroll", onScroll)
-    return () => window.removeEventListener("scroll", onScroll)
-  }, [close])
+    if (ref.current) {
+      const onScroll = () => close()
```
mpotomin: Can we pass `close` directly as a handler, without adding a `const onScroll`?

src/DataTable/useViewMore.ts
```diff
@@ -10,32 +10,34 @@ const useViewMore = () => {
     false,
   )
 
-  const close = React.useCallback(() => {
-    if (viewMorePopup) {
+  const close = React.useCallback((immediately?: any) => {
```
stereobooster: objections?

PR #1301, opened 2020-02-19
  **Breaking:** CON-1155 Improve Tree

mpotomin:
# Why breaking:
Changes the existing API of the Tree component by renaming `color` prop to `tagColor`. 
Motivation: to make API more explicit and clear, avoid implicitly confusing the prop name with the font color and other color props, that this PR adds. 

# Why draft:
- [ ] Apply updated tree in the UI and make sure all the cases follow the requirememtns
- [x] Improve selected state of the tree items
- [x] Make it possible to ignore search words highlight for selected items
- [x] Add custom right padding to be able to align row actions
- [x] Add handling for "loading" and "error" states
- [x] Make the tree able to handle "+ Add table" type of items. 

# Summary
This PR:
1. Fixes the jumps of the Accordion's section header when toggled
1. Extends the customisation options for the Tree component items:

# Related issue
https://contiamo.atlassian.net/browse/CON-1155

# To be tested

Me

Tester 1

Tester 2

### DIFF
--- a/src/AccordionSection/AccordionSection.tsx
+++ b/src/AccordionSection/AccordionSection.tsx
@@ -39,6 +39,7 @@ const Header = styled(SectionHeader)<{ expanded: boolean }>(({ theme, expanded }
   ".no-focus &:focus": {
     background: theme.color.background.almostWhite,
   },
+  paddingLeft: theme.space.small,
   paddingRight: 0,
   userSelect: "none",
 }))
@@ -50,7 +51,7 @@ const Panel = styled.div`
   /* we need it because of overflow: hidden; above */
   overflow: auto;
   height: 100%;
-  padding: ${({ theme }) => theme.space.element}px;
+  padding: ${({ theme }) => theme.space.small}px;
   background-color: ${({ theme }) => theme.color.white};
 `
 
--- a/src/AccordionSection/AccordionSection.tsx
+++ b/src/AccordionSection/AccordionSection.tsx
@@ -51,7 +51,6 @@ const Panel = styled.div`
   /* we need it because of overflow: hidden; above */
   overflow: auto;
   height: 100%;
-  padding: ${({ theme }) => theme.space.small}px;
   background-color: ${({ theme }) => theme.color.white};
 `
 
--- a/src/AccordionSection/AccordionSection.tsx
+++ b/src/AccordionSection/AccordionSection.tsx
@@ -12,10 +12,11 @@ export interface AccordionSectionProps extends DefaultProps {
   // for internal use
   _index?: number
   _expanded?: boolean
+  _last?: boolean
   _toggleSection?: (index: number) => void
 }
 
-const Container = styled.div<{ expanded: boolean }>`
+const Container = styled.div`
   label: AccordionSectionContainer;
   /* to make sure it respects parrent grid's row height */
   overflow: hidden;
@@ -26,10 +27,10 @@ const Container = styled.div<{ expanded: boolean }>`
   position: relative;
 `
 
-const Header = styled(SectionHeader)<{ expanded: boolean }>(({ theme, expanded }) => ({
+const Header = styled(SectionHeader)<{ expanded: boolean; last: boolean }>(({ theme, expanded, last }) => ({
   cursor: "pointer",
   borderTop: `1px solid ${theme.color.border.invisible}`,
-  borderBottom: `1px solid ${expanded ? "transparent" : theme.color.border.invisible}`,
+  borderBottom: `1px solid ${last && !expanded ? theme.color.border.invisible : "transparent"}`,
   background: theme.color.background.almostWhite,
   // disable browser focus to customise focus state
   ":focus": {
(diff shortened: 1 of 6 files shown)


src/Tree/TreeItem.tsx
```diff
@@ -88,15 +102,25 @@ const highlightStyle: React.CSSProperties = {
...
   /* Split the label by caract properly and show the first line only */
   overflow-wrap: break-word;
   overflow: hidden;
   text-overflow: ellipsis;
   height: 16px;
 
-  font-size: ${({ theme }) => theme.font.size.small}px;
-  font-weight: ${({ theme, hasChildren }) => (hasChildren ? theme.font.weight.bold : theme.font.weight.medium)};
+  line-height: 16px;
+  color: ${({ fontColor }) => (fontColor ? expandColor(constants, fontColor) : "unset")};
```
stereobooster: why `unset`?

src/Tree/ChildTree.tsx
```diff
@@ -18,13 +18,15 @@ const Container = styled.div<{ hasChildren: boolean; disabled: boolean }>`
 `
 
 const ChildTree: React.SFC<Props> = ({
+  paddingLeft,
+  paddingRight,
   initiallyOpen,
   highlight,
   tag,
+  tagColor,
```
stereobooster: `**Breaking:**`?

mpotomin: Yes totally, good catch! 👍

src/Tree/TreeItem.tsx
mpotomin: @stereobooster 
The idea is: if no `fontColor` provided explicitly - make the font color of the label same as the the one of parent - `<Header/>` in this case. 

Alternatively, I'd suggest using `inherit` here as a value. Would `inherit` convey the intent better in this case?

stereobooster: yes 👍

src/AccordionSection/AccordionSection.tsx
```diff
@@ -29,7 +29,7 @@ const Container = styled.div<{ expanded: boolean }>`
 const Header = styled(SectionHeader)<{ expanded: boolean }>(({ theme, expanded }) => ({
   cursor: "pointer",
   borderTop: `1px solid ${theme.color.border.invisible}`,
-  borderBottom: expanded ? `1px solid ${theme.color.border.invisible}` : "none",
+  borderBottom: `1px solid ${expanded ? "transparent" : theme.color.border.invisible}`,
```
stereobooster: transparent vs invisible 🤔

mpotomin: Invisible is only a title 🤷‍♂, poor naming selection though, I agree. What would you suggest in this case @stereobooster?

stereobooster:
Highlighted text makes non-emphasised lines to jump, for example type "Limited"

stereobooster:
Can tab into disabled item

Screenshot 2020-02-21 at 16 47 51

src/Tree/TreeItem.tsx
```diff
@@ -169,24 +206,36 @@ const TreeItem: React.SFC<TreeItemProps> = ({
...
+        strong={Boolean(strong)}
+        fontSize={fontSize ? fontSize : constants.font.size.small}
+        emphasized={Boolean(emphasized)}
+        monospace={Boolean(monospace)}
+      >
+        <Highlighter
+          textToHighlight={label}
+          highlightStyle={highlightStyle}
+          sanitize={sanitizeInput}
+          searchWords={!ignoreSearchWords ? searchWords : []}
```
stereobooster: It maybe good idea to use `const emptyArray = []`, to get the same instance every time here

src/Tree/Tree.tsx
```diff
@@ -7,19 +7,28 @@ import { IconComponentType } from "../Icon"
...
   iconColor?: string
   onClick?: (event: React.MouseEvent) => void
   onContextMenu?: (event: React.MouseEvent<HTMLDivElement, MouseEvent>) => void
   cursor?: string
   actions?: React.ReactNode
   forwardRef?: (element?: HTMLElement | null) => any
+  strong?: boolean
+  fontSize?: number
+  fontColor?: string
+  emphasized?: boolean
```
stereobooster: `strong vs emphasized` 🤔

src/Tree/ChildTree.tsx
```diff
@@ -33,11 +36,17 @@ const ChildTree: React.SFC<Props> = ({
...
   cursor,
   searchWords,
+  ignoreSearchWords,
   level,
   actions,
   hasIconOffset,
   onMouseEnter,
   onMouseLeave,
+  strong,
+  fontSize,
```
stereobooster: Does it make sense to change fontSize across tree nodes? Shouldn't they all be the same size 🤔

stereobooster:
Generally looks good -  the only concern is "jumping" text on highlight in the tree

src/Tree/TreeItem.tsx
mpotomin: Yep, makes sense!

src/Tree/ChildTree.tsx
mpotomin: It is same size if not modified, but current designs in Zepplin require smaller sizes for items in the deeper levels of nesting. Let's make it tweakable so we have a flexibility to decide later 

Screenshot 2020-02-21 at 17 11 21

mpotomin:
> Highlighted text makes non-emphasised lines to jump, for example type "Limited"

Let me tweak the highlight style a bit, maybe that can be fixed easily

PR #1302, opened 2020-02-25
[WiP] **Feature:** use context

stereobooster:
# Summary

Just idea how to liberate confirm API from page component.

**Broken for this case**:

Screenshot 2020-02-25 at 15 17 21

# Related issue

# To be tested

Me

Tester 1

Tester 2


stereobooster:
My recommendation is to deprecate confirm and modal and instead use `Modal` component

PR #1306, opened 2020-03-06
**Fix:** select in modal

stereobooster:
# Summary

Before:
Screenshot 2020-03-06 at 13 13 54

After:
Screenshot 2020-03-06 at 13 14 37

# Related issue

https://github.com/contiamo/operational-ui/pull/1305

# To be tested

Me

Tester 1

Tester 2

### DIFF
--- a/src/utils/constants.ts
+++ b/src/utils/constants.ts
@@ -216,14 +216,14 @@ export const space = {
 const zIndex = {
   /** `0` */
   default: 0,
-  /** `300` */
-  selectOptions: 300,
   /** `299` */
   formFieldError: 299,
   /** `400` */
   tooltip: 400,
   /** `500` */
   modal: 500,
+  /** `550` */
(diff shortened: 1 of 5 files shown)


src/utils/constants.ts
```diff
@@ -216,14 +216,14 @@ export const space = {
...
-  /** `300` */
-  selectOptions: 300,
   /** `299` */
   formFieldError: 299,
   /** `400` */
   tooltip: 400,
   /** `500` */
   modal: 500,
+  /** `550` */
+  selectOptions: 550,
```
mpotomin: This whole `zIndex ` object and changes in it remember me so much of this game:
https://www.youtube.com/watch?v=AxtTkkANR6A 🤣 

Trying to think of cases when we need to have open select below a modal 🤔, can't find those so far. The demo looks good. Let's test it in the application

PR #1307, opened 2020-03-06
**Fix:** z-index for select

stereobooster:
# Summary

**Do not merge**

See `Components/DropdownButton`

Before:

Screenshot 2020-03-06 at 15 08 26
Screenshot 2020-03-06 at 15 08 34

After:

Screenshot 2020-03-06 at 15 10 28
Screenshot 2020-03-06 at 15 10 34

# Related issue

https://github.com/contiamo/operational-ui/pull/1305

# To be tested

Me

Tester 1

Tester 2

### DIFF
--- a/src/ContextMenu/ContextMenu.tsx
+++ b/src/ContextMenu/ContextMenu.tsx
@@ -77,7 +77,7 @@ const InvisibleOverlay = styled("div")(({ theme }) => ({
   right: 0,
   left: 0,
   cursor: "default",
-  zIndex: theme.zIndex.selectOptions + 1,
+  zIndex: theme.zIndex.selectContainer + 1,
 }))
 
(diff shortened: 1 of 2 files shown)


mpotomin:
@stereobooster I honestly feel very frustrated about the need of the 3rd PR in a row to fix the same thing. It's a lot of back-and-forth of reviewing and testing, resulted in wasted time and energy. Could you please confirm that you applied this PR to contiamo-ui locally and tested it in all the use use-cases?

mpotomin:
Just a random idea. Would it make sense to provide z-index for the Select's drop-down from the outside, as a prop? So we could handle the edge cases like "Select within the Modal", "Select above some 3rd-party component in-place, as an escape-hatch?

PR #1308, opened 2020-03-07
**Feature:** limit width of select

stereobooster:
# Summary

Instead of z-index war I decided to limit width of context menu in Select.

Problems:

- `ContextMenu` reused in 4 components: `Select`, `DropdownButton`, `DataTableHeader`, `Toparselect`. As soon as we change it in one place it breaks in different
- z-index war. As soon as we start to assign z-index we will find a scenario where it doesn't work and we need to assign even higher z-index to something else. Instead it is better to rely on browsers natural flow.

# Related issue
 
#1305

# To be tested

Me

Tester 1

Tester 2


src/Select/README.md
```diff
@@ -461,57 +458,9 @@ const MyComponent = () => {
...
-      `}
-      </Code>
-    </>
-  )
-}
-
-;<MyComponent />
-```
-
-### Select in Modal
```
mpotomin: @stereobooster Let's keep this use case in the doc, could be very helpful for the future.

PR #1310, opened 2020-03-11
**Fix:** filterable select

stereobooster:
# Summary

Before
Screenshot 2020-03-11 at 14 46 04

After
Screenshot 2020-03-11 at 14 45 50

# Related issue

#1308

# To be tested

Me

Tester 1

- [x] Test all components that use ContextMenu
- [x] Test all Selects

Tester 2

### DIFF
--- a/src/ContextMenu/ContextMenu.Item.tsx
+++ b/src/ContextMenu/ContextMenu.Item.tsx
@@ -30,7 +30,7 @@ export interface IContextMenuItem<TValue = any> {
 
 export const rowHeight = 32
 
-const Container = styled("div")<Props>(({ condensed, disabled, align, theme, width }) => {
+const Container = styled.div<Props>(({ condensed, disabled, align, theme, width }) => {
   return {
     userSelect: "none",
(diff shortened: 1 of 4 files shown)


mpotomin:
@stereobooster 
Didn't find regressions. Tested so far with:
- ContextMenu
- Select
- Action menu
- Autocomplete
- Card section menu
- Context menu
- DataTable(Header)
- TopbarSelect

Is this PR still a draft?

stereobooster:
@mpotomin yes PR ready. I see some regressions (I think), but not connected to this PR (probably was introduced before). For example this looks a bit strange

Screenshot 2020-03-11 at 16 26 25

mpotomin:
> @mpotomin yes PR ready. I see some regressions (I think), but not connected to this PR (probably was introduced before). For example this looks a bit strange
> 
> Screenshot 2020-03-11 at 16 26 25

I've noticed some jumps in the Selects too, like:



But this jumping was there before the whole `Select-width` Saga started, so not a blocker for this particular PR.

PR #1318, opened 2020-03-17
**Fix:** Toggle tree only when clicking on the arrow

stereobooster:
# Summary

Now onClick action and toggle action are separated. Toggle triggered only when use click arrow (or with space key)

# Related issue

N/A

# To be tested

Me

Tester 1

Tester 2

### DIFF
--- a/src/Tree/ChildTree.tsx
+++ b/src/Tree/ChildTree.tsx
@@ -64,14 +64,16 @@ const ChildTree: React.SFC<Props> = ({
   )
 
   const onNodeClick =
-    !disabled && onClick
+    !disabled && (onClick || onNodeContextMenu)
       ? (e: React.MouseEvent<HTMLDivElement>) => {
           e.stopPropagation()
           if (e.altKey && onNodeContextMenu) {
             onNodeContextMenu(e)
             return
           }
-          onClick(e)
+          if (onClick) {
+            onClick(e)
+          }
         }
       : undefined
 
--- a/src/Tree/ChildTree.tsx
+++ b/src/Tree/ChildTree.tsx
@@ -64,7 +64,7 @@ const ChildTree: React.SFC<Props> = ({
   )
 
   const onNodeClick =
-    !disabled && (onClick || onNodeContextMenu)
+    !disabled && (onClick || onNodeContextMenu || hasChildren)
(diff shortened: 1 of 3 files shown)


mpotomin:
With this change it is hard to see if anything is supposed to happen when I click a row in the tree or not as there is no visual communication. (For comparison - If there is a context menu attached to the row - we show the "3 dots" icon). @kemal-contiamo Please confirm that this is what is expected.

Also, clicking on the row of the tree in the Workbench will also not work anymore, user will have to target the chevron icon with the click. @kemal-contiamo please approve that this is the expected behaviour as well (see gif below for the example of current tree behaviour)

tree

kemal-contiamo:
> With this change it is hard to see if anything is supposed to happen when I click a row in the tree or not as there is no visual communication. (For comparison - If there is a context menu attached to the row - we show the "3 dots" icon). @kemal-contiamo Please confirm that this is what is expected.
> 
> Also, clicking on the row of the tree in the Workbench will also not work anymore, user will have to target the chevron icon with the click. @kemal-contiamo please approve that this is the expected behaviour as well (see gif below for the example of current tree behaviour)
> 
> tree

@mpotomin I did recommended to @stereobooster to implement the behaviours as follows:
- Clicking on the arrow would collapse/uncollapse tree.
- Clicking on other area, not the arrow, would select the item and change the panel on the right accordingly.
The thing is, I recommended those behaviour in the usage context of the tree inside our new custom field management, as I can see it needs different behaviour pattern to deal with the tree there. And as I know that the component is used across the platform, I was suggesting that perhaps we can just use the recommended behaviour universally. But, looking at the context of workbench again, I see that I might be making a wrong assessment and that perhaps in workbench context, the current pattern should be maintained. The question now is if splitting the component into two behaviour patterns is feasible technically? What do you think? FYI @nathanedavis1

stereobooster:
@mpotomin what do you think about latest commit?

mpotomin:
> @mpotomin what do you think about latest commit?

Looks like a good option to me. As we agreed on a [different solution](https://contiamo.slack.com/archives/C0PF194MT/p1584537843008100?thread_ts=1584535258.002200&cid=C0PF194MT) for now in the chat, let's keep this idea in mind in case if another iteration happens

PR #1319, opened 2020-03-18
**Breaking:** CON-1281 Introduce more constants for primary color variations

mpotomin:
# Summary
Add more constants that represent primary color variations

# Related issue
CON-1281

# To be tested

Me

Tester 1

Tester 2

### DIFF
--- a/src/utils/constants.ts
+++ b/src/utils/constants.ts
@@ -84,7 +84,6 @@ const textColors = {
  * A specialized color palette for borders.
  */
 const borderColors = {
-  primary: setAlpha(0.1)(primaryColor),
   /** `#c0c0c0` */
   default: "#c0c0c0",
   /** `#ddd` */
@@ -123,6 +122,7 @@ const color = {
   ghost: "hsla(0, 0%, 100%, 0.33)",
   white: whiteColor,
   highlight: setAlpha(0.05)(primaryColor),
+  accent: setAlpha(0.1)(primaryColor),
   /** `#000` */
   black: "#000",
   background: backgroundColors,
--- a/src/utils/constants.ts
+++ b/src/utils/constants.ts
@@ -243,7 +243,7 @@ const zIndex = {
 const shadows = {
   pressed: "inset 0 1px 1px rgba(0,0,0,0.15)",
   topBar: "0px 1px 5px #d3d1d1",
-  focus: `inset 0 1px 1px rgba(0,0,0,.075), 0 0 8px ${setAlpha(0.1)(primaryColor)}`,
+  focus: `inset 0 1px 1px rgba(0,0,0,.075), 0 0 8px ${color.accent}`,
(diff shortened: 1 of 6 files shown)


src/utils/constants.ts
```diff
@@ -84,6 +84,7 @@ const textColors = {
  * A specialized color palette for borders.
  */
 const borderColors = {
+  primary: setAlpha(0.1)(primaryColor),
```
stereobooster: We already have:

```ts
const color = {
  primary: primaryColor,
```

the same name would be confusing

stereobooster:
Do you want to consolidate all occurrences of `setAlpha(0.05)(theme.color.primary)`?

src/utils/constants.ts
mpotomin: Sure, what would you suggest instead?

stereobooster: Hm 🤔 `lightPrimary` maybe? (there is darkPrimary)
Or maybe "accent" color?

mpotomin: We already have the `highlight`, that is 5% alpha of the `primary`, so let's try `accent` for 10% alpha of the `primary`, that we could occasionally use for borders as well

stereobooster: `highlightBorder` maybe?

mpotomin: Sounds too specific for the border. There are cases in the product where we need it not for the border only. Shall we settle for `constants.color.accent `?

mpotomin: Just brought it up in the team sync, agreed to use `constants.color.accent` in this case. See [this ticket](https://contiamo.atlassian.net/browse/CON-1281)

mpotomin:
@stereobooster I think I've addressed all the open discussions. Please take a look again 🙏

src/utils/constants.ts
```diff
@@ -122,6 +122,7 @@ const color = {
   ghost: "hsla(0, 0%, 100%, 0.33)",
   white: whiteColor,
   highlight: setAlpha(0.05)(primaryColor),
+  accent: setAlpha(0.1)(primaryColor),
```
stereobooster: I wonder if this will work with dynamically changed `primaryColor` via theme 🤔

mpotomin: Ah snap, I guess it won't 😭

mpotomin: Should be fixed now

src/utils/constants.ts
```diff
@@ -121,8 +121,6 @@ const color = {
   basic: "#636363",
   ghost: "hsla(0, 0%, 100%, 0.33)",
   white: whiteColor,
-  highlight: setAlpha(0.05)(primaryColor),
```
stereobooster: this can be a breaking change

mpotomin: Good point, updated the description

Issue #1322, opened 2020-03-20
Tree selection highlight color


mpotomin:
If we submit a different theme primary color (i.e. green), the Tree search highlight dies not follow and stays blue:

<img width="636" alt="Screenshot 2020-03-20 at 10 01 50" src="https://user-images.githubusercontent.com/639406/77149684-717cdf80-6a92-11ea-9a4d-8320ebd6725a.png">

mpotomin:
Fixed

PR #1323, opened 2020-03-20
**Fix** usages of primary theme color for action text

mpotomin:
# Summary
This PR fixes the case when updated theme primary color was not applied to the action text 

# Related issue
https://github.com/contiamo/operational-ui/issues/1322

# To be tested

Me

Tester 1

Tester 2

### DIFF
--- a/src/Tree/Tree.tsx
+++ b/src/Tree/Tree.tsx
@@ -4,13 +4,9 @@ import { Draggable, DraggableProps, Droppable, DroppableProps, DroppableStateSna
 import styled from "../utils/styled"
 import ChildTree from "./ChildTree"
 import { IconComponentType } from "../Icon"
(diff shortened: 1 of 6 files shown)


src/Tree/Tree.tsx
```diff
@@ -70,9 +75,18 @@ const Tree: React.SFC<TreeProps> = ({
   placeholder,
   searchWords,
   freeze,
+  theme,
 }) => {
   const isLowestLevel = trees.length === 0 || trees.some(tree => !tree.childNodes || !tree.childNodes.length)
 
+  const highlightStyle = useMemo(
```
mpotomin: Caching it here on the Tree level with `useMemo` to avoid creating a new instance per tree item.

mpotomin: Fixed it differently

PR #1325, opened 2020-04-21
**Feature:** Support arrays in DataTable

stereobooster:
# Summary

Support arrays in DataTable

With `String()`:
Screenshot 2020-04-21 at 10 27 15

With `JSON.stringify`:
Screenshot 2020-04-21 at 10 27 15

Or I can implement custom "stringifier"?

# Related issue

# To be tested

Me

Tester 1

Tester 2

### DIFF
--- a/src/DataTable/CellContent.tsx
+++ b/src/DataTable/CellContent.tsx
@@ -11,9 +11,9 @@ export interface CellContentProps {
   close: () => void
 }
 
-const stringifyBooleanAndNull = (value: any) => {
+const stringifyBooleanAndNullAndArray = (value: any) => {
   // We compare booleans like this and without typeof for perf
-  return value === true || value === false || value === null ? String(value) : value
+  return value === true || value === false || value === null || Array.isArray(value) ? String(value) : value
 }
 
 const CellContent: React.FC<CellContentProps> = ({ cell, open, close }) => {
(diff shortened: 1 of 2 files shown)


stereobooster:
@mpotomin WDYT?

mpotomin:
> @mpotomin WDYT?

Could you elaborate what exactly are asking for?

stereobooster:
I'm sorry. I implemented support for arrays with `String()` (see first screenshot, it doesn't work for `[null,null]`). I also can use `JSON.stringify` (second screenshot), or I can implement something else. What your thoughts on that

mpotomin:
> I'm sorry. I implemented support for arrays with `String()` (see first screenshot, it doesn't work for `[null,null]`). I also can use `JSON.stringify` (second screenshot), or I can implement something else. What your thoughts on that

using `JSON.stringify` sounds like a good idea, since it will take care of the issue with `null`
Screenshot 2020-04-21 at 11 05 36

@stereobooster What would be the advantage of using a custom "stringifier"?

stereobooster:
> What would be the advantage of using a custom "stringifier"?

Some kind of custom behaviour, if we need one

mpotomin:
> > What would be the advantage of using a custom "stringifier"?
> 
> Some kind of custom behaviour, if we need one

Do we need one now? If not, `JSON.stringify` could be good enough

stereobooster:
JSON stringify can throw Exceptions, we probably don't need to worry about it...

> Throws a TypeError ("cyclic object value") exception when a circular reference is found.
> Throws a TypeError ("BigInt value can't be serialized in JSON") when trying to stringify a BigInt value.

micha-f:
JSON.stringify looks good

PR #1329, opened 2020-05-04
**Feature:** Speed up the dev environment

fabien0102:
# Summary

Quick and simple action to make the dev env a bit faster -> don't parse the props when we are in dev.

Impact:
* No more props info in dev
image

* 30sc less compilation time (on my beasty computer)

# Related issue

# To be tested

Me

Tester 1

Tester 2


TejasQ:
Great find! 👌

PR #1334, opened 2020-05-12
**Fix:** first tree item sometimes ignores initially open flag

stereobooster:
# Summary

first tree item sometimes ignores `initiallyOpen` flag.

In new example it is possible to open close items, if key value changes

# Related issue

# To be tested

Me

Tester 1

Tester 2

### DIFF
--- a/src/Tree/README.md
+++ b/src/Tree/README.md
@@ -737,6 +737,7 @@ import { Tree, Button, OlapIcon } from "@operational/components"
 const initialTrees = [
   {
     label: "ERP",
+    key: 10,
     childNodes: [
       {
         label: "Region",
@@ -762,6 +763,7 @@ const initialTrees = [
   },
   {
     label: "Legal Entity",
+    key: 20,
     childNodes: [
       {
         label: "Limited Liability Company",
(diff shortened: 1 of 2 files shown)


src/Tree/README.md
```diff
@@ -725,3 +727,74 @@ const MyComponent = () => {
...
+    ],
+  },
+]
+
+const Example = () => {
+  const [trees, setTrees] = React.useState(initialTrees)
+  const open = React.useCallback(() => {}, [setTrees])
+  return (
+    <>
+      <Button onClick={() => setTrees(initialTrees.map(x => ({ ...x, initiallyOpen: true, key: x.key + 1 })))}>
```
mpotomin, line 788: Does it mean, that every time we update the tree, we need to increment the key manually?

stereobooster: If we mean to reset state - yes. But the idea is that if we have one state and then loading and then new state it would be enough for us.

1. `[key=1,key=2,key=3]`
2. `[key=loading]`
3. `[key=1,key=2,key=3,key=4]`

Having ability to control key would give us a bit more granularity, without need to change API (from `initiallyOpen` - uncontrolled, to `open` - controlled).

We can go the other way and change API as well, I just thought that this could work

src/Tree/README.md
```diff
@@ -351,6 +351,7 @@ To offer the most flexibility as possible in your drag & drop implementation, we
 
 FIXME: https://github.com/styleguidist/react-styleguidist/issues/1278
 
+<!--
```
fabien0102, line 354: The styleguidist issue seams to be fixed, can you maybe have a look to fix this example?

stereobooster: Proposed solution (to put acorn in dev dependencies) doesn't help. On the other hand we don't use dnd feature in the tree

fabien0102: Not yet, but since we have a TableBuilder feature, I will have some dnd on the product again 😉

PR #1336, opened 2020-05-15
**Fix:** no-focus

stereobooster:
# Summary

- use mousedown instead of click, because click is triggered onkeydown on button for space and enter
- add "no-focus" on initial render, to prevent blink of focus state on first click

# Related issue

# To be tested

Me

Tester 1

Tester 2


mpotomin:
Can I please have more info how to verify the fix?

stereobooster:
To verify `document.body.classList.add("no-focus")`:

On first click on item with `no-focus` (tree in this case), we can see focus state, which is an error, because we should not see focus state on click (only on keyboard)

Screenshot 2020-05-18 at 11 13 55

Second issue is that, we use use onClick to check that click happened, but onClick is triggered on keyboard ("Enter" or "Space") for buttons, so we hide focus even though user is using keyboard (not sure this is visible in our examples)

PR #1344, opened 2020-08-01
**Fix:** Remove hasIconOffset, add <Example/> instead of <Example>

### LINKED ISSUE #1321: 4th level of tree nesting is misaligned with the parent
The items on the 4th level of nesting are not aligned vertically with the parent's label. See example for the tree "With freeze":

http://operational-ui.netlify.com/#/Components/Tree

Screenshot 2020-03-19 at 09 07 57

rimildeyjsr:
Remove hasIconOffset property to fix the extra padding for trees with no children, removes misalignment
Use <Example/> instead of <Example> to fix the warning in the code example

Fixes #1321

# Summary

The last level of a nested tree was being misaligned with the previous level. 

The cause was the `hasIconOffset` being set true for when you have an icon or a tag along with a tree that has no children. It was adding some extra padding using the formula `paddingLeft + theme.space.element * level + (hasIconOffset ? theme.space.element : 0)}px;`

On removing the `hasIconOffset`, the last levels get aligned with appropriate padding. I am removing this property altogether because any nested tree with no children was getting misaligned, and the property did not provide any other functionality.

Also, added arrows to the style of the freeze tree, as suggested in the issue

# Related issue
Fixes #1321

# To be tested

Me

Tester 1

Tester 2

### DIFF
--- a/src/Tree/ChildTree.tsx
+++ b/src/Tree/ChildTree.tsx
@@ -8,7 +8,6 @@ type Props = TreeProps["trees"][-1] & {
   searchWords?: string[]
   freeze?: boolean
   level: number
-  hasIconOffset: boolean
   onMouseEnter?: (e: React.MouseEvent<HTMLDivElement>) => void
   onMouseLeave?: (e: React.MouseEvent<HTMLDivElement>) => void
 }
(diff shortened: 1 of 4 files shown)


mpotomin:
Hi @rimildeyjsr! Looks like there is an issue with the columns still: columns are children of tables, but there is no indentation showing that (columns should be shown one level deeper under tables)
Screenshot 2020-08-12 at 09 03 41

Screenshot 2020-08-12 at 09 03 36

rimildeyjsr:
@mpotomin - Ah had not taken this into account, will fix this as well!

PR #1374, opened 2021-09-20
feat(CON-3973): Add icons for org management

vijayprwyd:
# Summary

Implemented following icons :

* Active directory group
* Active directory user icon
* Organization group icon
* Project group icon
* Organization group label
* Project group label

https://contiamo.atlassian.net/browse/CON-3973

image

image

# Related issue

# To be tested

Me

Tester 1

Tester 2


brunellaa (changes requested):
The code looks fine, but I'm noticing size differences between them, not sure if this creates disparity when used, but it looks a little off in the Component library.

Might want to recheck the viewBox attribute

image

vijayprwyd:
> The code looks fine, but I'm noticing size differences between them, not sure if this creates disparity when used, but it looks a little off in the Component library.
> 
> Might want to recheck the viewBox attribute
> 
> image

Yes noticed the same , apart from view box also noticed that the `path` tag does not inherit the width of parent which is causing the issue . eg: If svg has width of `32`px  the `path` just occupies about `24`px . Need to re-validate and make some edits on the svg icon.

LucasRoesler (commented):
LGTM :shipit:

brunellaa (approved):
The Icons look consistent now. Approved!

